PR #140, opened 2012-10-30
Test for adding version missing one node.

tomason:


mswiderski:
no reply from pull request creator so closing it

PR #152, opened 2013-02-17
Update jbpm-persistence-jpa - JpaProcessPersistenceContext to make visible constructor from parent class 

michalbali:
Adding another constructor for no-JTA environments. In other words making parent class constructor that takes isJTA parameter visible.

### DIFF
--- a/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java
+++ b/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java
@@ -17,6 +17,10 @@ public class JpaProcessPersistenceContext extends JpaPersistenceContext
     public JpaProcessPersistenceContext(EntityManager em) {
         super( em );
     }
+    
+    public JpaProcessPersistenceContext(EntityManager em, boolean isJTA) {
+        super( em, isJTA );
+    }
 
     public void persist(ProcessInstanceInfo processInstanceInfo) {
         getEntityManager().persist( processInstanceInfo );


psiroky:
@krisv @mswiderski can you please check this one out and either merge the changes or close the PR?

mswiderski:
looks like a safe change to me, @krisv wdyt?

mrietveld:
@psiroky I'm pretty sure this does not compile.. 

The closest thing that does compile is: 

    public JpaProcessPersistenceContext(EntityManager em, TransactionManager txm, boolean isJTA) {
        super( em, isJTA, txm );
    }
    
I don't really have a problem with exposing this if users want it -- except that **we do not support local transactions for the process**. So that makes exposing this method a little misleading, since it gives user an option that we do _not_ support!

[I've worked on support for local transactions in the past -- except that when I presented the work to @krisv and @mdproctor, the reactions were that we don't need to support local transactions here.. ]

mrietveld:
@psiroky, @michalbali: I think your local master branch may be behind a number of commits? Could you rebase your commit on top of the latest master commits and push force the pull request branch? 

Thanks!

psiroky:
@mrietveld The local branch in behind master for sure, the PR is more that 2 years old.

I was just wondering if we should include this or not. If yes, we should be able to merge and fix locally.

But since you mentioned that the local transactions are not supported for process, then I believe this should be not exposed through the API (constructor). I agree that it would only confuse things.

mrietveld:
@psiroky Okay, close it then. 

(if I was the only one who disagreed, I was willing to let it go. )

@psiroky Thank you very much for going through all of the pull requests!

psiroky:
I am closing this as per previous comments.

@michalbali please reopen this (or create new PR) if you have a use for this or think that it might be generally useful. We can then discuss it further.

PR #157, opened 2013-03-21
WorkItemHandler proper evaluation of parameter expression to object

vtatai:
This is a fix for the problem I described at:

https://community.jboss.org/thread/222708?tstart=0

Basically the issue is that, when an expression resolves to a single arbitrary object, this object is converted to String before being handled to the work item handler. AFAICT this was done because you can have multiple expressions, such as '#{name} - #{description}'. What I've done is, iff there is one expression, the evaluated value is passed directly to the work item handler.

Please notice that I have no idea if this is correct, but all the tests have passed.

BTW in a separate commit I had to change some method calls to drools which were failing to compile, since apparently the method name has changed.


mswiderski:
5.4.x branch is no more maintained so closing it, please submit new PR for latest branch (master)

PR #160, opened 2013-04-05
OpenJPA compatibility

eggeral:
Patches to make human-task-services work with OpenJPA. 

All tests still pass when using Hibernate. Most tests pass when using OpenJPA but there seams to be a problem with Weld and OpenJPA which I could not fix. I will look into that later.

There are two maven profiles: hibernate and openjpa. hibernate is default.

Run mvn clean test -P openjpa to execute the tests using OpenJPA


mswiderski:
this cannot be merged any longer,, too many conflicts closing, feel free to update it to master and submit new PR, thanks

PR #198, opened 2014-04-11
merged changes from task-list-refactor2-search to rebased master

hlund:
merged towards rebased master - query performance tests passed


mswiderski:
this cannot be merged any longer, out of date as it does refer to classes that does not exist any more thanks

PR #202, opened 2014-05-18
JBPM-3916 - Keeping multi instance node (instance) ids consistent

mrietveld:
This solves [JBPM-3916](https://issues.jboss.org/browse/JBPM-3916) as well as some related issues. 

However, with regards to the ids for node instances, it obviously breaks backwards compatibility.. I'm not sure whether or not it's a good idea to merge it for 6.1, thus.. :/


mrietveld:
@krisv FYI: using this code to fix [JBPM-4611](https://issues.jboss.org/browse/JBPM-4611). For 4611, we need (absolutely) unique id's for timers in multi-instance subprocesses, because otherwise there's no way to identiy which NodeInstance (in which multi-instance CompositeNodeInstance) to trigger.

mrietveld:
well, not per se "this code", but the same logic basically..

PR #207, opened 2014-08-28
Added logging to empty catch block

selrahal:
I wanted to add at least warn logging when unable to dispose a runtime engine.

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/tx/DisposeSessionTransactionSynchronization.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/tx/DisposeSessionTransactionSynchronization.java
@@ -18,6 +18,8 @@ package org.jbpm.runtime.manager.impl.tx;
 import org.drools.persistence.OrderedTransactionSynchronization;
 import org.kie.api.runtime.manager.RuntimeEngine;
 import org.kie.api.runtime.manager.RuntimeManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Transaction synchronization that disposed <code>KieSession</code> instance on transaction completion during
@@ -25,6 +27,7 @@ import org.kie.api.runtime.manager.RuntimeManager;
  *
  */
 public class DisposeSessionTransactionSynchronization extends OrderedTransactionSynchronization {
+	private Logger logger = LoggerFactory.getLogger( getClass() );
 
 	private RuntimeEngine runtime;
 	private RuntimeManager manager;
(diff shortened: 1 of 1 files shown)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/tx/DisposeSessionTransactionSynchronization.java
```diff
@@ -42,7 +45,7 @@ public void afterCompletion(int status) {
 	    try {
 	        manager.disposeRuntimeEngine(runtime);
 	    } catch (Throwable e) {
-	        // catch exception as it's only clean up and should not affect runtime
+	        logger.warn("Could not dispose runtime engine", e);
```
mswiderski: could we make it debug instead, there could be number of cases where this will make logs looks bit scary while it should not be seen as any serious issue but might be worth looking into while running in debug, wdyt?

selrahal: but isn't this a serious issue? for singleton strategy disposing is a no-op, but both per_request and per_process are using ThreadLocals to cache the runtimeEngine(s). couldnt a failure in disposal lead to stale runtime engines being used in a thread and heap issues?

mswiderski: in normal cases this should not fail unless invoked multiple times which then could lead into false alarms - like mixing manual dispose with automatic dispose done on transaction completion

selrahal: but isn't mixing manual dispose with automatic dispose something to avoid? I agree that in normal cases it should not fail which is why I initially put it at warn level.

mswiderski: indeed, but then how often we are in situation where users actually do it differently.

selrahal: hopefully not frequently :) I would rather have debug logging than no logging, but I prefer warn level. Should I just update it and move along....

mrietveld: Please do -- in general, there are very few (if any) dispose or cleanup operations that should be warn. We only use warn for operations that directly prevent or impair normal operation. 

If you can push -f (force) a fix to the pull request, that would be great! Thanks.

selrahal: Updated the pr

psiroky:
@mswiderski I believe this can be merged now?

mswiderski:
yup, feel free to merge that in if you have some spare time

psiroky:
Merged into master https://github.com/droolsjbpm/jbpm/commit/b29c2a550ce5b925c56ce9c055ff693b0918c721

PR #238, opened 2015-03-04
fix addition of a csv resource to the knowledge builder

mariofusco:


manstis:
A nice generic fix ;)

You probably want to look at Score Cards too.. IIRC they have a similar configuration class.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/SimpleRuntimeEnvironment.java
```diff
@@ -106,8 +109,17 @@ public void init() {
...
      */
     public void addAsset(Resource resource, ResourceType type) {
-        this.kbuilder.add(resource, type);
-        if (this.kbuilder.hasErrors()) {            
+        if (resource.getSourcePath() != null && resource.getSourcePath().endsWith(".csv")) {
+            Properties prop = new Properties();
+            prop.setProperty(ResourceTypeImpl.KIE_RESOURCE_CONF_CLASS, DecisionTableConfigurationImpl.class.getName());
+            prop.setProperty(DecisionTableConfigurationImpl.DROOLS_DT_TYPE, DecisionTableInputType.CSV.toString());
+            ResourceConfiguration conf = ResourceTypeImpl.fromProperties(prop);
+            this.kbuilder.add(resource, type, conf);
```
mswiderski: can't we handle that in kbuilder itself? Reason I ask is that other users might run into exactly same issue when not using RuntimeManager API

mrietveld:
https://github.com/droolsjbpm/jbpm/pull/239 has replaced this.

PR #239, opened 2015-03-06
SWITCHYARD-2583, BZ-1199256 Switchyard component can not tell difference CSV/XLS

mrietveld:


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/SimpleRuntimeEnvironment.java
```diff
@@ -106,8 +109,30 @@ public void init() {
      * @param type type of the asset
      */
     public void addAsset(Resource resource, ResourceType type) {
-        this.kbuilder.add(resource, type);
-        if (this.kbuilder.hasErrors()) {            
+        if (resource.getSourcePath() != null ) { 
+            String path = resource.getSourcePath();
+           
+            String typeStr = null;
+            if( path.endsWith(".csv") ) {
```
manstis: It might be better to check the file extension against org.kie.api.io.ResourceType.DTABLE before building a DecisionTableConfigurationImpl. Furthermore you could use the DTABLE.defaultExtension to differentiate between XLS and CSV... Just my 2c... saves hard coding file extensions when KIE has a Registry of them already....

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/SimpleRuntimeEnvironment.java
```diff
@@ -106,8 +109,30 @@ public void init() {
...
+            } else if( path.endsWith(".xls") ) { 
+                typeStr = DecisionTableInputType.XLS.toString();
+            } 
+            
+            if( typeStr != null ) { 
+                Properties prop = new Properties();
+                prop.setProperty(ResourceTypeImpl.KIE_RESOURCE_CONF_CLASS, DecisionTableConfigurationImpl.class.getName());
+                prop.setProperty(DecisionTableConfigurationImpl.DROOLS_DT_TYPE, typeStr);
+                ResourceConfiguration conf = ResourceTypeImpl.fromProperties(prop);
+                this.kbuilder.add(resource, type, conf);
```
mswiderski: I'll repeat my comment here as the other pr was closed:
can't we handle that in kbuilder itself? Reason I ask is that other users might run into exactly same issue when not using RuntimeManager API

or this comment does not make sense at all... I might be missing some context..

mrietveld:
I've merged this to master. This is a *temporary* fix because of bXms and SwitchYard/Fuse deadlines. 

Ideally, I think that everyone agrees that this should probably be fixed in the KBuilder interface and implementations.

PR #241, opened 2015-03-19
JBPM-4579 fix + some other Wiser configuration cleanup in other tests.

baldimir:
See JIRA: https://issues.jboss.org/browse/JBPM-4579.
+ contains some small changes in Wiser configuration in other tests. The properties that were there used are not needed.

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesBaseTest.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesBaseTest.java
@@ -64,15 +64,9 @@ public abstract class EmailDeadlinesBaseTest extends HumanTaskServicesBaseTest {
     
     
     public void setup() {
-        Properties conf = new Properties();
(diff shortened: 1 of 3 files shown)


mswiderski:
why would you like to remove the use of properties to setup mail server? 
There are cases where this is rather important e.g. continuous integration environment where multiple jobs can run in parallel and use of con files can be pregenerated to avoid port conflicts.

baldimir:
I didn't thought about that that way. Maybe this should be more obvious from the code. I.e. I didn't find any file named email.conf in the project. There should be some reference file even if the file is generated at runtime i.e. by CI. The other cases ditto. The properties are instantiated in setup() method and only host and port are used from the properties. What do you think? If it is as you say I will revert the changes and leave only the fix for the test.

mswiderski:
I'd prefer to leave it based on properties just to be on safe side for backward compatibility as I believe it won't hurt. When it comes to the file itself, it should be named drools.email.conf as it goes via ChainedProperties class.

So my vote is to include the fixes but still use properties.
Thanks

baldimir:
Ok I made changes as we agreed on IRC.

mswiderski:
squashed commits and merged into master, thanks

PR #242, opened 2015-03-20
Fix some issues from Sonar report

livthomas:
I have fixed some issues discovered by Sonar. I tried to avoid fixing the ones where I was unsure about side effects or developer intentions like catching Throwable or commented code which may be needed in the future. So I hope these my fixes should not cause any problems.


mswiderski:
merged into master, thanks

PR #243, opened 2015-03-23
JBPM-4586 - Removal of loops with calls to Thread.sleep().

baldimir:
See JIRA: https://issues.jboss.org/browse/JBPM-4586

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/StartEventTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/StartEventTest.java
@@ -54,6 +54,13 @@ import org.slf4j.LoggerFactory;
 @RunWith(Parameterized.class)
 public class StartEventTest extends JbpmBpmn2TestCase {
 
+    // Defines extra sleep time required in tests. This time is added to the standard sleep time used in some tests
+    // to make sure that tested events passed or tested threads finished.
+    // I.e. test uses timer that generates some event each second and checks if 5 events were generated in 5 seconds time.
+    // Because of JVM nature, one second takes not exactly 1000ms so for waiting Thread.sleep(5000 + EXTRA_SLEEP_TIME)
+    // call is used in the test.
+    private static final Integer EXTRA_SLEEP_TIME = 250;
+
     @Parameters
     public static Collection<Object[]> persistence() {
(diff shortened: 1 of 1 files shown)


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/StartEventTest.java
```diff
@@ -424,9 +425,9 @@ public void afterProcessStarted(ProcessStartedEvent event) {
         });
 
         assertEquals(0, list.size());
-        for (int i = 0; i < 2; i++) {
-            Thread.sleep(1200);
-        }
+        // Timer in the process takes 1000ms, so after 2 seconds, there should be 2 process IDs in the list.
+        // 2100ms is here because it is important to be sure that 2 seconds have passed.
+        Thread.sleep(2100);
```
psiroky: Based on the previous code, shouldn't the wait be 2500ms? Or maybe the 1200ms is wrong and the process indeed takes 1000ms?

baldimir: I made some changes to this. Take a look if it is OK. The main difference is that I created a constant that defines extra time needed for Thread.sleep() calls so the sleep time should be easily configurable in the future. Now it is 250ms. I think that this should be enough now to cover all tests that are in StartEventTest class because now there are no loops with Thread.sleep().

mswiderski:
just tried to merge this but several builds fails with this:
Failed tests:
  StartEventTest.testTimerStartCron:227->Assert.assertEquals:542->Assert.assertEquals:555->Assert.assertEquals:118->Assert.failNotEquals:743->Assert.fail:88 expected:<5> but was:<6>
  StartEventTest.testTimerStartCron:227->Assert.assertEquals:542->Assert.assertEquals:555->Assert.assertEquals:118->Assert.failNotEquals:743->Assert.fail:88 expected:<5> but was:<6>

baldimir:
Hmm... Strange.  I tested it several times. OK I will recheck it on Monday.

baldimir:
Changed the cron test so it waits approximately the same time as it was before (5s as was before, but without Thread.sleep in cycle). Tested it again, it should work.

mswiderski:
merged into master, thanks @baldimir !

PR #245, opened 2015-03-25
Fix parent for jbpm-executor-ejb

triceo:
jbpm-executor-ejb references jbpm-services as it's parent, skipping one level in the hierarchy. This pull request fixes that.
(Discovered this due to master not building for me on my local machine.)


mswiderski:
thanks Lukas, forgot to updated that after moving the module to ejb modules.

PR #248, opened 2015-04-20
JBPM-4611 Duplicate Timer calls by a Boundary Timer inside a MultipleIns...

tkobayas:
...tances

- Test case


mrietveld:
@tkobayas Thanks for the test case -- I took your test and used the same logic to make a test in the jbpm-bpmn2 module. 

See https://github.com/droolsjbpm/jbpm/commit/486103f4cf0be2bdd86037480436c75887162fcc#diff-17c30fd4f13b8467b4eb332d0bfedfc3R2036

PR #249, opened 2015-04-22
JBPM-4620 - Enhanced functionality for ServiceTaskHandler

jomarko:
Full qualified name of interface to be instantiated, will be ServiceTaskHandler search  in two parameters of WorkItem. Interface and interfaceImplementationRef.

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/handler/ServiceTaskHandler.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/handler/ServiceTaskHandler.java
@@ -33,11 +33,26 @@ public class ServiceTaskHandler implements WorkItemHandler {
     
     public void executeWorkItem(WorkItem workItem, WorkItemManager manager) {
         String service = (String) workItem.getParameter("Interface");
+        String interfaceImplementationRef = (String) workItem.getParameter("interfaceImplementationRef"); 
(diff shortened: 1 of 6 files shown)


mswiderski:
Jozef,

better if you could apply this change to this handler [1] instead as it is the one being registered by default

[1] https://github.com/droolsjbpm/jbpm/blob/master/jbpm-workitems/src/main/java/org/jbpm/process/workitem/bpmn2/ServiceTaskHandler.java

jomarko:
@mswiderski , should I stash changes which I have added to [1]? Or should be these changes kept in both ServiceTaskHandlers?

[1] https://github.com/jomarko/jbpm/blob/JBPM-4620/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/handler/ServiceTaskHandler.java

mswiderski:
in long run we should have only one handler by default for Service Task but if you prefer feel free to add that logic in both. One in bpmn2 module is bit lighter as it does not provide web service support so still might be of use for some cases

jomarko:
I will refactor tests, because they differ only in used ServiceTaskHandler.

mswiderski:
merged into master, thanks @jomarko 

P.S.
just noticed that the jira was resolved even though PR was not yet merged, would be better to keep it open until PR is merged so it won't be missed.

PR #251, opened 2015-04-23
Release resources in case of exception thrown in tearDown method

livthomas:
If an exception is thrown in tearDown() method of JbpmJUnitBaseTestCase, the data source is not closed properly. This influences other tests run in the same batch and they fail with the following exception:

Caused by: java.lang.IllegalStateException: A resource with uniqueName 'jdbc/jbpm-ds' has already been registered. Cannot register XAResourceProducer 'a PoolingDataSource containing an XAPool of resource jdbc/jbpm-ds with 5 connection(s) (5 still available)'.

So I've put cleaning methods to try block and resources are released in its finally block. This is just a simple fix. If you think anything else needs to be fixed regarding this issue, just let me know and I'll push another commit.


mswiderski:
same here: I am in the middle of something but will merge that in once done, so assigning to myself

mswiderski:
merged into master, thanks @livthomas

PR #252, opened 2015-04-28
Add more tests for LDAPUserGroupCallbackImpl and LDAPUserInfoImpl

livthomas:
I have moved some tests from QE test suite that can be used as unit tests for LDAPUserGroupCallbackImpl and LDAPUserInfoImpl. Besides moving the existing tests, I have also added some new test cases for IS_ENTITY_ID_DN=true and it seems that getMembersForGroup and hasEmail from LDAPUserInfoImpl ignore this property and do not work correctly when you pass a group with DN as a parameter. Is this an intended behavior? See 4 ignored tests for more details.


mswiderski:
pushed to master and fixed the issue for test cases that were ignored as part of JBPM-4565, thanks @livthomas

PR #254, opened 2015-05-19
Change impact

mrietveld:
DO NOT MERGE THIS PR!

I made this PR so that @mswiderski can take a look at the code and give feedback, if he wants to. :)

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
@@ -19,7 +19,9 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 
 import org.drools.compiler.builder.impl.KnowledgeBuilderConfigurationImpl;
 import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
@@ -28,14 +30,25 @@ import org.drools.compiler.compiler.BPMN2ProcessProvider;
 import org.drools.core.io.impl.ByteArrayResource;
 import org.drools.core.util.StringUtils;
 import org.jbpm.bpmn2.xml.BPMNDISemanticModule;
-import org.jbpm.bpmn2.xml.BPMNExtensionsSemanticModule;
+import org.jbpm.kie.services.impl.bpmn2.builder.DataServiceExpressionBuilder;
+import org.jbpm.kie.services.impl.bpmn2.builder.dialect.ThreadLocalAbstractBuilderFacade;
+import org.jbpm.kie.services.impl.bpmn2.builder.dialect.java.DataServiceJavaProcessDialect;
+import org.jbpm.kie.services.impl.bpmn2.builder.dialect.mvel.DataServiceMvelProcessDialect;
 import org.jbpm.kie.services.impl.model.ProcessAssetDesc;
+import org.jbpm.process.builder.ActionBuilder;
+import org.jbpm.process.builder.AssignmentBuilder;
+import org.jbpm.process.builder.ProcessClassBuilder;
+import org.jbpm.process.builder.ReturnValueEvaluatorBuilder;
+import org.jbpm.process.builder.dialect.ProcessDialect;
+import org.jbpm.process.builder.dialect.ProcessDialectRegistry;
 import org.jbpm.services.api.DefinitionService;
 import org.jbpm.services.api.DeploymentEvent;
 import org.jbpm.services.api.DeploymentEventListener;
 import org.jbpm.services.api.model.ProcessDefinition;
 import org.jbpm.services.api.model.UserTaskDefinition;
+import org.kie.api.definition.process.Process;
 import org.kie.api.io.ResourceType;
+import org.kie.api.runtime.KieContainer;
 import org.kie.internal.builder.KnowledgeBuilder;
 import org.kie.internal.builder.KnowledgeBuilderError;
 import org.kie.internal.builder.KnowledgeBuilderFactory;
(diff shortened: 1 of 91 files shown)


mrietveld:
@mswiderski With regards to the local-thread architecture for extracting class info from scripts/expr's in BPMN2 resources, it's a little complicated. 

I've explained it the best I can in the [org.jbpm.kie.services.impl.bpmn2.builder.dialect.ThreadLocalAbstractBuilderFacade](https://github.com/mrietveld/jbpm/blob/e01c58cc26ac07d8d7993c9a3fdfd9f93d199912/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/builder/dialect/ThreadLocalAbstractBuilderFacade.java) class.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -47,17 +60,32 @@
...
+    private static final BPMN2DataServiceExtensionSemanticModule EXTENSIONS_MODULE = new BPMN2DataServiceExtensionSemanticModule(MODULE);
     
-    private ConcurrentHashMap<String, Map<String, ProcessDescRepoHelper>> definitionCache = 
+    private ConcurrentMap<String, Map<String, ProcessDescRepoHelper>> definitionCache = 
     		new ConcurrentHashMap<String, Map<String, ProcessDescRepoHelper>>();
+   
+    private static final String [] dialectNames = { "java", "mvel" };
+    private static final ProcessDialect [] dataServiceDialects = { 
+            new DataServiceJavaProcessDialect(),
+            new DataServiceMvelProcessDialect()
```
mswiderski: these are just extensions to the default ones, meaning they provide same capabilities so it's safe to be replaced as it affects overall parsing of the processes as it's static stuff?

Was thinking if we can use just one implementation to avoid this extensions in services. Though it might be too big change right now... ideally it would be good to have this "analysis" done on core engine level and provide that to the services somehow - kbase model repository or so... just an idea for future

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -132,14 +206,113 @@ public ProcessDefinition buildProcessDefinition(String deploymentId,String bpmn2
         } finally {
         	BPMN2ProcessFactory.setBPMN2ProcessProvider(originalProvider);
             BPMN2DataServiceSemanticModule.dispose();
+            // BPMN2DataServiceExtensionSemanticModule.dispose() <-- not needed because the modules share the same ThreadLocal instance
```
mswiderski: so we don't need to remove anything from thread local any more? It can be safely reused, right

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/ProcessGetBusinessRuleHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import org.jbpm.bpmn2.xml.ImportHandler;
+import org.jbpm.workflow.core.Node;
+import org.jbpm.workflow.core.node.RuleSetNode;
+import org.w3c.dom.Element;
+import org.xml.sax.Attributes;
+import org.xml.sax.SAXException;
+
+/**
+ * This handler adds classes imported (via the &gt;extensionElement&lt;) to the list
+ * of referenced classes.
```
mswiderski: I guess the docs should refer to rule flow groups instead of classes imported?

mswiderski:
good piece of work Marco!

Main thing to discuss and eventually address is the use of ThreadLocal as initially it was added to compensate the way we handle it on engine side to parse - using static providers that can be only one. Since we replaced it (temporarily) we had to use thread local. But the more I think about it the more I believe we should address that at the root of the issue - core engine parser. 
As mentioned in one of the comments I'd like to have this information gathering on parsing time and then make it available via some kbase methods. That way we could:
- avoid thread locals in services
- avoid duplicated parsing - which I believe is happening now
- avoid various extensions e.g. builders

wdyt?
It might not be good time to do it now but most certainly for v7 it should be addressed to simplify this code

mrietveld:
@mswiderski I agree -- at the moment we're parsing BPMN2 resources twice, which is a shame. 

We unfortunately do not have the time do it now, but v7 it should be addressed: part of the reason that we don't have time to do this now is that we will have to develop internal data structures to hold this information, and we will have to think carefully about that because of memory leaks which can quickly become an issue when modifying the kbase, I think. 

However, I'm glad to hear you bring this up -- the same thoughts occurred to me as well. :)

mswiderski:
agreed, let's put a jira task for this work so we can address it as soon as we enter v7 work. 

+1 on careful thinking for the data structure and in addition how we are going to clean it up as in general we can add/remove processes in kbase so that needs to be covered too.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
mrietveld: Yes -- well, sort of. In the builders that the dialects provide, I was only planning on extracting information and skipping the actual compilation of the expressions/scripts (to save time). However, it's very easy to extend them to make sure that the expressions/scripts are compiled and added.

(So for v7, it's easy to have one class that also extracts the info, depending on a flag or whatever.. )

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
mrietveld: Ah, no in this case, the BPMN2DataServiceSemanticModule.dispose() does the cleanup for both modules.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -132,14 +206,113 @@ public ProcessDefinition buildProcessDefinition(String deploymentId,String bpmn2
...
+            
+            // reset dialects
+            resetDialectExpressionBuilders();
```
mrietveld: @mswiderski this does the clean up for the Dialect  code.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/ProcessGetBusinessRuleHandler.java
mrietveld: Oops, yup. good catch!

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
mswiderski: reason I asked is what will happen if at the same time (different request/threas) will attempt to parse different processes. Let's say Thread A is extracting information about processes for the sake of services (DefinitionService) while Thread B it simply building kbase with other process. Since we replaced the statically  dialects could it be that Thread B will not compile the expressions?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
mswiderski: cool, missed the other module so we are good here, thanks

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
mrietveld: Ah, yes. :) 

See the javadoc here: https://github.com/droolsjbpm/jbpm/pull/254/files#diff-6e6dc1ba9a9f5984b8dc659bd08ee925R17

Basically, there is a layered structure of ThreadLocal instances. The ThreadLocal in the data service dialects switches between the singleton instance of the (normal) ActionBuilder (and other builder) and the singleton instance of the data service action builder. Within the data service action builder, there's _another_ ThreadLocal that refers to the thread local process description repo helper instance..

mswiderski:
closing this one as #255 was merged into master

PR #255, opened 2015-05-27
Change impact: bpmn2 meta data

mrietveld:
Maciej, 

This is the finished work that you saw earlier. The purpose is to extend the DefinitionService so that it would return information that we need in the BPMN2 indexer for change impact. 

The important information in this case are 
- referenced java classes
- referenced rules

There is still some edge cases (classes in constraint rules in bpmn2 defs) that are not covered yet, but I have a test case for that, in any case. 

Because I'm going on PTO (from the 30th until the 22rd), and because the functionality is 98% done, I wanted to submit the PR now. 

Thanks!
Marco

### DIFF
--- a/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/DefinitionService.java
+++ b/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/DefinitionService.java
@@ -21,6 +21,7 @@ import java.util.Map;
 
 import org.jbpm.services.api.model.ProcessDefinition;
 import org.jbpm.services.api.model.UserTaskDefinition;
+import org.kie.api.runtime.KieContainer;
 
 /**
  * Provides details from definition point of view which is extracted from BPMN2 definitions.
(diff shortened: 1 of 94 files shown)


mswiderski:
Marco,

shall I merge that or it's more for review and it shall wait until you're
back from PTO?

2015-05-27 20:45 GMT+02:00 Marco Rietveld <notifications@github.com>:

mrietveld:
Please merge it -- unless you see a problem with it. 

I've run the full jbpm build with it. I didn't find anything in jbpm-console-ng that would be impacted because of the new methods on the DefinitionService. 

Thanks!

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/DefinitionService.java
```diff
@@ -41,13 +42,14 @@
...
-	 * @param classLoader class loader that should be used while parsing the BPMN2 in case custom classes are referenced
+	 * @param kieContainer the {@link KieContainer} instance that contains the deployment project: this should be used when 
+	 *          parsing the BPMN2 in case custom classes or other project resources (processes, rules) are referenced
 	 * @param cache indicates if the definition service should cache this <code>ProcessDefinition</code>
 	 * @return fully populated <code>ProcessDefinition</code>
 	 * @throws IllegalArgumentException in case build operation cannot be completed
 	 */
 	ProcessDefinition buildProcessDefinition(String deploymentId, String bpmn2Content,
-			ClassLoader classLoader, boolean cache) throws IllegalArgumentException;
+			KieContainer kieContainer, boolean cache) throws IllegalArgumentException;
```
mswiderski: what is the reason to change classLoader to kieContainer? Asking mainly as this changes the api interface that can be used by users since it is public api for jbpm services - used for example by ejb. Most likely it's not so frequently used as standalone so most likely not big impact but the bigger issue might be that we expect users to always have kieContainer to use this method. While previously it was completely kieContainer agnostic.

mswiderski:
merged in master

mrietveld:
@mswiderski Thank you for merging! 

I look forward to fixing this at a deeper level (the changes you talked about for 7.0 where this functionality is basically done by default).

PR #257, opened 2015-06-04
BigDecimals are immutable

selrahal:


mswiderski:
merged into master, thanks @selrahal

PR #258, opened 2015-06-05
JBPM-4670 Class TransactionLockInterceptor not found exception

cybermaniax:


mswiderski:
merged into master, thanks @cybermaniax

PR #260, opened 2015-06-18
BPMSPL-111 - Add test coverage of task update events

livthomas:
I have written some tests for task update events. Some of them are marked as ignored since they fail and I am not sure about those use cases. I think that either it should not be allowed to pass null values or the behavior should be consistent so if you change some value to null this action will also be stored in task events. Now, the value is changed but there is no task event record for that.


mswiderski:
merged and fixed the inconsistencies, thanks @livthomas

PR #263, opened 2015-06-23
BZ-1234592 - Race condition with multiple job executor threads on Oracle

mswiderski:
for review only, please do not merge


mswiderski:
closing this as tests proved that it does not resolve the problems in oracle so it needs to be discarded

PR #265, opened 2015-06-29
Jbpm 4426

akoskm:
The test which was added for JBPM-4426 to ensure that `TaskCleanUpProcessEventListener` works fine uses `SINGLETON` strategy.

This pull request contains a unit test which uses `PROCESS_INSTANCE` strategy and breaks JBPM-4426. I wasn't able to detect what causes the problem so I ignored the test for now.

All I was able to figure out is that `WorkflowProcessInstanceImpl.setState` fires `fireAfterProcessCompleted` for both processes. 

The problem is that in `ProcessEventSupport.fireAfterProcessCompleted(ProcessInstance instance, KnowledgeRuntime kruntime)` the iterator `Iterator iter = this.getEventListenersIterator();` doesn't contain the `TaskCleanUpProcessEventListener`.

### DIFF
--- a/jbpm-test/src/test/java/org/jbpm/test/tasks/admin/AdminAPIsWithListenerTest.java
+++ b/jbpm-test/src/test/java/org/jbpm/test/tasks/admin/AdminAPIsWithListenerTest.java
@@ -24,16 +24,23 @@ import javax.persistence.EntityManager;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.Persistence;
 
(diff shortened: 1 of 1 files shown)


akoskm:
For PROCESS_INSTANCE strategy the TaskCleanUpProcessEventListener should be registered before the RuntimeEngine is retrieved. Test case fixed.

mswiderski:
commits squashed and merged into master, thanks @akoskm

PR #267, opened 2015-06-30
JBPM-4147 Able to use 'Task Variables' as expressions in Task properties

baldimir:
Modified existing test so it tests variables from issue description + new test.


mswiderski:
the sessionId works fine when it comes to task variables but it is not loaded by queries such as tasks assigned as potential owners and thus it's not set on TaskSummary. You would need to load the actual Task instance instead of summary to assert on session id. The reason it is not loaded is due to session id is not relevant in most of the cases.

baldimir:
Ok, thanks for help.. will do that.

mswiderski:
shall we keep this pr open for the changes or?

baldimir:
Yes please, I will write a comment when it is done.

baldimir:
Ok now I think there's all I wanted to add.

mswiderski:
commits squashed and merged into master, thanks @baldimir

PR #268, opened 2015-06-30
Added deploymentId to work item of throwing message event

btison:


mswiderski:
added test cases and jira into commit message and merged into master, thanks @btison

PR #269, opened 2015-07-02
typo mistake fix

jomarko:


mswiderski:
@jomarko I merged it though with small change, see here https://github.com/droolsjbpm/jbpm/commit/23f8a0f2ea6cbd4033e839e4262596b91a655612

since this method works with Process and not ProcessInstances it does not make sense to cast to ProcessInstance of any form. Does it make sense?

jomarko:
@mswiderski I am fine with it. Thanks.

PR #271, opened 2015-07-06
JBPM-4691 name() for fluent API CompositeNode is missing

tkobayas:
Added name() method to CompositeNodeFactory.


mrietveld:
Thanks, @tkobayas!

PR #272, opened 2015-07-07
JBPM-4653 - Create test coverage module and move tests from QE regression test suite

livthomas:
After we decided to move most of QE tests to community repositories, we agreed on creating a new module which will include only tests. The tests from jbpm-test module should also be moved to this one since jbpm-test should have only one responsibility - to support testing.

Right now it includes tests that were part of QE test suite. Some of them are ignored and they still need to be reviewed. I am sending a pull request even though they are not fixed because we need this module right now for test development. We will look at the ignored tests later.

Sometimes there is a problem with Executor Service tests. If you see any random fail, either try to fix it or mark it as ignored and I will look at it later.


mswiderski:
started to look at this and if all goes well should be merged tomorrow. @livthomas will ping you in case of questions on irc

psiroky:
@mswiderski @livthomas Do you plan to include this module into the standard maven build (e.g. it will be executed whenever some simply calls "mvn clean install")? Or will it be disabled by default and only ran on demand? How long does it take to run the tests?

mswiderski:
I was planning to include it only with full build - what's your opinion?

It takes around 11 min for me - only the jbpm-test-coverage module.

psiroky:
I was thinking about not running the tests as part of the standard (or full) build, but rather just make sure the tests compile and run them after the whole jbpm repo build completes. Basically divide the the unit tests and integration tests execution. Unless all the unit tests pass, the integration tests would not be executed. This would I think create better feedback loop, as we would have some initial results much faster (when not running the integration tests). WDYT?

On (usually) slower CI machines it will likely take 2-3 times longer.

mswiderski:
sounds very good to me.

@psiroky Do you mind adding that configuration on top of this branch? I still didn't have much time to work on this PR. Just executed it few times and confirm what @livthomas mentioned about random failures.

mswiderski:
merged into master with small improvements that made the test work well locally. Though until changes made by @psiroky are not in I commented out the module so it's not built with entire project. Closing this PR, thanks a lot @livthomas

psiroky:
@mswiderski thanks, I will shortly submit PR with the build changes.

PR #273, opened 2015-07-08
BPMSPL-128 - Add test-coverage for JavaScript support

domhanak:
I have written tests for JavaScript support in core engine, however I can't compile JavaScript expressions as it is done in Java and MVEL tests as drools does not provide JavaScriptDialect that can compile expression. Thanks to that I can't correctly test  globals  with JavaScript.


mswiderski:
fixed the issue with globals and merged (after squashing) tests, thanks @domhanak

PR #274, opened 2015-07-09
BZ1241358 SingleSessionCommandService Interceptor stack is lost when …

tkobayas:
…rollback

- test case


mswiderski:
fix applied in drools, test case from this PR merged with small enhancements to check all interceptors and their order. thanks again @tkobayas

PR #275, opened 2015-07-09
BZ1240215 Create a BoundaryEvent with fluent API

tkobayas:
- Test case and fix. Tested only boundaryTimer case

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ProcessFactoryTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ProcessFactoryTest.java
@@ -17,6 +17,7 @@
 package org.jbpm.bpmn2;
 
 import org.jbpm.bpmn2.xml.XmlBPMNProcessDumper;
+import org.jbpm.persistence.session.objects.TestWorkItemHandler;
 import org.jbpm.ruleflow.core.RuleFlowProcess;
 import org.jbpm.ruleflow.core.RuleFlowProcessFactory;
 import org.junit.Test;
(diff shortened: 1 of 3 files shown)


mswiderski:
@tkobayas is this complete solution? I see only timerCycle support in there or is it more for sharing idea and that needs to some more work?

tkobayas:
Hi Maciej, sorry about the lack of explanation.

I was expecting that BoundaryEventNodeFactory.eventType() and BoundaryEventNodeFactory.setMetaData() work for general use cases. timeCycle() is a convenient/fluent method for BoundaryTimerEvent case. But yes, only one timeCycle() looks odd so I have added some more fluent methods in the new commit.

However, this is not yet complete in terms of all BoundaryEvent use cases. Actually, I tried to implement BoundaryErrorEvent use case but it would require many changes. For example, RuleFlowProcessFactory should hold org.jbpm.bpmn2.core.Definitions for error definitions so jbpm-flow will depends on jbpm-bpmn2 project. Is it the right way to go?

Probably we may merge the commit only for BoundaryEventTimer use case and file a JIRA for other BoundaryEvent use cases for future enhancement. WDYT?

mswiderski:
sounds good to me, will merge it today (hopefully) and feel free to open another jira for further enhancement. Anyway we cannot support all possible bpmn2 constructs via fluent api I believe as it might be rather extensive work effort ... so gradually improving it might be the way to go.

Thanks Toshiya!

mswiderski:
merged into master, thanks

tkobayas:
Thanks Maciej,

Filed a JIRA https://issues.jboss.org/browse/JBPM-4696

PR #276, opened 2015-07-10
[JBPM-4653] run jbpm-test-coverage tests in specific profile

psiroky:
* the tests are only compiled during the standard build. This
   is to make sure that compilation failures are detected as
   early as possible.

 * the tests can be executed using "-Pintegration-tests" or
   "-Dintegration-tests"

@livthomas please let me know if you have some concerns about this.

### DIFF
--- a/jbpm-test-coverage/pom.xml
+++ b/jbpm-test-coverage/pom.xml
@@ -70,6 +70,11 @@
     <maven.jdbc.db.port/>
     <maven.jdbc.db.server/>
     <maven.jdbc.url>jdbc:h2:mem:test;MVCC=true</maven.jdbc.url>
+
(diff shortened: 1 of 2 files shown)


livthomas:
@psiroky I think separating these tests from unit tests is a good way to go. However, we also plan to move all tests from jbpm-test module to jbpm-test-coverage. Is it alright that these tests will not be run with unit tests?

mswiderski:
I am fine to have both running in the integrations-tests profile (both tests from QE being migrated now and those that will come from jbpm-test). That will speed up a build and I'd like to have one jenkins job setup to run daily with integration tests. So if we agree with this +1 to merge this

psiroky:
I definitely agree with moving the tests from jbpm-test into this module, and thus removing them from the first (unit tests) build.

PR #277, opened 2015-07-13
JBPM-4621 Async continuation test coverage

ibek:


mswiderski:
merged into master, thanks @ibek

PR #278, opened 2015-07-14
JBPM-4516 Async throwing signal events test coverage

ibek:


mswiderski:
@ibek when running these tests as maven build two of them are failing:
- AsyncIntermediateCatchSignalTest.testCorrectProcessStateAfterExceptionSignalCommandMulti:115 expected:<null> but was:<WorkflowProcessInstance5 [processId=org.jbpm.test.functional.async.AsyncIntermediateCatchSignal,state=1]>
- AsyncThrowSignalTest.testCorrectProcessStateAfterExceptionThrowSignal:78 expected:<null> but was:<WorkflowProcessInstance1 [processId=org.jbpm.test.functional.async.AsyncIntermediateCatchSignal,state=1]>

it could be related to the Thread.sleep as it's not reliable. Especially if it does not fail for you. Do you mind taking a look at them?

mrietveld:
-10 for tests that use Thread.sleep(..). In my experience, they always eventually end up giving false negatives (or positives!). 

If possible, please look into using the java.util.concurrent classes (maybe with a task node that uses a custom `WorkItemHandler`) or otherwise, if you're up to it, use byteman. Byteman is _awesome_!

ibek:
Alright, I made the changes not to use the thread sleep and I changed also the previous async task tests.

@mswiderski please try the tests again with the changes, I experienced the same problems but they disappeared (not by extending sleep time). Not sure whether it might be caused by the first call of sync signal (testSyncGlobalSignal) and to somehow influence the executor service or anything else.

mrietveld:
Thank you, Ivo (no thread.sleep()).

mswiderski:
all looks good now, merged into master, thanks @ibek

PR #279, opened 2015-07-15
JBPM-4679 - Added test for negative and composite retry delays.

sutaakar:
Test verifying executor behaviour for executions with negative retry delay, for delay defined by combination of time values.


mswiderski:
merged into master, thanks @sutaakar

PR #280, opened 2015-07-18
Get PersistenceStrategy from Environment

carrchang:


mrietveld:
+1

@mswiderski Any objections to this?

mswiderski:
as long as PersistenceStrategy can be used by multiple threads at the same time, then I am ok with it. Environment is shared across all threads running same ksession so that needs to be taken into account.

carrchang:
So, I can't get AuditService properly from org.kie.api.runtime.manager.RuntimeEngine#getAuditService when I using jBPM with Spring.  

@mswiderski You mean I need to get AuditService by new an org.jbpm.process.audit.JPAAuditLogService#JPAAuditLogService(javax.persistence.EntityManagerFactory, org.jbpm.process.audit.strategy.PersistenceStrategyType), or your advice is?

mswiderski:
what I meant is that as the PersistenceStrategy can have different implementations can we guarantee it is safe to be used in multi threaded environment? Reason I ask this is that Environment instance is shared between all threads using same ksession and that might become an issue in certain scenarios - especially CMT based.
@mrietveld any concerns with our out of the box implementations of the strategies?

mrietveld:
@mswiderski Thanks for catching that: I hadn't seen the fact that this meant that `PersistenceStrategy` implementations could be shared -- that is definitely something we do **not** want. 

I don't have a problem with an environmental variable that allows the user to specify the `PersistenceStrategy` type to be used -- maybe in a `PersistenceStrategyFactory` class or something like that?

mswiderski:
+1 on specifying the type instead of actual implementation inside the environment. Or factory if we have support for this.

PR #281, opened 2015-07-20
BPMSPL-132 - Add signal scope tests

livthomas:
I have written some tests for a new signal scope feature. There are two test failures:

1)
SignalEventTest.testSignalIntermediateThrowStartCatchPerProcessInstanceStrategyKieSessionScope:269->testSignalIntermediateThrowStartCatch:241 
Expected size:<1> but was:<2> in:
<[Process 'org.jbpm.test.functional.event.Signal-intermediateThrow-default' [1], Process 'org.jbpm.test.functional.event.Signal-startCatch' [2]]>

I think if a signal with session scope is thrown in KieSession created with PerProcessInstance strategy and there is another process definition with a signal start event, this should not be started since only one process instance can be run in this KieSession. The other process is probably started in a different session because there will be exception otherwise. I am not really sure if this is a valid test case.

2)
SignalEventTest.testSignalSubProcessPerProcessInstanceStrategyRuntimeManagerScope:353->testSignalSubProcess:325 expected:<[2]> but was:<[1]>

When a signal with project scope is thrown in a subprocess, it is not caught by boundary event if per process instance strategy is used. However, it is caught by another process instance. This is very strange behavior and I think it is a bug.


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/event/SignalEventTest.java
```diff
@@ -0,0 +1,361 @@
...
+import org.jbpm.test.JbpmTestCase;
+import org.junit.Test;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.manager.Context;
+import org.kie.api.runtime.manager.audit.ProcessInstanceLog;
+import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.internal.runtime.manager.context.EmptyContext;
+import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
+
+public class SignalEventTest extends JbpmTestCase {
```
triceo: Looking at this test, it could've been made a lot shorter if ```Strategy``` and ```Scope``` were a parameter of the test class, see ```@Parameterized``` in JUnit.

livthomas:
@triceo Thanks for your suggestion. I was considering it at the beginning but it was not possible since there were many different tests. I put them into this unified form at the end but forgot about the parametrization. I will make those changes.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/event/SignalEventTest.java
```diff
@@ -0,0 +1,274 @@
...
+
+    private final Strategy strategy;
+    private final Scope scope;
+
+    public SignalEventTest(Strategy strategy, Scope scope) {
+        this.strategy = strategy;
+        this.scope = scope;
+    }
+
+    @Parameterized.Parameters
```
triceo: ```Parameters``` has a ```name``` attribute - I suggest you use it to describe what each parameter means.

livthomas: Done. Thank you.

mswiderski:
merged and issues fixed, thanks @livthomas @triceo

PR #282, opened 2015-07-21
BPMSPL-128 - Modify ScriptTaskTest to test JavaScript usage

domhanak:
Test currently fails on line 78: Where actual result of tested variable does not match the expected result.
org.junit.ComparisonFailure: 
Expected :"JavaScript Node: Vandrovec"
Actual : {}

This happens because process variables aren't accessible from JavaScript code inside ScriptTask. For example person. They can be accessed using kcontext.getVariable() but it is not possible to directly access them, like  it is done  in MVEL or Java scripts.
Accessing them throws  ReferenceError. 

Test should run correctly once process variables are accessible from JavaScript.


mswiderski:
merged into master and issue fixed with direct access to process variables from java script action and evaluator, thanks @domhanak

PR #284, opened 2015-07-26
BPMPL-165 - migrating to the JPA Criteria API and impoving the Jbpm Query (REST) API

mrietveld:
# _DO NOT MERGE THIS PR!_

@krisv, @mswiderski, @Salaboy: feel free to look through this code. I will be pushing it later this week. 

Comments, questions and (constructive!) criticism are always welcome! Feel free to post them on github.


mrietveld:
@krisv: with regards to the `potentialOwner` bug we ran across earlier this year, as part of the PR, I have an "coverage" test which automagically tries every query method in order to validate that _every_ query method works and does not through an exception. (I <3 reflection! ;D). 

For an example (the `jbpm-human-task-audit` query builders, in this case), see
- The [TaskAuditQueryBuilderCoverageTest](https://github.com/mrietveld/jbpm/blob/_query-work/jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/TaskAuditQueryBuilderCoverageTest.java#L128-150)

"Coverage" tests exist for all query builders, and all make use of the following test/method:
- [QueryBuilderCoverageTestUtil](https://github.com/mrietveld/jbpm/blob/_query-work/jbpm-query-jpa/src/test/java/org/jbpm/query/QueryBuilderCoverageTestUtil.java#L89)

mrietveld:
FYI: I need to first finish the code for `kie-remote-services`, that's why I'm waiting to push this.

livthomas:
@mrietveld Why there is a new module just for task query tests? If it is not possible to have them in a single module within jbpm-human-task-*, what about putting them into the jbpm-test-coverage module?

mrietveld:
@livthomas see [here](https://github.com/droolsjbpm/jbpm/pull/284/files#diff-c71d3d46e12344f6a3623b7189ed45ebR106). 

I ran into a bug in the version of Hibernate we now use (4.2.9.Final) -- luckily, it's been fixed in 4.2.19.Final, so it looks like we will be moving to that version for the kie/drools/jbpm 6.3.0.Final release (this coming week).

mrietveld:
@livthomas When the version is upgraded, I'll move the tests back to the `jbpm-human-task-core` module. 

See the [KeepQueryTestsInATheJbpmHumanTaskQueryTestsModuleTest](https://github.com/mrietveld/jbpm/blob/95883c02fb743b5db42cbf6323676ececdbd3712/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/query/KeepQueryTestsInATheJbpmHumanTaskQueryTestsModuleTest.java).

Salaboy:
@mrietveld I'm worried about changing the jbpm-audit and jbpm-human-task-audit modules at this point. All the UI and services rely on these two modules for displaying data. Now the integration with the DataSets are also relying on the data generated by this modules. So I would like to avoid testing everything again due to this changes. I would recommend to leave the changes in those two specific modules for after the release. Improving the old REST endpoints is ok IMO, but we need to make sure that we don't break any of the new kie-server implementation

mrietveld:
@Salaboy Could you point to specific changes in the jbpm-audit and jbpm-human-task-audit modules? I can leave the jbpm-human-task-audit changes out, but the jbpm-audit changes are necessary.

mrietveld:
@Salaboy Please try to remember that I want _constructive_ criticism! Focus on solutions, not on problems! :)

Salaboy:
@mrietveld I'm worried about changes in behaviour and how the data is generated in the ProcessInstanceLog, VariableInstanceLog and AuditTaskImpl classes. If the behaviour is the same and the changes only impacts the way that the REST endpoints are going to consume that data, then the changes are fine IMO. I want to avoid problems in the release that's why I'm adding comments here.

mrietveld:
@Salaboy Ah, I get that: nothing is changing in how the log instances is generated. 

The only thing that's being is changed is how the log instances are being queried.

mrietveld:
@Salaboy also, the only change in behavior has to do with how the query builders work, with regards to the operator precedence of and/.intersection() and or/.union() operators -- but only for the query builder instances. Before, the operator precedence was basically random, now it's the same as in SQL.

Salaboy:
@mrietveld Thanks, I just wanted confirmation on that side. The builders looks cool by the way!

mswiderski:
nice piece of work Marco! looks good to me assuming all tests are green. 

I believe that with this commit we do change the API that might have been used in 6.2 (and possibly earlier) though I think it is rather limited scope so should not be a huge issue.

Delete builders are still based on string concatenation, correct? There is no criteria API for delete/update queries in the version of JPA?hibernate we use at the moment...or?

mrietveld:
Delete builders still (unfortunately) use string concatentation: the delete/update criteria API is only available in JPA 2.1 (hibernate 4.3.x). However, it will be very easy to move to that because they will of course use the same framework that the normal query builders use. 

Yes, there are some changes to kie-internal -- but none to kie-api. That's why I kept the interfaces in kie-internal: I expected there to be changes. :) 

Thanks for your comment!

mrietveld:
See the commits pushed to master.

PR #285, opened 2015-07-27
[JBPM-4708] Add a Parser service to Service Repository

jesuino:


mswiderski:
slightly enhanced and merged into master, thanks @jesuino

PR #286, opened 2015-08-05
BPMSPL-124 TaskSubjectAndDescriptionTest

ibek:


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/TaskSubjectAndDescriptionTest.java
```diff
@@ -0,0 +1,98 @@
...
+        RuntimeEngine runtimeEngine = getRuntimeEngine();
+        kieSession = runtimeEngine.getKieSession();
+        taskService = (SynchronizedTaskService) runtimeEngine.getTaskService();
+    }
+
+    @Test
+    public void testSubjectAndDescriptionProperties() {
+        ProcessInstance processInstance = kieSession.startProcess(HUMAN_TASK_ID);
+
+        List<TaskSummary> list = taskService.getTasksAssignedAsPotentialOwner("john", "en-UK");
```
triceo: There's a whole lot of repetition in this test... "john", "en-UK", "Bake a cake"etc. ... Can you make those into constants and reuse?

ibek: I could but this way it is clearer what we expect there when the test failed than to look for values of the constants. The rest of the tests do not use any constants too. If we ever decide to use some enumerations and constants, it would be better to refactor all the tests at once. WDYT?

triceo:
I agree that, by using plain strings, it is immediately clear what we're expecting. However, you're also losing some very important information - if you use a constant instead of two identical strings, you are in fact telling everyone that those two lines are using the *same thing*; not just two unrelated strings that happen to be the identical.

The fact that it had been done wrong in the past is no reason to not start doing it better now. :-) Eventually, the amount of good code will overshadow the amount of old code.

mswiderski:
are there plans to do any further changes to this PR or shall I go ahead and merge it?

ibek:
Maciej, I would say that it is all, you can merge the PR. When it is done, I can do cherry pick to 6.3.x where we need it too.

mswiderski:
merged into master, thanks @ibek

PR #287, opened 2015-08-05
DataServiceItemDefinitionHandler NPE fix

mrietveld:
@mswiderski Could you look at this?

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
@@ -147,8 +147,7 @@ public class BPMN2DataServiceImpl implements DefinitionService, DeploymentEventL
             BPMN2ProcessFactory.setBPMN2ProcessProvider(provider);
         }
         try {
-            ProcessDescRepoHelper threadLocalHelper = new ProcessDescRepoHelper();
-	        BPMN2DataServiceSemanticModule.setRepoHelper(threadLocalHelper);
+            ProcessDescRepoHelper threadLocalHelper = BPMN2DataServiceSemanticModule.getRepoHelper();
 	        BPMN2DataServiceExtensionSemanticModule.setRepoHelper(threadLocalHelper);
 	      
 	        KnowledgeBuilder kbuilder = null;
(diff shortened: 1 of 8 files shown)


mrietveld:
@sutaakar I made a byteman test! And I improved another test by using the `CyclicBarrier` class. If you have any questions, please let me know. :)

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -147,8 +147,7 @@ public ProcessDefinition buildProcessDefinition(String deploymentId,String bpmn2
             BPMN2ProcessFactory.setBPMN2ProcessProvider(provider);
         }
         try {
-            ProcessDescRepoHelper threadLocalHelper = new ProcessDescRepoHelper();
-	        BPMN2DataServiceSemanticModule.setRepoHelper(threadLocalHelper);
+            ProcessDescRepoHelper threadLocalHelper = BPMN2DataServiceSemanticModule.getRepoHelper();
```
mswiderski: so is this the main change to avoid the NPE? was there some sort of race condition that get in the way?

mrietveld: The main change is actually the one onder here (Overriding the `ThreadLocal<ProcessDescRepoHelper>.get()` method in `BPMN2DataServiceSemanticModule`). 

If you change the code back to the way it was and run the attached test, you'll see an NPE caused by the race condition. The race condition here is that the ` BPMN2ProcessFactory.setBPMN2ProcessProvider(provider)` method sets the provider for *all* threads, while the repo helper is only set for the local threads. 

Because the data service `BPMN2ProcessProvider` implementation references handlers that use the repo helper (which, in the case of other threads, is null), an NPE is thrown. I've seen this NPE regularly during startup of kie-wb.

mswiderski: alright, thanks for the explanation. looks good to me. You can go ahead with the push or I can merge that.

mrietveld:
Thanks @mswiderski.

Merged @ https://github.com/droolsjbpm/jbpm/commit/c7666467421b2c2d6021c4d840e83101039c0da5

PR #288, opened 2015-08-10
BPMSPL-193 - added tests and various corrections

sutaakar:
Added tests for pagination of returned requests and for clearing all requests.
Various fixes:
- In memory executor didn't delete processed requests when was called clearAllRequests()
- In memory executor missed pagination implementation for some methods
- Wrong process file name


mswiderski:
merged into master and 6.3.x, thanks @sutaakar

PR #289, opened 2015-08-11
BPMSPL-209 - Provide alternative to DB based job executor

mswiderski:
ready for review


jbpm-services/jbpm-executor/data/journal/server.lock
```diff
@@ -0,0 +1 @@
+P00�Ѻ�;��ڹL�:
```
mrietveld: Is this test data?

mrietveld: github support: adding a line comment is possible. 

However, the "pencil" icon in the upper right hand corner is showing a message that I'm not able to add a comment unless I'm signed in and have push access. 

My further apologies for the vagueness.. :/

jbpm-services/jbpm-executor/pom.xml
```diff
@@ -156,6 +163,40 @@
       <artifactId>jbpm-audit</artifactId>
       <optional>true</optional>
     </dependency>
+    
+    <!-- Test: jms -->
+    <dependency>
+      <groupId>org.hornetq</groupId>
+      <artifactId>hornetq-server</artifactId>
```
mrietveld: Should this be test scope?

jbpm-services/jbpm-executor/pom.xml
```diff
@@ -156,6 +163,40 @@
...
+          <groupId>io.netty</groupId>
+          <artifactId>netty</artifactId>
+        </exclusion>
+      </exclusions>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.hornetq</groupId>
+      <artifactId>hornetq-jms-client</artifactId>
+      <exclusions>
```
mrietveld: And this too? (test scope)

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
```diff
@@ -0,0 +1,255 @@
...
+    
+                    request.setStatus(STATUS.DONE);
+                     
+                    executorStoreService.updateRequest(request);
+                    processReoccurring = true;
+                    
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                } catch (Throwable e) {
+                	callbacks = classCacheManager.buildCommandCallback(ctx, cl);
```
mrietveld: This is just personal preference: 

Would it be possible to put this logic into it's own method? 
That makes it more readable and it also makes it easier to catch any eventual bugs from later modifications.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
```diff
@@ -0,0 +1,255 @@
...
+                        if (callbacks != null) {
+                            for (CommandCallback handler : callbacks) {                        
+                                handler.onCommandError(ctx, e);                        
+                            }
+                        }
+                        processReoccurring = true;
+                    }
+    
+                } finally {
+                	if (processReoccurring && cmd != null && cmd instanceof Reoccurring) {
```
mrietveld: Same as above: maybe in it's own method?

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorImpl.java
```diff
@@ -56,12 +65,20 @@
...
     private int retries = Integer.parseInt(System.getProperty("org.kie.executor.retry.count", "3"));
     private int interval = Integer.parseInt(System.getProperty("org.kie.executor.interval", "3"));
     private int initialDelay = Integer.parseInt(System.getProperty("org.kie.executor.initial.delay", "100"));
     private TimeUnit timeunit = TimeUnit.valueOf(System.getProperty("org.kie.executor.timeunit", "SECONDS"));
+    
+    
+    // jms related instances
+    private boolean useJMS = Boolean.parseBoolean(System.getProperty("org.kie.executor.jms", "true"));
+    private String connectionFactoryName = System.getProperty("org.kie.executor.jms.cf", "java:/JmsXA");
+    private String queueName = System.getProperty("org.kie.executor.jms.queue", "queue/KIE.EXECUTOR");
```
mrietveld: Documentation? :)

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
```diff
@@ -0,0 +1,255 @@
...
+                        }
+                    }
+    
+                    request.setStatus(STATUS.DONE);
+                     
+                    executorStoreService.updateRequest(request);
+                    processReoccurring = true;
+                    
+                } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
```
mrietveld: What's throwing the InterruptedException?

mrietveld:
+1

Thanks for sharing the PR!

jbpm-services/jbpm-executor/data/journal/server.lock
mswiderski: correct, it should not be there, will remove before the merge

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
mswiderski: it can be thrown by bitronix on shutdown of the transaction manager, not sure if that can happen on regular app servers but if the thread was not interrupted it affected all other test executions

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
mswiderski: you mean to put all code from catch and finally into own methods? If so that's actually good idea, will make that happen :)

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorImpl.java
mswiderski: yup, will describe this in docs when merged but will add just few lines of comments here any way

mswiderski:
thanks Marco, will leave the PR open for one or two days more and then merge it by the end of the week.

mswiderski:
merged into master and 6.3.x, kie-wb-distribution-wars PR will be merged once BPMSPL-210 is merged so it gets complete setup of JMS

PR #292, opened 2015-08-13
BPMSPL-210 - Provide an alternative approach of notifying the parent …

mswiderski:
…process when using waitForCompletion=true


jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/AbstractNodeHandler.java
```diff
@@ -544,13 +545,28 @@ protected void writeVariableName(EventNode eventNode, StringBuilder xmlDump) {
...
+            "workItem.setDeploymentId((String) kcontext.getKnowledgeRuntime().getEnvironment().get(\"deploymentId\"));" + EOL +
+            "workItem.setParameter(\"Signal\", org.jbpm.process.instance.impl.util.VariableUtil.resolveVariable(\""+ signalName + "\", kcontext.getNodeInstance()));" + EOL +
+            "workItem.setParameter(\"SignalProcessInstanceId\", kcontext.getVariable(\"SignalProcessInstanceId\"));" + EOL +
+            "workItem.setParameter(\"SignalWorkItemId\", kcontext.getVariable(\"SignalWorkItemId\"));" + EOL +
+            "workItem.setParameter(\"SignalDeploymentId\", kcontext.getVariable(\"SignalDeploymentId\"));" + EOL +
+            (variable == null ? "" : "workItem.setParameter(\"Data\", " + variable + ");" + EOL) +            
+            "((org.drools.core.process.instance.WorkItemManager) kcontext.getKnowledgeRuntime().getWorkItemManager()).internalExecuteWorkItem(workItem);";
+        } else {
+            signalExpression = signalExpression +  "org.jbpm.process.instance.impl.util.VariableUtil.resolveVariable(\""+ signalName + "\", kcontext.getNodeInstance()), " + (variable == null ? "null" : variable) + ");";
```
mrietveld: Is it possible to add this logic to a class, and then have the script expression instantiate that class and execute a method? (A very little bit like of like a `WorkItemHandler`)

(Personal preference: I'd rather not have this amount of "constant" text)

jbpm-flow/src/main/java/org/jbpm/process/core/async/AsyncStartProcessCommand.java
```diff
@@ -0,0 +1,103 @@
...
+import org.kie.internal.process.CorrelationAwareProcessRuntime;
+import org.kie.internal.process.CorrelationKey;
+import org.kie.internal.process.CorrelationKeyFactory;
+import org.kie.internal.runtime.manager.RuntimeManagerRegistry;
+import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
+
+
+public class AsyncStartProcessCommand implements Command {
+
+    private CorrelationKeyFactory correlationKeyFactory = KieInternalServices.Factory.get().newCorrelationKeyFactory();
```
mrietveld: `private static`, since the correlation key factory is thread-safe? (I looked at both implementations, they looked thread-safe to me..?)

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/AbstractNodeHandler.java
```diff
@@ -16,9 +16,6 @@
 
 package org.jbpm.bpmn2.xml;
 
-import static org.jbpm.bpmn2.xml.ProcessHandler.PROCESS_INSTANCE_SIGNAL_EVENT;
-import static org.jbpm.bpmn2.xml.ProcessHandler.RUNTIME_MANAGER_SIGNAL_EVENT;
-import static org.jbpm.bpmn2.xml.ProcessHandler.RUNTIME_SIGNAL_EVENT;
```
mrietveld: Should these constants be deleted from the `ProcessHandler` class?

mswiderski: no, it should still be there as it is used by other types of events in different format. I removed a shared once for signals to make use of different approach, but others still need it.

jbpm-flow/src/main/java/org/jbpm/process/core/async/AsyncStartProcessCommand.java
mswiderski: good point, will make it static

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/AbstractNodeHandler.java
mrietveld: Ah, thanks!

PR #293, opened 2015-08-14
JBPM-4717 - fixed error brought by cherrypick

sutaakar:


mswiderski:
@sutaakar please go ahead and push that to master and 6.3.x if needed (I am in the middle of other work so cannot easily merge it now)

sutaakar:
@mswiderski Done in 6.3.x. In master branch there is correct file name, just forgot to remove old file there - so they are duplicated now - will remove file with wrong name in master.

PR #294, opened 2015-08-21
moved jbpm-test tests into jbpm-test-coverage

mswiderski:


mswiderski:
@psiroky whenever you find some time - this should reduce the overall build time and I tried to stabilize some of the timer based tests to not rely on thread.sleep so heavily. 

On top of it (second commit) aims at fixing this errors recently reported with random failures of jbpm-test-coverage.

psiroky:
+10 from me. Should I merge it? (I was thinking that someone else with deeper jbpm knowledge would look at that, but I am fine with running the build locally to verify that everything is working and pushing it).

mswiderski:
if you could please go ahead with local build and see how it behaves - for me it builds fine - tried few times already but that's not a guarantee it will work everywhere. Next we can push to master and let it build few times on jenkins and then cherry-pick, wdyt?

psiroky:
OK, I will run the tests locally to see if everything works as expected.

Maybe we could ask someone from QE to briefly review these changes since they are related to tests?

mswiderski:
+1 to get additional review

psiroky:
Running the tests locally, I got one test failure:
`PessimisticLockTasksServiceTest.testPessimisticLockingOnTask:177

java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:743)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:555)
	at org.junit.Assert.assertEquals(Assert.java:542)
	at org.jbpm.test.functional.task.PessimisticLockTasksServiceTest.testPessimisticLockingOnTask(PessimisticLockTasksServiceTest.java:177)
`

Sorry, I don't have any more details (logs). Trying to run the build again.

psiroky:
I ran the the build again three times - it was successful in two cases and once it failed with one test error and one test failure. However, I don't think the failures are actually caused by the move of the tests. If I remember correctly I have seen the PessimisticLockTasksServiceTest failed before as well.

mswiderski:
Thanks Petr. Any idea about how to improve that test to be more reliable?
Feel freeto ignore it and merge if you like

Dnia 25 sie 2015 o godz. 09:02 Petr Široký <notifications@github.com> napisał(a):

psiroky:
@ibek @livthomas could one of you please review this PR?

psiroky:
I did following changes to the `PessimisticLockTasksServiceTest` https://gist.github.com/psiroky/fd7049bdef52ab92aa36 and ran the tests 22 times over the night. The build was successful 17 times and failed 5 times, but because of different test failures. Once the `GlobalQuartzDBTimerServiceTest>GlobalTimerServiceBaseTest.testTimerStart` failed and in 4 case `ConcurrentGlobalTimerServiceTest.testSessionPerProcessInstance` failed.

The `PessimisticLockTasksServiceTest` seems to be stable now, but looks like there are other tests worth looking into.

psiroky:
@domhanak would you be able to look at this? (Your name came up in discussion with @ibek ).

livthomas:
@mswiderski 
+1 for changing the parent of those test classes to JbpmTestCase since this class is used to set up persistence when run jBPM tests on commercial database systems.
+1 for fixing the Bitronix problem when using the executor. I have noticed the problem the problem long ago but have never been able to find the cause.

Anyway, I tried to compile the whole jBPM in this branch but got some compilation errors. Can you fix them? However, I have been able to compile the master, switch to this branch and then just run tests. I have got two failures:

PessimisticLockTasksServiceTest.testPessimisticLockingOnTask:177->Assert.assertEquals:542->Assert.assertEquals:555->Assert.assertEquals:118->Assert.failNotEquals:743->Assert.fail:88 expected:<1> but was:<0>
  MultipleTimerServicesTest.testGlobalTimerServiceOnIndependentSingletonManager:111->testGlobalTimerServiceOnIndependentManager:201->Assert.assertNull:656->Assert.assertNull:646->Assert.failNotNull:664->Assert.fail:88 expected null, but was:<WorkflowProcessInstance2 [processId=IntermediateCatchEvent,state=1]> 

It looks like @psiroky fixed the first one but I am not sure about the second one. Maybe it is again one of the random fails and we just need to stabilize the tests in jbpm-test-coverage module more.

psiroky:
Yeah, once there is a `timer` in the name of the test, it is likely a random failure. (However, this is exactly the reason why should not have random failures, because even if the test founds a real bug in the future it will be probably treated as random failure).

mrietveld:
+1 for merging this first and fixing the smaller problems after the merge. 

Otherwise, the PR gets harder to merge.. :/

psiroky:
+1 for merging this (assuming the tests will be fixed in timely manner :))

mswiderski:
merged into master and 6.3.x

PR #295, opened 2015-08-28
JBPM-4749 Persistence tests don't support Postgres Plus

baldimir:
Added support for com.edb.Driver class to data source properties handling.

Please if it's possible, merge it also to 6.3.x and 6.2.x branch. Without this we cannot execute community tests on Postgres Plus.


mrietveld:
Cherry-picked to 6.3.x and 6.2.x

baldimir:
Thanks a lot :)

PR #298, opened 2015-09-02
JBPM-4714 Case management test coverage

ibek:


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/casemgmt/WorkCaseTest.java
```diff
@@ -0,0 +1,198 @@
...
+
+    @Test(timeout = 30000)
+    public void testProlongDueDate() {
+
+        String responsible = "john";
+        long caseId = wcs.createWork("Integrate module X into project Y", "Module X working in the project Y by ...", responsible);
+        wcs.prolongTimeoutTo(caseId, 15);
+
+        try {
+            Thread.sleep(11000);
```
mrietveld: Is it possible to make the sleeps shorter? (max 2 seconds?)

What is the thread waiting for?

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/casemgmt/WorkCaseTest.java
```diff
@@ -0,0 +1,198 @@
...
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+
+        ProcessInstanceLog pil = auditService.findProcessInstance(caseId);
+        Assertions.assertThat(pil.getStatus()).isEqualTo(ProcessInstance.STATE_ACTIVE);
+
+        try {
+            Thread.sleep(5000);
```
mrietveld: Could you explain what's happening here? The way the test is currently written, it will cause false negatives (either the process will complete too quickly (above assert fails) or the process will take too long (below assert fails)).

jbpm-test-coverage/src/main/java/org/jbpm/test/jobexec/CheckCallCommand.java
```diff
@@ -0,0 +1,40 @@
...
+import org.kie.api.executor.ExecutionResults;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class CheckCallCommand implements Command {
+    
+    private static final Logger logger = LoggerFactory.getLogger(CheckCallCommand.class);
+
+    private static Boolean commandExecuted = false;
+    private static Object LOCK = new Object();
```
mrietveld: Instead of using both the Boolean and Object/LOCK, it might be more efficient to use a CyclicBarrier or CountDownLatch (a CyclicBarrier is essentially a reusable CountDownLatch).

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/casemgmt/WorkCaseTest.java
ibek: Ok I will shorter the timeout. Basically, there is an intermediate timer for timing out. Currently, by default it is set to timeout in 10s. I update the timer delay to 15s and here the thread waits 11s (more than 10s) to check below that the process instance is still active. Then, it waits another 5s -> it means in total it waits 5+11s = 16s which is more than 15s in the timer and so process should complete.

jbpm-test-coverage/src/main/java/org/jbpm/test/jobexec/CheckCallCommand.java
ibek: I believe that I always need some boolean var to determine if the command was really called (in assertion) because the waiting for the lock may timeout and without the boolean var I do not know if the lock was released by the command or timeout. I am not sure how CyclicBarrier would improve it.

mrietveld: A CyclicBarrier or CountDownLatch would (I think?) eliminate the need to have both a boolean and a LOCK, and also eliminate the need for a sleep or wait on the LOCK field. 

If you want, I can go ahead and merge it and then change the code to show you how I would do it?

ibek: ok, go ahead ... I am curious about your solution :)

jbpm-test-coverage/src/main/java/org/jbpm/test/casemgmt/UpdateTimerCommand.java
```diff
@@ -0,0 +1,86 @@
+package org.jbpm.test.casemgmt;
```
psiroky: Please add licence header to this file (and other as well). Template can be found at https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/blob/master/ide-configuration/LICENSE-ASL-2.0-HEADER.txt

ibek: done

mrietveld:
@ibek I'm getting one failure on the build: 


    Results :
    
    Failed tests: 
      ParallelAsyncJobsTest.testRunBasicAsync:80 [More than 2 async jobs should have been executed] 
    Expecting:
     <3>
    to be less than or equal to:
     <2> 
    
    Tests run: 235, Failures: 1, Errors: 0, Skipped: 21

Could you look at this test and fix it? Thanks!

ibek:
@mrietveld  that's strange, it works fine for me on master:

Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 19.487 sec - in org.jbpm.test.regression.ParallelAsyncJobsTest

@livthomas could you look at the test since you've developed it?

livthomas:
I have not developed that test only moved it here from our regression test suite. I know there is a problem and it fails every know and then but I have not been able to fix it so far. Right now, it passes on my side. Anyway, from what I have seen in e-mails and other PRs, the stabilization of tests is in progress and this PR does not modify any files that can influence ParallelAsyncJobsTest so I think it can be merged.

PR #303, opened 2015-09-14
JBPM-4714: select random user from the case roles

ibek:
Changes required after discussion with @krisv in the BZ https://bugzilla.redhat.com/show_bug.cgi?id=1257975


mswiderski:
merged into master and 6.3.x

PR #305, opened 2015-09-17
- adding making sure that an empty list is returned if no groups

Salaboy:
This Pull Request is fixing the NPE reported in the BZ, it should be merged also in 6.3.x


jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/LocalWithCustomCallbackTaskAuditTest.java
```diff
@@ -0,0 +1,119 @@
+/*
+ * Copyright 2012 JBoss by Red Hat.
```
psiroky: 2015

psiroky:
If it is directly related to some BZ (it seems so from the description), would it be possible to add the BZ number to the commit message?

mswiderski:
I'll updated these two while merging

PR #307, opened 2015-09-25
JBPM 4714: fixed UpdateTimerCommand to inherit from ProcessInstanceId…

ibek:
…Command

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/instance/command/UpdateTimerCommand.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/instance/command/UpdateTimerCommand.java
@@ -16,6 +16,12 @@
 
 package org.jbpm.process.instance.command;
 
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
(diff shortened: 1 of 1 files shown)


mswiderski:
@ibek any reason to include jaxb annotations? Are you expecting this to be available over remote interfaces? Personally I consider it to be sort of internal command that should be used only in certain situations. 
Then implementing ProcessInstanceIdCommand is not so relevant... Or am I missing something?

ibek:
@mswiderski  Marco(@mrietveld ) sent me email that the command should inherit from ProcessInstanceIdCommand because of the failing test:

https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/master/job/jbpm/org.jbpm$jbpm-kie-services/310/testReport/junit/org.jbpm.kie.services.impl/CommonUtilsTest/testProcessInstanceIdCommands/   (every command that uses processInstanceId as a parameter has to inherit from the interface)

Well, I do not know what plans we have for adding case management operations to remote API but this command could be one of them, so I added the jaxb annotations for the future when I was changing the command.

mrietveld:
Thank you for your quick response, Ivo!

PR #309, opened 2015-09-25
BZ-1266092 - NPE caused by race condition in data services

mrietveld:
@mswiderski *when you get back from PTO* ;), please take a look at this and let me know what you think. 

See [BZ-1266092](https://bugzilla.redhat.com/show_bug.cgi?id=1266092) for more info.


mswiderski:
Marco, now I think we are entering the point of complexity with this thead local handling that we should invest day or two to move this logic into regular parsing phase and keep it as part of kbase/process definitions meta data. If you like I can take a look at this next week

mrietveld:
Hahah.. agreed! :) 

Because this is a fix for 6.3, I approached this as a (band-aid solution)[https://en.wikipedia.org/wiki/Band-Aid], but I agree that this should be in the regular parsing phase (e.g. in jbpm-bpmn2). 

 If you have time this week, that would be great. Otherwise, we might have time to discuss it next week.

mswiderski:
@mrietveld go ahead and merge this to solve the problem short term and we'll discuss it next week to figure out more appropriate solution to avoid this (and similar issues) in the future

PR #310, opened 2015-09-25
BZ-1265725 - The /rest/query/task operation is returning lists with duplicate TaskSummary's

mrietveld:
@baldimir Could you check if this is correct? 

Unfortunately, I can't think of a good unit test for this (that doesn't not involve DB2 or Oracle). If you have any ideas, please let me know. 

Thanks!


mrietveld:
Sometimes the code orders by columns not returned in the select, which invalidates using `distinct`. 

Back to the drawing board for a solution to this one..

baldimir:
The main problem here is that order by is applied at last in SQL. So it is applied also after distinct, which is an aggregate function. (Please correct me if I'm wrong with this)

I think we have three options here: 
- Not order by column that is not in select clause. I think maybe we should go with this option, if it is possible, because I don't know if all production DB systems, that we use, support using column that is not part of select clause in order by clause. 
- Order by with an aggregate function, if the column is not in select clause.  E.g. "order by min(Task.id)". This should help, because it states, which value of unaggregated column should it use for ordering. 
- Order by in a subquery and then use distinct. This will do the ordering before distinct. This is the second most preferred option for me.  

What do you think?

mrietveld:
@baldimir Fixed by using distinct, after all, see https://github.com/droolsjbpm/jbpm/commit/0987f9564285cc279b0801eb7a43686e7b0e1714

PR #311, opened 2015-10-01
HumanTaskServicesBaseTest: use null jurnal for bitronix to avoid inte…

psiroky:
…rruptep thread destroying transaction manager when using default disc journal

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/HumanTaskServicesBaseTest.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/HumanTaskServicesBaseTest.java
@@ -37,6 +37,7 @@ import javax.xml.bind.JAXBContext;
 import javax.xml.bind.Marshaller;
 import javax.xml.bind.Unmarshaller;
 
+import bitronix.tm.TransactionManagerServices;
 import org.jbpm.persistence.util.PersistenceUtil;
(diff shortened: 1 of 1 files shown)


psiroky:
Please cherry-pick this into 6.3.x as well, if accepted.

mrietveld:
Looks fine -- although I don't know anything about the problem. Could you explain what the problem is? 

If you could also add it to the persistence test classes in drools-persistence-jpa, that would also be nice -- if it's applicable there. 

Thanks.

psiroky:
TBH I don't really know that much about this. I just copied the changes @mswiderski did here https://github.com/droolsjbpm/jbpm/commit/ae6c2de28c231893fe5c270d72b584d624338fad (after checking with him). So this or may not be applicable to other tests (don't know).

mswiderski:
in general for test purpose we can switch all tests to use NULL journal instead of disc one which is fragile to thread interruption. @psiroky would you mind extend this PR to include others as well?

psiroky:
@mswiderski what others do you mean? All the top-level/abstract test classes?

mswiderski:
yes, all base classes for modules starting at jbpm-persistence-jpa and up in maven module hierarchy

psiroky:
Ouch. That will likely take more than few minutes (at least to me). I can take a look at that, but now sure when.

mswiderski:
no worries, whoever gets into this earlier then will do it

mswiderski:
closing this one as #314 has been merged, thanks @psiroky

PR #313, opened 2015-10-02
JBPM-4792 - Refactor DefinitionService (BPMN2DataServiceImpl) to avoid complexity with thread local

mswiderski:
@mrietveld take a look and let me know what you think?

P.S.
I removed byteman as it did not see a reason to have it as it was testing the thread locals. If I misunderstood the use of that test we can bring it back - though needs to be refactored as there is no extended BPMN2 xml provider. Second commit is just to make it compile after addition in drools to filter dependencies of kjar


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -131,9 +83,29 @@ private void validateNonEmptyDeploymentIdAndProcessIdAndTaskName(String deployme
             throw new IllegalStateException( msg.toString() );
         }
     }
+    
+    public void addProcessDefinition(String deploymentId, String processId, Object processDescriptor, KieContainer kieContainer) {
+        Map<String, ProcessDescriptor> definitions = null;
+        synchronized (definitionCache) {
+            definitions = definitionCache.get(deploymentId);
+            if (definitions == null) {
```
mrietveld: Could you use putIfAbsent() here instead of a null check? The null check will lead to a race condition.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
```diff
@@ -293,23 +319,29 @@ protected Object getInstanceFromModel(ObjectModel model, KieContainer kieContain
...
                     }
-                    process.setEncodedProcessSource(Base64.encodeBase64String(processString.getBytes()));
-                    process.setDeploymentId(unit.getIdentifier());
-                    deployedUnit.addAssetLocation(process.getId(), process);
                 } catch (UnsupportedEncodingException e) {
-                	throw new IllegalArgumentException("Unsupported encoding while processing process " + fileName);
+                    throw new IllegalArgumentException("Unsupported encoding while processing process " + fileName);
                 }
-            } else if (fileName.matches(".+ftl$") || fileName.matches(".+form$")) {
+            }if (fileName.matches(".+ftl$") || fileName.matches(".+form$")) {
```
mrietveld: I think this `else if` should stay in here for readability (and to make absolutely sure that the file is processed correctly).

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -131,9 +83,29 @@ private void validateNonEmptyDeploymentIdAndProcessIdAndTaskName(String deployme
...
+                definitions = new ConcurrentHashMap<String, ProcessDescriptor>();
+                definitionCache.put(deploymentId, definitions);
+            }
+            ProcessDescriptor descriptor = (ProcessDescriptor) processDescriptor;
+            ProcessAssetDesc definition = descriptor.getProcess();
+
+            if( kieContainer != null && descriptor.hasUnresolvedReusableSubProcessNames() ) {
+                descriptor.resolveReusableSubProcessNames(kieContainer.getKieBase().getProcesses());
+            }
```
mrietveld: I don't think this will work when you're dealing with imported processes..

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
```diff
@@ -141,172 +113,51 @@ public ProcessDefinition buildProcessDefinition(String deploymentId,String bpmn2
...
+           helper.resolveReusableSubProcessNames(kieContainer.getKieBase().getProcesses());
+        }
+        definition.setReusableSubProcesses(helper.getReusableSubProcesses());
+
+        // cache the data if requested
+        if (cache) {
+        	Map<String, ProcessDescriptor> definitions = null;
+        	synchronized (definitionCache) {
+        		definitions = definitionCache.get(deploymentId);
+        		if (definitions == null) {
```
mrietveld: putIAbsent(), please.. :)

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/BPMN2DataServiceImplMultiThreadBytemanTest.java
```diff
@@ -1,231 +0,0 @@
...
-        Set<String> dataServiceModuleThreads = testItemDefinitionHandler.getUsedByThreads();
-        assertEquals( "Only 1 thread should have accessed the data service handlers!",
-                      1, dataServiceModuleThreads.size() );
-        assertEquals( "Only the data service thread should have accessed the data service handlers!",
-                      dataServiceThreadName, dataServiceModuleThreads.iterator().next() );
-
-        // This will throw an exception if the process instance has not been created well
-        Map<String, Object> params = new HashMap<String, Object>(1);
-        params.put("name", "Poppy");
-        kbuilderHolder[0].newKnowledgeBase().newKieSession().startProcess("hello", params);
```
mrietveld: I think we need tests that test not only to make sure that any process built by the data services can also be executed.

mrietveld:
Looks good, Maciej -- I have some questions that I'll ask you tomorrow or next week.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
mswiderski: good point, I overlooked it

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/BPMN2DataServiceImpl.java
mswiderski: I don't think I can use the putIfAbsent here as I need to check and only once initializes the map. So I need to check and get it anyway on subsequent calls. And the race condition is unlikely as that would mean same deployment has to be added at exact same time.

mrietveld: putIfAbsent is *meant* for this situation. :) 

1. Only add the map if the map is not there! 
2. Otherwise, use the existing map (that another thread got to first)

Lastly, this is being used when we add a process definition -- given that we want to support scalable and multi-tenant solutions, multiple concurrent deployment requests (I add a process and deploy, you add a different new process and also want to deploy) going to be more and more likely, IMHO.

PR #314, opened 2015-10-14
cleanup of (timer, executor) tests to avoid Thread.sleep usage

mswiderski:
initial work that aims at stabilizing tests by avoiding Thread.sleep in tests as much as possible. It accommodates change o bitronix journal to avoid failed tests due to interrupted exceptions - so once reviewed it will replace this pr https://github.com/droolsjbpm/jbpm/pull/311

comments and ideas are more than welcome. 

P.S.
this is not final as jbpm-test-coverage might use some of the work there too. Though some help with that module would be needed as there are lots of tests.


mrietveld:
Great work!

mrietveld:
ok to test

psiroky:
@mrietveld sorry Marco, the CI setup is not yet done for this repo. But I am working on this right now, so should be available soon (today).

mrietveld:
Thanks, @psiroky!

mswiderski:
jenkins, retest this

mswiderski:
there are few test failures from jbpm-test-coverage but that is expected as most of the test coverage module was not refactored yet. Anyway, I opt to merge this to avoid making this PR stale. Votes?

mrietveld:
+1 Please merge.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TaskHandler.java
```diff
@@ -270,7 +270,7 @@ public Object end(final String uri, final String localName,
 				String uniqueId = (String) node.getMetaData().get("UniqueId");
 				forEachNode.setMetaData("UniqueId", uniqueId);
 				node.setMetaData("UniqueId", uniqueId + ":" + uniqueIdGen++);
-				node.setMetaData("hidden", true);
+//				node.setMetaData("hidden", true);
```
mrietveld: This is a typo, I think.

mrietveld:
There are a lot of changes to src/main here -- was that on purpose?

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TaskHandler.java
mswiderski: not really, it was incorrect behavior as it was skipping tasks within for each - multi instance on task (not subprocess) and by that process event listeners where not invoked.

mswiderski:
@mrietveld if you refer to changes within jbpm-executor then yes, they are done on purpose to provide (atm internal) event listeners handling in jbpm executor so we could laverage same approach for improving tests, though it might become more of user exposed feature later one. But for now it's only within jbpm-executor scope so it's not kie-api nor kie-internal as it needs more thorough thinking if that's all that is needed. Main requirements for this improvement was driven by tests needs although they to provide general observer mechanism for async jobs.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TaskHandler.java
mrietveld: Okay, would you mind removing the line then, instead of commenting it out? I think I would get confused by it I saw the commented-out line later.

mrietveld:
@mswiderski ah, that makes sense: thank you for explaining it. 

Great work here: please merge it ASAP! :)

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/TaskHandler.java
mswiderski: will do, thanks for pointing that out

mswiderski:
thanks for review Marco, will work on merge today. Shall that go into master only or 6.3.x as well?

mrietveld:
Both 6.3.x and master.

PR #318, opened 2015-10-26
JBPM-4768 - SQL upgrade scripts - changing SessionInfo primary key column data type fails

spolti:


baldimir:
One thing we should be aware of is that these scripts introduce new delimiter "@", so we should modify our SQL scripts tests so they accompany this delimiter. The influenced class is here [1]. Maybe I can create separate PR for that. 

[1] https://github.com/droolsjbpm/jbpm/blob/f2ab6908ae8c2a2cd2e23fc3c6cbf6a01badb34a/jbpm-installer/src/test/java/org/jbpm/persistence/scripts/util/SQLScriptUtil.java

mswiderski:
what's the reason to change the semi colon to at sign? Is this to allow the custom sql procedures to be defined?

spolti:
Hi Marco, I tried for a while create a procedure using ";" as delimiter, but i some errors, when changed to @ worked. As the script worked with @ i didn't test with ;

krisv:
Merged in master (with some additional changes):
https://github.com/droolsjbpm/jbpm/commit/b1b5c3480

PR #320, opened 2015-10-29
BZ-1276039 upgrade script (bpms-6.0-to-6.1.sql): null value in column…

lazarotti:
… id violates not-null constraint


lazarotti:
The BZ was opened about Postgres, but certainly Oracle should has the same behavior. I do not have an Oracle instance set but the fix in this PR seems enough. For Postgres I've tested and it is now working.

psiroky:
ok to test

mswiderski:
@baldimir do we have tests for these upgrade scripts for postgres and oracle? If so I was wondering how this could work before.

In general this change looks good to me so +1

baldimir:
@mswiderski Yes, I ran the tests also on Oracle and Postgres. You're right that this shouldn't work. I think that the tests passed because the majority of tables were empty when running the scripts, so the insert statement inserted nothing so no PK constraint was violated. We should add some more basic data to the tests. Especially Task data, because I see that in many cases values from that table are selected and inserted into other tables.

mrietveld:
@baldimir +1 on filling the tables beforehand, good catch.

krisv:
Merged (with some other improvements)

PR #322, opened 2015-10-29
BZ1276350 - added ordering for user task comments and attachments

sutaakar:
Added sorting also to attachments to behave consistently for both lists.

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/TaskCommentTest.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/TaskCommentTest.java
@@ -20,7 +20,6 @@ import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.assertNotEquals;
-import static org.junit.Assert.assertThat;
 
 import bitronix.tm.resource.jdbc.PoolingDataSource;
 import java.io.StringReader;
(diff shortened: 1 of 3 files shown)


mswiderski:
@sutaakar do you mind adding same order by to the named query https://github.com/droolsjbpm/jbpm/blob/master/jbpm-human-task/jbpm-human-task-jpa/src/main/resources/META-INF/Taskorm.xml#L1128

and would be good to add a test case even though if with default (H2) it will be always successful it will be used for other dbs tests as well.

sutaakar:
Added ordering to named query.

Test cases are already implemented in jbpm-human-task-core: see TaskCommentTest(updated to check order) and TaskAttachmentBaseTest.

mrietveld:
Hmm.. The only thing I wonder about is whether or not this will affect the code that sets the description field? I don't think so, but please double-check that.

mrietveld:
This is the logic I was thinking of: 
https://github.com/droolsjbpm/jbpm/blob/master/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java#L341

sutaakar:
That logic doesn't interact with comments/attachments, it shouldn't be affected.

mswiderski:
@sutaakar thanks for extending this PR, looks good from my point of view so +1

@mrietveld agree with @sutaakar I don't think descriptions will be affected by this change at all. If you're ok with that I'll go ahead and merge it in.

mrietveld:
@mswiderski +1 on merging. 

Thanks for checking out that logic: I didn't think so either, but wasn't sure if I was missing somethign..

PR #324, opened 2015-11-05
move bitronix setting to top level test to be applied to all test cases

mswiderski:
@psiroky here is the pr we talked about


psiroky:
+1 to merge once the build is done without any related issue/test fail. I know there still may be some false negatives.

psiroky:
The test failures are not related (most likely intermittent fails). We should fix them anyway (of course not as part of this PR) otherwise the PR builds loose their value as we will never have "green" build.

psiroky:
We should cherry-pick this into 6.3.x as well.

mswiderski:
done, both master and 6.3.x

PR #325, opened 2015-11-05
JBPM-4812 - Allow to query by task variables

mrietveld:
Depends on: 
- https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/105

Needed by: 
- https://github.com/droolsjbpm/droolsjbpm-integration/pull/221


mrietveld:
@mswiderski Please check.

mrietveld:
ok to test

mrietveld:
The failure is unrelated to the PR: 

    Caused by: org.h2.jdbc.JdbcSQLException: Deadlock detected. The current transaction was rolled back. Details: "
    Session #45 (user: SA) is waiting to lock PUBLIC.NODEINSTANCELOG while locking PUBLIC.PROCESSINSTANCEINFO (exclusive), PUBLIC.PROCESSINSTANCELOG (exclusive).
    Session #44 (user: SA) is waiting to lock PUBLIC.PROCESSINSTANCELOG while locking PUBLIC.NODEINSTANCELOG (exclusive), PUBLIC.WORKITEMINFO (exclusive), PUBLIC.VARIABLEINSTANCELOG (exclusive)."; 
    SQL statement: update public.ProcessInstanceLog set correlationKey=?, duration=?, end_date=?, externalId=?, user_identity=?, outcome=?, parentProcessInstanceId=?, processId=?, processInstanceDescription=?, processInstanceId=?, processName=?, processVersion=?, start_date=?, status=? where id=? [40001-168]

On the other hand, that's a fricking scary exception: a deadlock. I'm really hoping it's because [MVCC](http://www.h2database.com/html/advanced.html#mvcc) is not enabled on H2 and H2 is doing a table-level lock, and not because this is a symptom of a real bug..

mswiderski:
I applied the changes to fix ejb remote but then found other issues that actually need to be fixed before the merge as they currently breaks separation:
- TaskVariableImpl class was moved from jbpm-human-task-audit to jbpm-human-task-jpa and that introduced osgi split package as same package is no in both modules
- move of TaskVariableImpl to jpa modules from audit breaks the modules separation as it now opens the door to start joining runtime tables with audit tables which will then make more difficult in version 7 to go towards CQRS approach for tasks

with that in mind I'd like to suggest following:
- move back TaskVariableImpl to jbpm-human-task-audit
- either revert to basic TaskAuditService operation to query for task variables or use TaskAudit??QueryBuilder for these operations

mrietveld:
Hmm.. I was afraid of this. 

There's a solution that I have in mind that allows us to move `TaskVariableImpl` back to ht-audit. I'll modify the PR to do this.

mrietveld:
@mswiderski Updated the PR: 

1. The TaskVariableImpl class is back in the jbpm-human-task-audit module
2. The "normal" methods have been restored. The "query" method throws an unsupported exception in the EJB Impl.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -799,15 +801,98 @@ public UserTaskInstanceDesc getTaskById(Long taskId) {
...
+        if( statuses != null && statuses.length > 0 ) {
+            queryBuilder.status(statuses);
+        }
+
+        return queryBuilder.build().getResultList();
+    }
+
+    @Override
+    public List<TaskSummary> getTasksByVariable( String userId, String variableName, List<Status> statuses,
+            int offset, int total ) {
```
mswiderski: can we use QueryContext instead of int for offset and total? I agree that QueryFilter should not be used but QueryContext is more than paging, it has the sorting in it as well

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -799,15 +801,98 @@ public UserTaskInstanceDesc getTaskById(Long taskId) {
...
+            queryBuilder.status(statuses.toArray(new Status[statuses.size()]));
+        }
+        queryBuilder.offset(offset).maxResults(total);
+
+        return queryBuilder.build().getResultList();
+    }
+
+    @Override
+    public List<TaskSummary> getTasksByVariableAndValue( String userId, String variableName, String variableValue,
+            Status... statuses ) {
```
mswiderski: why no paging support here?

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-impl/src/main/java/org/jbpm/services/ejb/impl/RuntimeDataServiceEJBImpl.java
```diff
@@ -107,4 +111,10 @@ public void onDeactivate(DeploymentEvent event) {
 		super.onDeactivate(event);
 	}
 
+    @Override
+    public TaskSummaryQueryBuilder taskSummaryQuery( String userId ) {
```
mswiderski: it's fine to have it here, it should not be exposed on Remote interface of this service, local ejbs should still allow to use query builders. The only thing I am wondering if they should be accessible via service interface or standalone factory or so....

mswiderski:
would it be possible to use the service methods in the tests as well and not only query builder: https://github.com/mrietveld/jbpm/blob/pr-JBPM-4812-marco/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java#L1318

mswiderski:
another thing is that I believe there is a wrong condition for handling paging in query module - you will observe it when using the build with paging or service methods (https://github.com/mrietveld/jbpm/blob/pr-JBPM-4812-marco/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java#L1318) as it will not return results if there is only one matching the criteria.
https://github.com/mrietveld/jbpm/blob/pr-JBPM-4812-marco/jbpm-query-jpa/src/main/java/org/jbpm/query/jpa/builder/impl/AbstractQueryBuilderImpl.java#L131-L145
there should be 0 instead of 1 as the condition

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
mrietveld: Okay, will switch to QueryContext.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
mrietveld: See the next method.

mrietveld:
Updated PR: 

- used QueryContext in methods
- updated tests to use both query builder and normal methods
- changed query builder offset/maxresults (paging) condition

mrietveld:
@mswiderski the amount of changes to this PR is starting to get out of hand. If you're unsatisfied with the PR, please feel free to close it and simply merge your original commits. 

There has to be a compromise between merging code (progress) and quality.

mswiderski:
thanks @mrietveld  I building everything locally and then merge it. Noticed some issues in droolsjbpm integration but will try to fix it to avoid another iteration

mswiderski:
closing as new PR including all changes will be submitted

PR #326, opened 2015-11-06
JBPM-4792 - Add a test of Object variable

MarianMacik:
Hi, here is the promised test with one added process which contains Object variable.


mswiderski:
merged into master, thanks @MarianMacik

PR #327, opened 2015-11-09
BZ-1213209 - Add reproducer

livthomas:
I'm adding reproducer for the following BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1213209

Note that the BZ is targeting product version 6.2 so this commit should be also cherry-picked to 6.3.x branch.


mswiderski:
merged into both master and 6.3.x, thanks @livthomas

PR #328, opened 2015-11-09
BZ-1234592 - Add reproducer

livthomas:
Add simple reproducer for the following BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1234592

It needs to use special hibernate dialect DisabledFollowOnLockOracle10gDialect so I have also modified JbpmJUnitBaseTestCase to be able to set additional persistence unit properties.

Since the BZ mentioned above targets product version 6.2, this commit should be also cherry-picked to 6.3.x branch.

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/regression/async/AsyncWIHOnOracleTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/regression/async/AsyncWIHOnOracleTest.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2015 JBoss Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.test.regression.async;
+
+import org.assertj.core.api.Assertions;
+import org.jbpm.executor.impl.wih.AsyncWorkItemHandler;
+import org.jbpm.persistence.jpa.hibernate.DisabledFollowOnLockOracle10gDialect;
+import org.jbpm.persistence.util.PersistenceUtil;
+import org.jbpm.test.JbpmAsyncJobTestCase;
+import org.junit.After;
+import org.junit.Test;
+import org.kie.api.executor.Command;
+import org.kie.api.executor.CommandContext;
+import org.kie.api.executor.ExecutionResults;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.process.WorkItemManager;
+import qa.tools.ikeeper.annotation.BZ;
+
+public class AsyncWIHOnOracleTest extends JbpmAsyncJobTestCase {
+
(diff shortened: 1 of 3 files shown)


jbpm-test/src/main/java/org/jbpm/test/JbpmJUnitBaseTestCase.java
```diff
@@ -933,6 +936,10 @@ public void addEnvironmentEntry(String name, Object value) {
     	customEnvironmentEntries.put(name, value);
     }
 
+    public void addPersistenceProperty(String name, Object value) {
```
triceo: Better name for this would be ```setPersistenceProperty``` - this way, it is made clear that using the same ```name``` twice will override the previous value.

jbpm-test-coverage/src/test/java/org/jbpm/test/regression/async/AsyncWIHOnOracleTest.java
```diff
@@ -0,0 +1,98 @@
...
+    public void testAsyncWIHExecutedMoreThanOnceOnOracle() throws InterruptedException {
+        KieSession ksession = createKSession(PROCESS);
+        WorkItemManager wim = ksession.getWorkItemManager();
+        wim.registerWorkItemHandler("async", new AsyncWorkItemHandler(getExecutorService(), CounterCommand.class.getName()));
+
+        ksession.startProcess(PROCESS_ID);
+
+        Thread.sleep(10000);
+
+        Assertions.assertThat(CounterCommand.getCounter()).as("The job has not been executed").isNotEqualTo(0);
```
triceo: This would seem redundant, since the very next line is asserting that the counter equals 1.

jbpm-test-coverage/src/test/java/org/jbpm/test/regression/async/AsyncWIHOnOracleTest.java
```diff
@@ -0,0 +1,98 @@
...
+    @Test
+    @BZ("1234592")
+    public void testAsyncWIHExecutedMoreThanOnceOnOracle() throws InterruptedException {
+        KieSession ksession = createKSession(PROCESS);
+        WorkItemManager wim = ksession.getWorkItemManager();
+        wim.registerWorkItemHandler("async", new AsyncWorkItemHandler(getExecutorService(), CounterCommand.class.getName()));
+
+        ksession.startProcess(PROCESS_ID);
+
+        Thread.sleep(10000);
```
triceo: Can we do this any other way than by using a timer? We generally try to avoid those now.

mswiderski:
@livthomas are you going to apply changes suggested by @triceo ?

livthomas:
@mswiderski yes, I will look at that

jbpm-test-coverage/src/test/java/org/jbpm/test/regression/async/AsyncWIHOnOracleTest.java
livthomas: This is done on purpose. You can see there are different error messages in those two assertions. The first one should notify us that nothing has been been executed (and so there is some problem with job executor in general) and the second one that more jobs has been executed (and so there is a regression).

jbpm-test/src/main/java/org/jbpm/test/JbpmJUnitBaseTestCase.java
livthomas: I was just following the name of the method above which also puts entries into Map and starts with  `add`. But I agree that `set` makes more sense in this case.

livthomas:
I have changed reproducer not to use Thread.sleep() and I have also renamed JbpmJUnitBaseTestCase method for setting persistence properties.

mswiderski:
is this now ready to be merged?

livthomas:
@mswiderski In my opinion you can merge it now. I have applied two suggestions from @triceo which I found relevant.

PR #329, opened 2015-11-10
BZ-1279821 - Synchronization of a deployment unit between EJB service…

mswiderski:
…s app and Business Central fails

problem in that bz is caused by additional work item handler registration via deployment descriptor for ejb app to register async work item handlers. Since it is then stored as part of job description so it can be replicated to all server instances working with same db the special work item handler should not be saved. It is anyway registered always when deploying via ejb services so no need to persist it.


mrietveld:
+1 looks fine, except for the following: 

https://github.com/droolsjbpm/jbpm/pull/329/files?w=1#diff-767386a46794ed909523818e3df3248aR49

Could you make sure that this works? (A test would be the best, but a manual confirmation is also fine. Otherwise, one other way to solve this would be to return a singleton/static final value. That might be safer (and more performant?).

mrietveld:
(ignore my comment, I see that the tests cover the xstream converter code).

mrietveld:
+1 for merging.

mrietveld:
Please make sure to add documentation.

mswiderski:
:) it does work that's why we have this https://github.com/droolsjbpm/jbpm/pull/329/files?w=1#diff-203fce2ba37f5438045bd742896c6ebfR317
thanks for instant review @mrietveld

PR #332, opened 2015-11-12
JBPM-4824 Same SQL commands from script jbpm-6.0-to-6.1.sql are present also in jbpm-6.1-to-6.2.sql DB2 script

baldimir:
- Removed duplicate SQL commands.

If it is needed I will create PR for 6.3.x.


mswiderski:
+1
will proceed with merge in a while

I guess it's only community ddl that is included here so I don't think we need to back port it to 6.3.x as next community will go from master

PR #335, opened 2015-11-12
JBPM-4826 DB2 upgrade script bpms-6.0-to-6.1.sql doesn't change SessionInfo.id datatype properly if there are no sessions stored

baldimir:
- Added NVL function so the counter have always some value.


mswiderski:
+1
will proceed with merge in a while

mswiderski:
merged into master, thanks

PR #337, opened 2015-11-12
kie-services: added InMemoryFormSkeletonProvider

ibek:
to be able to create form services in KIE Server

### DIFF
--- a/jbpm-services/jbpm-kie-services/pom.xml
+++ b/jbpm-services/jbpm-kie-services/pom.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
-<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
-         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+<project
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+  xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
   <modelVersion>4.0.0</modelVersion>
   <parent>
     <groupId>org.jbpm</groupId>
@@ -48,7 +49,7 @@
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-human-task-core</artifactId>
     </dependency>
-     <dependency>
+    <dependency>
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-human-task-audit</artifactId>
     </dependency>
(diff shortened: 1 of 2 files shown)


mswiderski:
@ibek do you mind creating jira so we can collect thoughts if this is something we would like to add?

jbpm-services/jbpm-kie-services/pom.xml
```diff
@@ -145,6 +145,11 @@
       <artifactId>reflections</artifactId>
     </dependency>
     
+	<dependency>
+	  <groupId>commons-lang</groupId>
+      <artifactId>commons-lang</artifactId>
+    </dependency>
```
psiroky: Looks like `tab`s somehow got there...

ibek: @mswiderski ok, I will get to it next week ... but if the discussion around that took too much time, I would like to see at least this file being updated [1], so we could create own KIE extension with the form service.

Thanks

[1] https://github.com/droolsjbpm/droolsjbpm-integration/pull/229/files#diff-14503dabad0998300652edd66e5b2052R269

ibek:
@mswiderski so I created the JIRA for the form operations ... see https://issues.jboss.org/browse/JBPM-4832

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/form/provider/InMemoryFormSkeletonProvider.java
```diff
@@ -0,0 +1,181 @@
...
+            
+            String lang = (String) renderContext.get("lang");
+            
+            templateString = filterXML(templateString, lang, inputs);
+            
+            return templateString;
+        }
+    }
+    
+    private String filterXML(String document, String lang, Map inputs) {
```
mswiderski: can you provide bit of an explanation what this method is for? what kind of filtering do we expect to have?

mswiderski:
we certainly need to have some tests here.

since there is another form provider for .form that comes with form modeler won't it crash in anyway meaning we intercept the default .form provider features?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/form/provider/InMemoryFormSkeletonProvider.java
ibek: there two types of filtering - filtering of the inputBinding where we replace the location (user/name) for its value (Ivo) .... the second is language filtering to situations when we need a form in a specific language, so all titles, names, labels, error messages, ... are presented only in this specific language because I believe there is no need to send all language strings to the client.

ibek:
sure, I will write some tests there.

Are all providers registered automatically in Workbench? Then it could be a problem which we would fix by lowering the priority so in Workbench, this provider would never be used. The aim is to have this skeleton provider mainly in KIE Server.

mswiderski:
@ibek closing this as we discussed, there will be new PR that builds this support around KIE Server extensions

PR #339, opened 2015-11-16
Register task service instead of kie session twice

selrahal:


mswiderski:
@selrahal do you mind adding a test for this to make sure we catch any regressions there? and jira would be good as well so we have issue tracked as well. Thanks

selrahal:
@mswiderski updated with test

mswiderski:
ok to test

selrahal:
@mswiderski Can you take a look at the failure? I can't reproduce that locally :/

psiroky:
The failure is likely not related to your changes. It is just the tests are not stable and are failing on some environments sometimes (flakey tests).

I saw that the changes contain tabs for indenting, would it be possible to change that to 4 spaces to follow our code style?

PR #340, opened 2015-11-16
BZ-1281864 - Marshalling error accessing remote EJB services app with…

mswiderski:
… Sybase DB

there is no direct test for it as it is already covered (and was actually found) by the existing test once configured with sybase db which is done by QE.


mrietveld:
Just for my information, could you explain what Sybase was doing that caused this bug? If this problem comes up again, it'll be helpful for me to know what was actually happening. Thanks!

mswiderski:
Sybase returns impl class for data/timestamp which is not available on client side - com.sybase.jdbc4.tds.SybTimestamp while all other drivers provide an actual java.util class which is always available.

mrietveld:
+1 Thanks!

PR #342, opened 2015-11-19
JBPM-4841 - Use VariableIndexer for process variables

mswiderski:
depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/108


mrietveld:
+1

Minor comment: we're starting to use the word "indexer" a lot, maybe "persister" would be a better match? 

However, feel free to ignore this suggestion.

mswiderski:
it's actually only doing the indexation/transformation and does not really do any persistence so I'd opt to stay with Indexer

mrietveld:
makes sense. thanks.

PR #343, opened 2015-11-19
JBPM-4843 SqlServer upgrade scripts - 2 statements miss semicolon at the end of them

baldimir:
- Added missing semicolons.


mrietveld:
+1

H2 has various compatibility modes (including mysql and SQL Server: www.h2database.com/html/features.html ). 

If it's possible to create basic validation tests for the upgrade scripts using that, that would be great.

baldimir:
Yes Marco, you're right. Will look into that after we release 6.2.

PR #344, opened 2015-11-19
JBPM-4843 SqlServer upgrade scripts - 2 statements miss semicolon at the end of them

baldimir:
- Added missing semicolons.
(cherry picked from commit 638f0a2)

Master PR: https://github.com/droolsjbpm/jbpm/pull/343

### DIFF
--- a/jbpm-installer/db/upgrade-scripts/sqlserver/bpms-6.0-to-6.1.sql
+++ b/jbpm-installer/db/upgrade-scripts/sqlserver/bpms-6.0-to-6.1.sql
@@ -46,9 +46,9 @@ AS
 	END
 GO
 -- Executing the procedure
-EXECUTE alter_table_session_info
+EXECUTE alter_table_session_info;
 -- Deleting the procedure to clean it from database
-DROP PROCEDURE alter_table_session_info
+DROP PROCEDURE alter_table_session_info;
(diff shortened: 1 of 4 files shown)


mrietveld:
+1

mswiderski:
unless we have a blocker bz for this I'd close it to avoid merge to 6.3.x anything that is not a blocker

baldimir:
This is related to this BZ [1], which is not a blocker, so this can be merged to later versions. Semicolons in MSSQL are currently not mandatory, but "not using them" is deprecated. See here [2]. You should search for "semicolon" in that article, because the list of deprecated features is pretty long. 

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1235091
[2] https://msdn.microsoft.com/en-us/library/ms143729%28v=sql.120%29.aspx

krisv:
This should indeed not (yet) go into 6.3.x branch, as we won't pick this up for 6.2.0 now.  It's suggested for 6.2.1 so might be useful to keep the PR open.

mswiderski:
@krisv @etirelli can this be now merged to 6.3.x or we are still in strict mode of what goes in?

mswiderski:
merged into 6.3.x, thanks @baldimir

PR #345, opened 2015-11-20
BZ-1279821 - Synchronization of a deployment unit between EJB service…

mswiderski:
…s app and Business Central fails - missing ejb deployment service change


etirelli:
@mswiderski Jenkins failed... false negative?

mswiderski:
@etirelli yes, it's false negative. there are still some tests that I need to somehow fix to avoid these false negatives... need to find some time for it

PR #347, opened 2015-11-25
annotated testNullVariableInScriptTaskProcess as @RequiresPersistence…

mswiderski:
… to avoid false alarms from builds

@psiroky as we discussed over email...


mswiderski:
any objections?

mrietveld:
+1

PR #350, opened 2015-11-27
JBPM-4852 TaskService.complete() throws NullPointerException without …

tkobayas:
…a meaningful message when a RuntimeManager with the deploymentId is not registered


tkobayas:
I think this test failure is not related to this PR.

mswiderski:
correct, test failure not related to this change +1

PR #351, opened 2015-11-27
BZ-1285815 - auto claim is performed even though multiple actors are …

mswiderski:
…specified in HT within swimlane


mrietveld:
+1

You're probably already submitting a doc PR, I'm guessing. :)

mrietveld:
Hmm.. please double-check test failures, the same tests are not failing on master, but may be related to the parallel builds test problem.

mswiderski:
these are still to be fixed (or removed if not possible to make them reliable) as from time to time they cause false alarms. I did execute complete test case locally to double check it did not break anything - no issues found.

mrietveld:
Thanks! +1 for merging then.

PR #352, opened 2015-11-30
Add GO delimiter to Sybase DLL scripts

livthomas:
It allows you to create all tables in a single batch now.

The commit also fixes line separators since CRLFs (Windows) were used before.


mswiderski:
test failures not related to the change.
looks good to me, @baldimir what do you think?

baldimir:
+1, looks fine.

PR #353, opened 2015-12-01
BZ-1287065 - WAS: stopping Business Central from admin console result…

mswiderski:
…s into an error in server log

simply catch the exception to avoid nasty stack traces in the logs as if it was either canceled or expired this is exactly what pre destroy methods aims for


mrietveld:
Please add a test, even if it's just a very simple one. Thanks!

mswiderski:
@mrietveld frankly speaking not sure what kind of test would be meaningful here to test the try/catch addition... especially that the Timer is ejb instance, same as the exception type we catch. So to have this tested we would have to have full ejb container and somehow cancel timer externally and then attempt to shutdown this service. if you have any idea how to achieve in unit test I'd be more than happy to add a test

mrietveld:
Oops, sorry! I think I was looking at a different pull request when I commented here. +1 for merge.

PR #355, opened 2015-12-02
BZ-1276350 - added test for ordering task comments

jakubschwan:
This test is for control of right order of comments (by ID). Sometimes were comments unsorted in random order.


mswiderski:
+1

PR #357, opened 2015-12-04
BZ-1288171 - RuntimeDataServiceImpl. getProcessInstanceHistory return…

mswiderski:
…ing incorrect active nodes for tasks inside ad-hoc subprocesses

this is just test to cover the use of adhoc subprocess with history checking


mrietveld:
+1

PR #359, opened 2015-12-08
Reduce output when running tests

psiroky:
* these changes reduce the test output
   by ~75 %

Before the changes, the whole output (clean install) was about 17MiB, which is far too much. Now it is about 4MiB.


mswiderski:
+1

psiroky:
I forgot to commit newly added `logback-test.xml` in jbpm-installer, fixed now.

PR #361, opened 2015-12-09
jbpm-human-task-core: 'fixing' a JSON (task) content serialization test

mrietveld:
This fixes the failing "org.jbpm.services.task.jaxb.JsonTaskSerializationTest.jaxbContentTest". 

Technically, this uncovers a bug of sorts (or is it a unimplemented feature?). However, it's not something we're going to fix, because fixing this requires using specialized JSON marshalling features. 

In the past (6.0?), we started using these specialized JSON marshalling features, but they end up causing too many problems because clients then also need to use the same features. Especially for Web Developers who are doing REST API calls via JavaScript, using these marshalling features makes their life more difficult, not to mention Java developers who then have to figure out exactly what the right configuration is for their JSON (un)marshallers. 

Because of this, I'm simply "fixing" the test so that it will pass.


mrietveld:
@mswiderski This fixes the test you e-mailed me about.

mswiderski:
+1, thanks @mrietveld will merge in few minutes

PR #363, opened 2015-12-10
[6.3.x] Reduce output when running tests

psiroky:
* these changes reduce the test output
   by 70 %

Same changes already merged into master. These are only test-related changes, so I believe it is safe to backport them into 6.3.x.


mswiderski:
merged into 6.3.x, thanks @psiroky

PR #364, opened 2015-12-11
BPMSPL-121 - Tests for UpdateTimerCommand

jakubschwan:
Created tests for UpdateTimerCommand with processes.

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/TimerUpdateTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/TimerUpdateTest.java
@@ -17,12 +17,10 @@ package org.jbpm.test.functional.timer;
 
 import java.util.ArrayList;
 import java.util.List;
+import org.assertj.core.api.Assertions;
 import org.jbpm.process.instance.command.UpdateTimerCommand;
 import org.jbpm.test.JbpmTestCase;
 import org.jbpm.test.listener.CountDownProcessEventListener;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
 import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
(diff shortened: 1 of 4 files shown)


triceo:
Due to the extensive use of timers, the test is going to be extremely unstable. I suggest you find a different way to do this, perhaps using some callback from the process itself.

psiroky:
Yes, we really can not use `Thread.sleep` in the tests. I believe there is a way to use a specific listeners which do get notified when the timer node gets executed. @mswiderski knows more.

mswiderski:
totally agree, we should avoid Thread.sleep completely and try to utilize listeners that are backed by CountDownLatch to be more reliable. An example here: https://github.com/droolsjbpm/jbpm/blob/master/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/StartEventTest.java#L129

mswiderski:
@jakubschwan do you mind updating the tests to apply mentioned enhancements?

jakubschwan:
Yes, of course. I'm working on it. And thanks for advice

jakubschwan:
Changed tests. There are skipped tests for StartTimer and BoundaryTimer. In this cases UpdateTimerCommand didn't work (didn't change timer), so tests for StartTimer failed and test for BoundaryTimer expired.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/TimerUpdateTest.java
```diff
@@ -0,0 +1,309 @@
...
+        kieSession.addEventListener(countDownListener);
+        final List<Long> list = new ArrayList<Long>();
+        kieSession.addEventListener(new DefaultProcessEventListener() {
+            @Override
+            public void beforeProcessStarted(ProcessStartedEvent event) {
+                list.add(event.getProcessInstance().getId());
+            }
+        });
+
+        assertEquals(0, list.size());
```
triceo: I believe that the project already uses AssertJ in places. Unless the others mind, I'd like this to be ```Assertions.assertThat(list).isEmpty()```. Much clearer in the code, and a very descriptive error message when it fails.

@mswiderski Do you agree?

(Ditto with every other assertion.)

mswiderski: agreed

jakubschwan: Changed and added to PR

mswiderski:
fixed the boundary event and removed start event tests - see comment in https://github.com/droolsjbpm/jbpm/pull/370 

closing this as https://github.com/droolsjbpm/jbpm/pull/370 includes both commits for tests and fix for boundary timer event.

PR #365, opened 2015-12-14
BZ-1281864 - Marshalling error accessing remote EJB services app with…

mswiderski:
… Sybase DB - revert this wrapping Date objects as sybase driver is required to be present on client side to ensure correctness of the remote invocation

This reverts commit 78a35ca456c5b6c4afcd6eb1551023147f1cd9da.


mrietveld:
+1

PR #366, opened 2015-12-15
JBPM-4880 - Create an artifact with the DDL (SQL) files needed for different database schema's

mrietveld:
It seems to me like this would be useful? 

@dvirgiln We talked briefly about this in september, I think?

### DIFF
--- a/jbpm-installer/pom.xml
+++ b/jbpm-installer/pom.xml
@@ -119,14 +119,6 @@
   <build>
     <testResources>
       <testResource>
-        <directory>db</directory>
-        <filtering>false</filtering>
-        <includes>
-          <include>**/*.sql</include>
-          <include>**/*.SQL</include>
-        </includes>
-      </testResource>
-      <testResource>
         <directory>src/test/resources</directory>
         <filtering>false</filtering>
       </testResource>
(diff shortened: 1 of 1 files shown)


mswiderski:
Marion, can you elaborate more on what this should do? I looked at jira but didn't find details.... thanks

manstis:
Is it the weekend already? I thought Marco preserved the use of "Marion"
until he was more relaxed ;-)

Marion, can you elaborate more on what this should do? I looked at jira but
didn't find details.... thanks

—
Reply to this email directly or view it on GitHub
<https://github.com/droolsjbpm/jbpm/pull/366#issuecomment-164927072>.

mswiderski:
that's just my too fat fingers trying to type late evening... it should be Marco of course, sorry for that :)

mrietveld:
I stuffed Marco in the closet for bad behavior. I'll let him out later after I do all the fun work. 

Marco talked with David Virgil Naranjo (of Fuse/Switchyard fame) in September about the fact that when Fuse does integration tests (with our stuff), it would be nice if they could just grab the DDL scripts to make the db schema for their tests. I believe it's also useful for them as part of their setup in the product/project. 

Does that make sense?

mswiderski:
all makes perfect sense, thanks

+1

PR #368, opened 2015-12-17
improve test stability

mswiderski:
this aims at making jbpm repository build (mainly async and timer based tests) to become more stable. still didn't manage to achieve 100% stability though it looks way better than before so a step forward.

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/ConcurrentHumanTaskTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/ConcurrentHumanTaskTest.java
@@ -19,16 +19,19 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.concurrent.CountDownLatch;
 
 import javax.persistence.EntityManagerFactory;
 
 import org.jbpm.process.audit.AuditLogService;
 import org.jbpm.process.audit.JPAAuditLogService;
 import org.jbpm.process.core.timer.TimerServiceRegistry;
+import org.jbpm.services.task.HumanTaskServiceFactory;
 import org.jbpm.services.task.identity.JBossUserGroupCallbackImpl;
 import org.jbpm.services.task.utils.ContentMarshallerHelper;
 import org.jbpm.test.JbpmTestCase;
 import org.junit.Assert;
+import org.junit.Before;
 import org.junit.Test;
 import org.kie.api.KieBase;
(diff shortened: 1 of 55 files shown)


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/ConcurrentHumanTaskTest.java
```diff
@@ -57,18 +62,33 @@ public ConcurrentHumanTaskTest() {
...
+	    ((InternalTaskService)taskService).addUser(TaskModelProvider.getFactory().newUser("john"));
+	    ((InternalTaskService)taskService).addUser(TaskModelProvider.getFactory().newUser("Administrator"));
+	    
+	    ((InternalTaskService)taskService).addGroup(TaskModelProvider.getFactory().newGroup("sales"));
+	    ((InternalTaskService)taskService).addGroup(TaskModelProvider.getFactory().newGroup("PM"));
+	    ((InternalTaskService)taskService).addGroup(TaskModelProvider.getFactory().newGroup("Administrators"));
+	}
 
-	@Test
+	@Test(timeout=10000)
```
mrietveld: I'd leave the timeout in, it can't hurt... (and if for some reason this does hang because of future changes, the build will then not hang forever.)

mrietveld:
+1 when the build completes. 

There are minor fixes that we can do later (duplicate code, etc): it's more important to simply get this code in first and improve it later.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/ConcurrentHumanTaskTest.java
mswiderski: not sure I understand, timeout was actually added here not removed...

mrietveld: Oops! It's still early morning here. :/ 

Ignore my comment! Deleting it now..

mswiderski: no worries, thanks for reviewing!

build finished and it's green :)

psiroky:
@mswiderski I think it would be great if we could apply these changes to 6.3.x as well.

mswiderski:
completely fine with me, though still don't know what's the situation for 6.3.x.... @krisv @etirelli

psiroky:
It seems these are only test-related changes, so should be fine to backport them.

PR #370, opened 2015-12-18
JBPM-4886 - Resetting timer cycle in a running workflow process

mswiderski:
replaces https://github.com/droolsjbpm/jbpm/pull/364 as this incorporates https://github.com/droolsjbpm/jbpm/pull/364 and fix for ignored test. 

start timer tests have been removed as they are not in scope as UpdateTimerCommand is only for active process instances and start timer event is not on active process instance as it starts the instance by timer  event. That means when process instance is started timer has already fired.


mrietveld:
+1 Nice tests!

PR #373, opened 2015-12-21
BZ-1274515 - [Drools][Quartz] The number of retries of a failed task …

mswiderski:
…are fixed

as you can see there is no test as it very (and I mean very) difficult to provide one that you can read about in the BZ comments. Though it was verified both by QE/GSS and the customer that it solves the issue


mrietveld:
If this were my BZ, I would add a test that makes sure that `jobq.setRequestsRecovery(true)` is called: while this may seem pointless, it does ensure that the build will fail if a developer removes this code in the future. 

You could test this by mocking the `scheduler` field and then  getting`jobq` argument (via Mockito: `ArgumentCaptor` + `verify`) from it.  

The test then serves as a sort of documentation for other developers later. Another option is to add documentation to the code itself, although this would make the code less readable IMHO. 

However, that's just how *I* would do things in cases like this, and definitely not per se the "right" thing to do. :) 

+1 for whatever you want to do (leave it like this or add a test).

mswiderski:
thanks @mrietveld will look into some tests and update pr

PR #374, opened 2016-01-07
JBPM-4899 - Plugable notification listeners for human task deadlines

mswiderski:
provides pluggable mechanism to allow custom handling of human task notifications (BPMSPL-172)

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2012 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.services.task.deadlines.notifications.impl;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.ServiceLoader;
+
+import org.jbpm.services.task.deadlines.NotificationListener;
+import org.kie.internal.task.api.UserInfo;
+import org.kie.internal.task.api.model.NotificationEvent;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Manages broadcasting of notification events to all found listeners 
+ *
+ */
+public class NotificationListenerManager {
+    
+    private static final Logger logger = LoggerFactory.getLogger(NotificationListenerManager.class);
+    
+    private static ServiceLoader<NotificationListener> listenersLoaded = ServiceLoader.load(NotificationListener.class);
+    
+    private static NotificationListenerManager INSTANCE = new NotificationListenerManager();
+    
+    private List<NotificationListener> listeners = new ArrayList<NotificationListener>();
+    
+    private NotificationListenerManager() {
+        for (NotificationListener listener : listenersLoaded) {
+            listeners.add(listener);
+        }        
+    }
+
+    /**
+     * Broadcast given event to all listeners independently meaning catches possible exceptions to 
+     * avoid breaking notification by listeners
+     * @param event notification event to be sent
+     * @param params additional parameters see NotificationListener.onNotification for details.
+     * 
+     * @see NotificationListener#onNotification(NotificationEvent, Object...)
+     */
+    public void broadcast(NotificationEvent event, UserInfo userInfo) {
+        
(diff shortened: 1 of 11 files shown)


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
```diff
@@ -60,19 +60,14 @@
...
     
-    public EmailNotificationListener(UserInfo userInfo) {
-    	this.userInfo = userInfo;
-    }
     
     @Override
-    public void onNotification(NotificationEvent event) {
-
+    public void onNotification(NotificationEvent event, Object... params) {
+        UserInfo userInfo = (UserInfo) params[0];
```
manstis: I admit, I don't know the code-base, but is this just waiting for a ```ClassCastException``` to happen at sometime in the future?

mswiderski: main reason I used varargs instead of direct UserInfo argument is that most likely we will have more parameters needed when providing custom types of notifications and I'd like to avoid changing the NotificationListener interface every time new param is added.
Though when more votes would be to keep it strict typed I will change it :) thanks @manstis

manstis: @mswiderski NP.. I was thinking more of an ```instanceof``` check here, than strictly typing the ```params```.. I'll leave it in your (more than) capable hands... (and be less nosey in the future!)

mswiderski: @manstis please don't stop the music ... seriously the more comments the better ... knowledge is everything :)

mrietveld: @mswiderski If you're expecting more types of parameters, I would maybe then either recommend a fluent interface (not sure that's possible here?) or having one of the arguments be a sort of DTO that could "carry" those extra types..

mswiderski: thanks @mrietveld, the more I think about it the more I am convinced this was not the best idea :) I'll stick to UserInfo only as this is rather important from notification point of view and more parameters can be given as variables of NotificationEvent class. How does that sound?

mrietveld: @mswiderski That sounds good. I was also thinking along similar lines.. :)

mswiderski:
jenkins retest this

mswiderski:
updates about the parameter/userinfo done

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
```diff
@@ -0,0 +1,76 @@
...
+            } catch (Exception e) {
+                logger.warn("Exception encountered while notifying listener {} with event {} - error {}", 
+                        listener, event, e.getMessage());
+            }
+        }
+    }
+    
+    public static NotificationListenerManager get() {
+        if (INSTANCE == null) {
+            INSTANCE = new NotificationListenerManager();
```
mrietveld: If you want to lazy-load the manager instance, would it be okay to use the [lazy-holder-idiom](https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom)? I believe that the above code is a minor race condition.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
```diff
@@ -60,19 +60,14 @@
...
-    public EmailNotificationListener(UserInfo userInfo) {
-    	this.userInfo = userInfo;
-    }
     
     @Override
-    public void onNotification(NotificationEvent event) {
-
+    public void onNotification(NotificationEvent event, UserInfo userInfo) {
+        
         if (userInfo == null || mailSession == null) {
```
mrietveld: It looks like the `mailSession` variable could be set to final, so that the `mailSession` null check would not be needed. Does that make sense?

mswiderski: it still can be null as it would then mean this (EmailNotificationListener) listener is disabled

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
mswiderski: it does not have to be lazy, so I guess I could simply move the new NotificationListenerManager() to be directly on the field, right?

mrietveld:
+1 except for the notification listener manager initialization. Changing the `mailSession` variable (finall/null check) is up to you.

PR #375, opened 2016-01-08
GUVNOR-2173 - Asset Oracle: Change Impact prediction

mrietveld:
In short, what's important is that `BaseDescr` instances that jbpm-flow-builder generates do _not_ have `null` `.resource` fields. 

This is in order to map the type references collected from a generated, compiled (java) file to the original actual resource/asset (bpmn2/DRL/etc.).

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/CompilationTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/CompilationTest.java
@@ -1,0 +1,103 @@
+package org.jbpm.bpmn2;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.drools.compiler.compiler.AnalysisResult;
+import org.drools.compiler.compiler.ReturnValueDescr;
+import org.drools.compiler.lang.descr.ActionDescr;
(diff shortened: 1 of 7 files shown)


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/CompilationTest.java
```diff
@@ -0,0 +1,103 @@
+package org.jbpm.bpmn2;
```
manstis: License!

manstis:
@mrietveld IDK what I'm reviewing, other than the missing license, so leave a "full review" to somebody else :)

mswiderski:
is this to show the source where the potential issue is found like in this or that source file? if so wondering if we could have another test that illustrates this ... though not sure it's possible in jbpm repo...

besides that tiny comment +1

mrietveld:
@mswiderski The issue (logic requires `descr.getResource()` to be non-null) is in code I'm writing in drools-compiler for GUVNOR-2135. I will make sure to add a test there that tests this.

mrietveld:
@manstis That has never stopped me before! (Not knowing what I was reviewing.. ) ;)

PR #377, opened 2016-01-18
JBPM-4889 - Process instance variables: Variable content is trimmed, …

mswiderski:
…maximum is 255 chars


mrietveld:
@mswiderski Nice! If you could add documentation for this, that would be great. :) 

(Ignore this if you've already done that? :) ) 

Thanks.

PR #378, opened 2016-01-18
Fix regex in test to allow more types of versions strings

psiroky:
Same fix already applied to jbpm-console-ng, https://github.com/droolsjbpm/jbpm-console-ng/commit/49074353d96bad8cef4798c0f3902c8f36fb5379


mswiderski:
+1

psiroky:
Thanks @mswiderski!

PR #380, opened 2016-01-22
Remove unused jbpm-human-task-performance module

psiroky:


triceo:
Do we know who introduced the module and why it's not used anymore?

psiroky:
Based on history it seems it was @Salaboy by https://github.com/droolsjbpm/jbpm/commit/0593ec226ab4f85adc28e0cb92114b5c589d6a5e. Not sure why it is not used anymore, maybe it was meant just a local thing and should not have landed in the upstream repo?

mswiderski:
I think it was used when introducing queries based on task audit tables. It was to make sure it won't degrade performance and to my knowledge was temporary only.

build failure not related, will merge this later today unless @Salaboy would like to keep it.

mswiderski:
merged into master and 6.4.x, thanks @psiroky

PR #383, opened 2016-01-27
JBPM-4910 - Deployed process ignores Runtime Strategy set in deployment descriptor, in managed repository project

mswiderski:
@mrietveld since this (JBPM-4910) depends on fixes from https://github.com/droolsjbpm/jbpm/pull/382 I did fix in in separate commit as part of this PR. Could you please double check it's valid?


mrietveld:
You read my mind, @mswiderski ! Although there are one or two things I'd do differently. 

After the meeting, I'll update the PR.

PR #384, opened 2016-01-27
JBPM-4913: Ability to read the underlying JNDI datasource name (6.4.x)

dgutierr:
This feature is required by the jBPM console in order to retrieve from the jBPM engine what is the data source JNDI name used by the datasets that feed the UI consoles.

### DIFF
--- a/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/settings/JpaSettings.java
+++ b/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/settings/JpaSettings.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2016 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *       http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/settings/JpaSettings.java
```diff
@@ -0,0 +1,79 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class JpaSettings {
+
+    static JpaSettings INSTANCE = null;
+
+    public static JpaSettings get() {
+        if (INSTANCE == null) {
+            INSTANCE = new JpaSettings();
```
mswiderski: can we avoid this null check and instantiate it on fields directly (+private constructor if we don't want to create it in any other way) - it's potential race condition 

if we need that for tests we could have it package scoped reset method

mrietveld: https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom? 

And leave the field non-final to allow test changes, if necessary.

dgutierr:
Using PowerMockito I managed the mock both the JpaSetting's singleton private instance and the calls to its private methods. Another benefit of this approach is that I didn't have to change JpaSettings to make it testable.

mswiderski:
+1

mswiderski:
merged int 6.4.x with https://github.com/droolsjbpm/jbpm/commit/ff6eb996ed30ac05962a97fb72d22196d230c528 thanks @dgutierr!

PR #382, opened 2016-01-26
BZ-1125077 - upgrading code to default behavior in 7.x

mrietveld:
@mswiderski Could you take a look when you get a chance? Thanks!

PR #385, opened 2016-01-27
JBPM-4913: Ability to read the underlying JNDI datasource name

dgutierr:
This feature is required by the jBPM console in order to retrieve from the jBPM engine what is the data source JNDI name used by the datasets that feed the UI consoles.


mswiderski:
+1, test failure not related and it going to be fixed with #382

PR #387, opened 2016-01-27
JBPM-4900 - Enhance TaskList Search capability - query service based …

mswiderski:
…on dashbuilder data sets

depends on https://github.com/droolsjbpm/drools/pull/620

@mrietveld  @krisv  - if you find spare minutes/hours/days please have a look

here you can find description of this work:
http://mswiderski.blogspot.com/2016/01/advanced-queries-in-jbpm-64.html


mrietveld:
@mswiderski I will go through this tomorrow morning. However, it might take me until monday to go through all of it. Is that okay?

mswiderski:
Sure thing, take the time you need

Dnia 27.01.2016 o godz. 20:39 Marco Rietveld <notifications@github.com> napisał(a):

mrietveld:
@mswiderski Could you separate the fix for the tests from the commits for this PR?

mswiderski:
I'll cherry pick that fix to master so it will be removed from this once in there.

mswiderski:
done, single commit, only for the feature in subject

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/persistence/QueryDefinitionEntity.java
```diff
@@ -0,0 +1,126 @@
...
+@SequenceGenerator(name="queryDefIdSeq", sequenceName="QUERY_DEF_ID_SEQ", allocationSize=1)
+public class QueryDefinitionEntity {
+    
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO, generator="queryDefIdSeq")
+    @Column(name = "id")
+    private Long id;
+ 
+    @Lob
+    @Column(name="Q_EXPRESSION", length=65535)
```
mrietveld: @mswiderski If possible, it would be better to keep the same naming conventions that we've used for all of our other tables: in other words, camel case starting with a lowercase letter. 

I can understand wanting to make the names different because they might be special, but IMHO, this just leads to more effort in the long run.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/persistence/QueryDefinitionEntity.java
```diff
@@ -0,0 +1,126 @@
...
+
+import org.jbpm.kie.services.impl.query.SqlQueryDefinition;
+import org.jbpm.services.api.query.model.QueryDefinition;
+import org.jbpm.services.api.query.model.QueryDefinition.Target;
+
+@NamedQueries(value={
+        @NamedQuery(name="getQueryDefinitions", query="select qd from QueryDefinitionEntity qd"),
+        @NamedQuery(name="getQueryDefinitionByName", query="select qd from QueryDefinitionEntity qd where name = :name")
+})
+@Entity
```
mrietveld: I've just noticed that this is the 3rd use of `@NamedQueries` that we have: 
- DeploymentStoreEntry
- ContextMappingInfo

For maintainability and readability, I would much rather have **all** of our queries either be via annotation **or* in xml. Again, if we were a small project, this wouldn't matter. But we're not -- there are thousands and thousands of LoC. 

IMHO, choosing one way to do things (xml via anno's) is part of saving energy so that we can spend it on the important issues. 

Please move these to an xml config -- or, if you think we should do everything via anno's, then that's also fine, but then we need to do that as well.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
```diff
@@ -176,7 +176,7 @@ public void deploy(DeploymentUnit unit) {
     	        	processResources(depModule, files, kieContainer, kmoduleUnit, deployedUnit, depModule.getReleaseId(), processDescriptors);
     	        }
             }
-            Collection<ReleaseId> dependencies = module.getJarDependencies(new DependencyFilter.ExcludeScopeFilter("test"));
+            Collection<ReleaseId> dependencies = module.getJarDependencies(new DependencyFilter.ExcludeScopeFilter("test", "provided"));
```
mrietveld: Doesn't this have a pretty large impact, as it also impacts how *all* kjar's are built? 

Could you explain why this is needed for this PR?

mswiderski: it was overlooked from the time when it was added, as it should always exclude test and provided scope. I discovered it while working on this when you define dependency to jbpm/drools dependencies e.g. jbpm-kie-services or jbpm-services-api to implement custom mappers or query param builders. Then it scans and attaches jbpm-kie-services and all its transitive deps into project class loader ... which is time consuming and unnecessary so decided to fix it here. Can be changed as separate commit too if we think that will be better from visibility point of view

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/persistence/QueryDefinitionEntity.java
mswiderski: main idea was that you don't need any other orm file to be added to your app to gain possibility to use entity and its queries in given app. orm files are not automatically discovered while entities are. But since we already have services orm file anyway that needs to be used it can be moved. 

what do you say if I merge it with this and then do post merge clean up to move both QueryDefinition and DeploymentStoreEntry as both exist in jbpm-kie-services. Will leave ContextMappingInfo as runtime manager does not have dedicated orm file. Thoughts?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/persistence/QueryDefinitionEntity.java
mswiderski: main idea was that you don't need any other orm file to be added to your app to gain possibility to use entity and its queries in given app. orm files are not automatically discovered while entities are. But since we already have services orm file anyway that needs to be used it can be moved. 

what do you say if I merge it with this and then do post merge clean up to move both QueryDefinition and DeploymentStoreEntry as both exist in jbpm-kie-services. Will leave ContextMappingInfo as runtime manager does not have dedicated orm file. Thoughts?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
mrietveld: Ah!! Yes, please -- if you could change it to a different commit, that would be great.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/persistence/QueryDefinitionEntity.java
mrietveld: That sounds good, thank you!

mrietveld:
+1 Nice work, Maciej!

mswiderski:
@mrietveld modifications added and separated into 3 commits

mrietveld:
+1 thanks!

krisv:
+1, great work

PR #388, opened 2016-01-29
JBPM-4303 - using a singleton EventListener instead of empty implementations

mrietveld:
@mswiderski Could you look at this? 

I talked with @krisv about this last night: it's a minor change that, even in the current implementations, is a little more readable (and faster, I believe, not that that matters really). 

It also helps me with the stackless work, which is my main reason for adding it. 

Thanks!

The changes can be seen [here](https://github.com/droolsjbpm/jbpm/pull/388/files?w=1). In short: 

- Remove empty implementations
- Add a singleton/constant "empty event listener" impl that can be used instead

Apologies for the space changes.. :/

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
@@ -16,7 +16,8 @@
 
 package org.jbpm.workflow.instance.impl;
 
-import java.io.Serializable;
+import static org.jbpm.workflow.instance.impl.DummyEventListener.EMPTY_EVENT_LISTENER;
+
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -78,6 +79,7 @@ public abstract class WorkflowProcessInstanceImpl extends ProcessInstanceImpl
 
 	private Map<String, List<EventListener>> eventListeners = new HashMap<String, List<EventListener>>();
 	private Map<String, List<EventListener>> externalEventListeners = new HashMap<String, List<EventListener>>();
+
 	private List<String> completedNodeIds = new ArrayList<String>();
 	private List<String> activatingNodeIds;
 	private Map<String, Integer> iterationLevels = new HashMap<String, Integer>();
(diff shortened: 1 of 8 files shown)


jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -79,39 +79,47 @@
...
 	private List<String> activatingNodeIds;
 	private Map<String, Integer> iterationLevels = new HashMap<String, Integer>();
 	private int currentLevel;
 	private boolean persisted = false;
 	private Object faultData;
-	
+
 	private boolean signalCompletion = true;
 
+	public static final EventListener EMPTY_EVENT_LISTENER = new EventListener() {
```
mswiderski: could we externalize this class so we could avoid import of WorkflowProcessInstanceImpl in here: https://github.com/droolsjbpm/jbpm/pull/388/files?w=1#diff-625e8789a66222e0e5be169c3f8c5af9R19

mswiderski:
if it does not have any side effects (which I cannot find any) it's fine with me. just one tiny comment about relationship between node and workflowprocessimpl

mswiderski:
two test cases ...NoIO fails after move to 8 as it seems it revealed an issues with not clearing parser state between processes, fixed in here https://github.com/mswiderski/jbpm/commit/9ba4411300fddff432073c09e6f4e4786c96f20a

the other one seems to be intermittent failure

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
mrietveld: Hi @mswiderski, I'm not sure I understand the problem: the import of `WorkflowProcessInstanceImpl` will be there regardless of whether or not it's a member class or not. 

The import is importing the static field, not the class..

mswiderski: @mrietveld what I mean is to move this EventListener to a public class (separate file) that will have only single instance and shall be accessed via static method only (so private constructor only) and then referenced from both WorkflowProcessInstanceImpl and EventNode which will give use no dependency of CompositeNodeInstance and EventNodeInstance to WorkflowProcessInstanceImpl

though this is just a minor thing

mrietveld:
Jenkins, please retest.

mrietveld:
@mswiderski Updated so that it's in a different class.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventNodeInstance.java
```diff
@@ -31,19 +32,18 @@
 import org.jbpm.workflow.core.node.EventNode;
 import org.jbpm.workflow.instance.impl.ExtendedNodeInstanceImpl;
 import org.jbpm.workflow.instance.impl.NodeInstanceResolverFactory;
+import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
```
mswiderski: this seems to be unused import

mswiderski:
excellent, that's exactly what I had in mind. thank you! once the build is over I'll merge it

mswiderski:
@mrietveld looks like #389 caused conflicts here. I tried to resolve them but there were quite few so if you don't mind rebase with master so I could merge this one... that would be great

mrietveld:
Rebased. (retesting just to be sure). 

Jenkins, please retest.

mswiderski:
@mrietveld thanks

P.S.
if you amend the commit, or add anything to it, jenkins rebuilds automatically, no need to as it :)

PR #390, opened 2016-02-03
JBPM-4900 - Enhance TaskList Search capability

mswiderski:
depends on https://github.com/droolsjbpm/drools/pull/629

back port from master


mswiderski:
merged into 6.4.x

PR #391, opened 2016-02-04
DO NOT MERGE! This is for a discussion about JBPM-4304/Stackless

mrietveld:


mrietveld:
https://github.com/mrietveld/jbpm/blob/_stackless-3/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/FaultNodeInstance.java#L64-L74

mrietveld:
https://github.com/mrietveld/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/FaultNodeInstance.java#L62-L73

PR #392, opened 2016-02-04
- removed @author tags which cause "code ownership" which discourage teamwork and initiative

mrietveld:
I ran a perl one-liner through the jbpm-flow code.. 

@psiroky Could you take a look at this?


psiroky:
+1. Thanks @mrietveld. We should remove these in the whole jBPM repo and also all the other KIE repos. I'll create a Jira for that. It should not be hard to do (especially since the Perl one-liner is already done).

I'll merge this one now and we can follow up with the rest of the changes later.

PR #393, opened 2016-02-04
DO NOT MERGE! This is for a discussion about JBPM-4304/Stackless

mrietveld:
https://github.com/mrietveld/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/FaultNodeInstance.java#L62-L73

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ * http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
@@ -16,24 +16,94 @@
 
 package org.jbpm.ruleflow.instance;
 
+import org.drools.core.common.InternalKnowledgeRuntime;
+import org.drools.core.impl.StatefulKnowledgeSessionImpl;
+import org.jbpm.process.instance.ProcessInstance;
+import org.jbpm.process.instance.context.exception.ExceptionScopeInstance;
 import org.jbpm.ruleflow.core.RuleFlowProcess;
 import org.jbpm.workflow.core.node.StartNode;
 import org.jbpm.workflow.instance.NodeInstance;
 import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
+import org.jbpm.workflow.instance.impl.queue.RunProcessInstance;
+import org.jbpm.workflow.instance.node.EventNodeInstanceInterface;
+import org.kie.api.runtime.process.EventListener;
 
 public class RuleFlowProcessInstance extends WorkflowProcessInstanceImpl {
 
     private static final long serialVersionUID = 510l;
-    
+
+    private transient RunProcessInstance runProcessInstance = new RunProcessInstance();
+
+    public RuleFlowProcessInstance() {
+    }
+
     public RuleFlowProcess getRuleFlowProcess() {
         return (RuleFlowProcess) getProcess();
     }
 
-    public void internalStart(String trigger) {
-    	StartNode startNode = getRuleFlowProcess().getStart(trigger);
-    	if (startNode != null) {
-    		((NodeInstance) getNodeInstance(startNode)).trigger(null, null);
-    	}
+    // STACKLESS
+
+    public void internalStart( String trigger ) {
+        StartNode startNode = getRuleFlowProcess().getStart(trigger);
+        if( startNode != null ) {
+            NodeInstance startNodeInstance = ((NodeInstance) createNodeInstance(startNode));
+            if( isStackless() ) {
+                runProcessInstance.addNodeInstanceTrigger(startNodeInstance, null, null);
+                executeProcessInstance(runProcessInstance);
+            } else {
+                startNodeInstance.trigger(null, null);
+            }
+        }
+    }
+
+    // purely adding instances to an EXECUTING action
+
+    @Override
+    public void addNodeInstanceTrigger( NodeInstance nodeInstance, NodeInstance from, String type ) {
+        runProcessInstance.addNodeInstanceTrigger(nodeInstance, from, type);
+    }
+
+    @Override
+    public void removeNodeInstanceTrigger( NodeInstance nodeInstance ) {
+        runProcessInstance.removeNodeInstanceTrigger(nodeInstance);
+    }
+
+    // TRIGGER EXECUTION
+
+    @Override
+    public void triggerCompletedAndExecute(EventNodeInstanceInterface eventNodeInstance) {
+        eventNodeInstance.triggerCompleted();
+        executeProcessInstance(runProcessInstance);
+    }
+
+    @Override
+    public void signalEventAndExecute( EventListener eventListener, String type, Object event ) {
+        runProcessInstance.addAndTriggerSignalEventAction(eventListener, type, event);
+        executeProcessInstance(runProcessInstance);
+    }
+
+    // TODO: use this instead of triggerCompletedAndExecute
+    public void executeQueueIfNotAlreadyExecuting() {
+        executeProcessInstance(runProcessInstance);
+    }
+
+
+    // DBG
+    private final boolean debug = true;
+
+    private void executeProcessInstance(RunProcessInstance runProcessInstance) {
+        if( debug ) {
+            runProcessInstance.execute();
+        } else {
+            ((StatefulKnowledgeSessionImpl) getKnowledgeRuntime()).addPropagation(runProcessInstance);
+            InternalKnowledgeRuntime kRuntime = getKnowledgeRuntime();
+            try {
+                kRuntime.startOperation();
+                kRuntime.executeQueuedActions();
+            } finally {
+                kRuntime.endOperation();
+            }
+        }
     }
 
 }
--- a/jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
+++ b/jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
@@ -16,6 +16,8 @@
 
 package org.jbpm.ruleflow.instance;
 
+import static org.kie.api.runtime.EnvironmentName.USE_STACKLESS_EXECUTION;
+
 import java.io.Externalizable;
 import java.io.IOException;
 import java.io.ObjectInput;
@@ -23,13 +25,51 @@ import java.io.ObjectOutput;
 
 import org.jbpm.process.instance.AbstractProcessInstanceFactory;
 import org.jbpm.process.instance.ProcessInstance;
+import org.kie.api.runtime.Environment;
 
 public class RuleFlowProcessInstanceFactory extends AbstractProcessInstanceFactory implements Externalizable {
 
     private static final long serialVersionUID = 510l;
 
-    public ProcessInstance createProcessInstance() {
-        return new RuleFlowProcessInstance();
+    // OCRAM: rename to "queue-based"
+    protected static final boolean SYSTEM_STACKLESS_EXECUTION_PROPERTY = Boolean.getBoolean("org.jbpm.exec.queue");
+
+    public ProcessInstance createProcessInstance(Environment env) {
+        RuleFlowProcessInstance processInstance = new RuleFlowProcessInstance();
+        boolean stacklessExecution = useStacklessExecution(env);
+        processInstance.setStackless(stacklessExecution);
+        return processInstance;
+    }
+
+    /**
+     * Determining stackless execution:
+     *
+     * in 6.x, 7.x:  default is FALSE
+     * 1. if env == null: FALSE
+     * 2. else if env variable == true, TRUE
+     * 3. else if system prop == true, TRUE
+     *
+     * in 8.x: default is TRUE
+     * 1. if env == null -> TRUE
+     * 2. else if env variable == FALSE ->  FALSE
+     * 3. else if system prop == FASLE -> FASLE
+     *
+     * @param env
+     * @return a {@link boolean}
+     */
+    public static boolean useStacklessExecution(Environment env) {
+        boolean stacklessExecution;
+        if( env == null ) {
+            stacklessExecution = false;
+        } else {
+            Object stacklessBoolObj = env.get(USE_STACKLESS_EXECUTION);
+            if( stacklessBoolObj instanceof Boolean ) { // null check
+                stacklessExecution = (Boolean) stacklessBoolObj;
+            } else {
+                stacklessExecution = SYSTEM_STACKLESS_EXECUTION_PROPERTY;
+            }
+        }
+        return stacklessExecution;
     }
 
     public void readExternal(ObjectInput in) throws IOException, ClassNotFoundException {
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
@@ -17,14 +17,18 @@
 package org.jbpm.workflow.instance.impl;
 
 import java.io.Serializable;
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Deque;
 import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 
 import org.drools.core.common.InternalKnowledgeRuntime;
+import org.drools.core.impl.StatefulKnowledgeSessionImpl;
+import org.drools.core.phreak.PropagationEntry;
 import org.drools.core.spi.ProcessContext;
 import org.jbpm.process.core.Context;
 import org.jbpm.process.core.ContextContainer;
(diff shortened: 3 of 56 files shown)


jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
```diff
@@ -16,24 +16,94 @@
...
+        executeProcessInstance(runProcessInstance);
+    }
+
+
+    // DBG
+    private final boolean debug = true;
+
+    private void executeProcessInstance(RunProcessInstance runProcessInstance) {
+        if( debug ) {
+            runProcessInstance.execute();
```
mswiderski: out of curiosity, how is this useful as it does something else than when not in debug?

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
```diff
@@ -23,13 +25,51 @@
...
 import org.jbpm.process.instance.ProcessInstance;
+import org.kie.api.runtime.Environment;
 
 public class RuleFlowProcessInstanceFactory extends AbstractProcessInstanceFactory implements Externalizable {
 
     private static final long serialVersionUID = 510l;
 
-    public ProcessInstance createProcessInstance() {
-        return new RuleFlowProcessInstance();
+    // OCRAM: rename to "queue-based"
```
mswiderski: what does OCRAM stand for?

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
```diff
@@ -23,13 +25,51 @@
...
+
+    /**
+     * Determining stackless execution:
+     *
+     * in 6.x, 7.x:  default is FALSE
+     * 1. if env == null: FALSE
+     * 2. else if env variable == true, TRUE
+     * 3. else if system prop == true, TRUE
+     *
+     * in 8.x: default is TRUE
```
mswiderski: do we really want/need to support both modes in version 8 and onwards? I opt at supporting only one as it will be rather difficult to maintain and enhance while still supporting both modes.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
```diff
@@ -123,52 +131,49 @@ public Node getNode() {
...
-    
+
     public void cancel() {
         nodeInstanceContainer.removeNodeInstance(this);
-        boolean hidden = false;
-        Node node = getNode();
-    	if (node != null && node.getMetaData().get("hidden") != null) {
-    		hidden = true;
-    	}
+        if( isStackless() ) {
```
mswiderski: could we move this isStackless chekgin into the removeNodeInstanceTrigger method instead to have less cleanup work afterwards?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
```diff
@@ -257,113 +264,119 @@ protected void triggerCompleted(String type, boolean remove) {
...
-        }
-    }
-    
+            Deque<TriggerAction> nodeInstances = new ArrayDeque<TriggerAction>(connections.size());
+            for (Connection connection: connections) {
+                nodeInstances.add(new TriggerAction(followConnection(connection), connection.getToType()));
+            }
+            Iterator<TriggerAction> iter;
+            if( isStackless() ) {
+                iter = nodeInstances.descendingIterator();
```
mswiderski: similar here, could this be hidden by like a method in NodeInstanceImpl that simply returns Iterator and the change what kind is hidden in that method? That way it will be more readable and less cleanup when moving to single engine mode only.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
```diff
@@ -257,113 +264,119 @@ protected void triggerCompleted(String type, boolean remove) {
...
-    	// trigger next node
-        nodeInstance.trigger(this, type);
+        boolean hidden = isHidden();
+        InternalKnowledgeRuntime kruntime = getProcessInstance().getKnowledgeRuntime();
+        if (!hidden) {
+            ((InternalProcessRuntime) kruntime.getProcessRuntime())
+                .getProcessEventSupport().fireBeforeNodeLeft(this, kruntime);
+        }
+        if( isStackless() ) {
+            addNodeInstanceTrigger(nodeInstance, this, type);
```
mswiderski: could this be included in nodeInstance.trigger that decides if it's stalkless or not and perform actions accordingly?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -409,12 +410,17 @@ public void start(String trigger) {
...
 	        Node[] nodes = getNodeContainer().getNodes();
+	        // OCRAM: reverse order for stackless
 	        for (Node node : nodes) {
 	            if (node instanceof EventSubProcessNode) {
 	                Map<Timer, DroolsAction> timers = ((EventSubProcessNode) node).getTimers();
 	                if (timers != null && !timers.isEmpty()) {
-	                    EventSubProcessNodeInstance eventSubprocess = (EventSubProcessNodeInstance) getNodeInstance(node);
-	                    eventSubprocess.trigger(null, org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
+	                    EventSubProcessNodeInstance eventSubprocess = (EventSubProcessNodeInstance) createNodeInstance(node);
+	                    if( isStackless() ) {
```
mswiderski: same here, could eventSubprocess.trigger hide this check and decision if it stackless or not

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -492,11 +506,16 @@ public void signalEvent(String type, Object event) {
 			        }
 				}
 				if (((org.jbpm.workflow.core.WorkflowProcess) getWorkflowProcess()).isDynamic()) {
-					for (Node node : getWorkflowProcess().getNodes()) {
+				    Node [] nodes = getWorkflowProcess().getNodes();
+				    // OCRAM: UGLY
+				    boolean stackless = isStackless();
```
mswiderski: agreed :) could you please elaborate a bit more why this loop with limits is needed?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -648,4 +667,13 @@ public void setSignalCompletion(boolean signalCompletion) {
         this.signalCompletion = signalCompletion;
     }
 
+    public boolean isStackless() {
+//        new Throwable("** stackless?").printStackTrace();
```
mswiderski: I guess it's only for test/debug...

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/RunProcessInstance.java
```diff
@@ -0,0 +1,85 @@
...
+        execute();
+    }
+
+    public void execute() {
+        if( ! executing ) {
+            executing = true;
+            internalExecute();
+            executing = false;
+        } else {
+            System.out.println( "X already executing" );
```
mswiderski: logger?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/RunProcessInstance.java
```diff
@@ -0,0 +1,85 @@
...
+        }
+    }
+
+    private void internalExecute() {
+        while( ! processActionQueue.isEmpty() ) {
+            ProcessActionTrigger nextProcessActionTrigger = processActionQueue.pollLast();
+            System.out.println( "! triggering: " + nextProcessActionTrigger.toString() );
```
mswiderski: logger?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/RunProcessInstance.java
```diff
@@ -0,0 +1,85 @@
...
+        executing = true;
+        eventListener.signalEvent(type, event);
+        if( ! alreadyExecuting ) {
+            internalExecute();
+            executing = false;
+        }
+    }
+
+    private void addAction( ProcessActionTrigger action ) {
+        System.out.println("+ adding: " + action.toString());
```
mswiderski: and logger again?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/SignalEventAction.java
```diff
@@ -0,0 +1,41 @@
...
+        return String.valueOf(((NodeInstanceImpl) eventListener).getUniqueId());
+    }
+
+    @Override
+    public String toString() {
+        String uniqueId;
+        if( eventListener instanceof NodeInstanceImpl ) {
+            uniqueId = (String) ((NodeInstanceImpl) eventListener).getMetaData().get("UniqueId");
+        } else {
+            uniqueId = "dummy";
```
mswiderski: maybe better would be to give it a like a unique id - UUID.toString or so...

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -117,14 +120,24 @@ public void internalTrigger(final org.kie.api.runtime.process.NodeInstance from,
 		}
         CompositeNode.NodeAndType nodeAndType = getCompositeNode().internalGetLinkedIncomingNode(type);
         if (nodeAndType != null) {
-	        List<Connection> connections = nodeAndType.getNode().getIncomingConnections(nodeAndType.getType());
-	        for (Iterator<Connection> iterator = connections.iterator(); iterator.hasNext(); ) {
+	        Deque<Connection> connections = new ArrayDeque<Connection>(nodeAndType.getNode().getIncomingConnections(nodeAndType.getType()));
```
mswiderski: can all this logic be hidden in connections.iterator() that would be great.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -135,9 +148,12 @@ public void internalTrigger(final org.kie.api.runtime.process.NodeInstance from,
         		if (node instanceof StartNode) {
         			StartNode startNode = (StartNode) node;
         			if (startNode.getTriggers() == null || startNode.getTriggers().isEmpty()) {
-    	                NodeInstance nodeInstance = getNodeInstance(startNode);
-    	                ((org.jbpm.workflow.instance.NodeInstance) nodeInstance)
-    	                	.trigger(null, null);
+    	                NodeInstance nodeInstance = createNodeInstance(startNode);
+    	                if( isStackless() ) {
```
mswiderski: similar here, can we hide it in nodeInstance.trigger?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
mrietveld: What do you mean?

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
mrietveld: Do you mean the `if( debug) {` logic?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/ForEachNodeInstance.java
```diff
@@ -144,51 +146,65 @@ public void internalTrigger(org.kie.api.runtime.process.NodeInstance fromm, Stri
...
             		VariableScopeInstance variableScopeInstance = (VariableScopeInstance)
             			nodeInstance.resolveContextInstance(VariableScope.VARIABLE_SCOPE, variableName);
             		variableScopeInstance.setVariable(variableName, o);
             		nodeInstances.add(nodeInstance);
             	}
-            	for (NodeInstance nodeInstance: nodeInstances) {
-            	    logger.debug( "Triggering [{}] in multi-instance loop.", ((NodeInstanceImpl) nodeInstance).getNodeId() );
-            		((org.jbpm.workflow.instance.NodeInstance) nodeInstance).trigger(this, getForEachSplitNode().getTo().getToType());
+            	Iterator<NodeInstance> iter;
+            	if( isStackless()) {
```
mswiderski: same iterator hiding as mentioned before ...

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
mrietveld: MARCO backwards: it's the tag I used to mark places that I need to keep track of and that are important.

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
mrietveld: Yes, we do.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -117,30 +117,35 @@ public void cancel() {
         	.deactivateRuleFlowGroup( getRuleFlowGroup() );
     }
 
+    @Override
 	public void signalEvent(String type, Object event) {
 		if (getRuleSetEventType().equals(type)) {
             removeEventListeners();
             retractFacts();
-            triggerCompleted();
+            if( isStackless() ) {
```
mswiderski: hide in triggerCompleted if possible?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
mrietveld: Good idea, I'll see if I can do that.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
mrietveld: Good idea, will do.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StartNodeInstance.java
```diff
@@ -53,21 +54,25 @@ public void signalEvent(String type, Object event) {
...
             EventTransformer transformer = getStartNode().getEventTransformer();
     		if (transformer != null) {
     			event = transformer.transformEvent(event);
     		}
-            
+
             variableScopeInstance.setVariable(variableName, event);
         }
-        triggerCompleted();
+        if( isStackless() ) {
```
mswiderski: move to triggerCompleted if possible to hide the decision logic

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
mrietveld: I'm not sure, although I understand what you mean. I'll look at trying to make it more elegant. 

However, what you're suggesting would make the code less readable and would decrease performance for the old execution model, which is why I'm reluctant to do it.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
mrietveld: Because in short, execution order in stackless (because it's stackless, and not recursive) is the "reverse" of what happens with the old/recursive execution model. It's the same logic that happens with the node instance trigger (actions). 

BTW, I've since made this logic a lot more pretty.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/RunProcessInstance.java
mrietveld: I'll add a logger instance, although I don't expect it to be used.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/RunProcessInstance.java
mrietveld: @mswiderski Did you think this was final code?!?!?! ;D

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/SignalEventAction.java
mrietveld: ..which would break the link to the node instance itself.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/queue/RunProcessInstance.java
mswiderski: wasn't that the point of PR being sent? :)

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
mswiderski: fair enough, if that has impact on performance then does not make sense to change it

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
mswiderski: we want or we need to support it?

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
mswiderski: hahaha nice one :)

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
mswiderski: yes, as it does bit different thing than the else part

mswiderski:
overall would be nice to get an idea what is actually changing and how would the stackless work - I mean mainly the difference. 
Does stackless mean that it will invoke each node directly when it entered or will it sort of record it and then play through it so to say... if you have such general description that would be great.

mrietveld:
I'll submit a cleaned-up PR with a better description.

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
mrietveld: When the PR/changes are finished, the `runProcessInstance.execute()` will be triggered by the `DefaultAgenda`(/`PropagationAction`) logic. However, at the moment, that just complicates debugging, so I'm running it as shown (in the same thread).

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceFactory.java
mrietveld: Argh -- sorry, I didn't read your comment correctly. 

No, in 8 we don't want to support recursive anymore, as far as I know. It won't make sense since lots of the new functionality won't even be available via the old execution model.

PR #394, opened 2016-02-08
[JBPM-4927] Unit tests

DuncanDoyle:
Unit-test for JBPM-4927. The test passes within a unit-test (H2), but it fails on a PostgreSQL DB in my BPMSuite runtime. In that test, the processes that are started via a signal appear twice in the ProcessInstanceLog.


mswiderski:
@DuncanDoyle closing this as I opened new PR with fix and this commit https://github.com/droolsjbpm/jbpm/pull/395, thanks a lot

PR #396, opened 2016-02-11
JBPM-4934 AuditService obtained from RuntimeEngine doesn't work with Local transacations

agiertli:


mrietveld:
@mswiderski Please hold off on merging this: 

- On the face of this, this is a straightforward addition of functionality that users will enjoy. 
- However, our policy with regards to local transactions (last time I checked?) is that we simply do not do them. I've talked about this with @mdproctor, and the consistent message that I got back was that local transactions are not something we do. 

(In the past, I've worked on changes to make `drools-persistence-jpa` and `jbpm-persistence-jpa` local transaction compatible.) 

Before we add this, I want to check with @krisv that we 1. are fine with allowing this but also 2. that we have a policy or a vision in place for this. 

1. Do we want to allow local transactions for use with non-engine related retrieval, but not core engine activities? 
2. Or do we really want to say: no, local tx's are not something we do -- the rest of our infrastructure is JTA based, it doesn't make sense to let local tx's in, even on the edges. 

Does that make sense?

mrietveld:
@agiertli Could you add some tests as well? All PR's should have tests, and if related, should also have proper documentation (e.g. a PR to [kie-docs](https://github.com/droolsjbpm/kie-docs/))

agiertli:
I understand if the proper way is to use JTA transactions, but if user is using Local transactions, it is not possible to use AuditService directly from RuntimeEngine - user must use  **org.jbpm.process.audit.AuditLogService** instead - which is internal API.

This fix allows to use **org.kie.api.runtime.manager.audit.AuditService** (public api!) obtained directly from the RuntimeEngine with Local transactions configured.

mswiderski:
@mrietveld we only support local transactions as part of spring and this is only for spring use case as pointed out by @agiertli regarding tests.

So I agree we don't support local transaction in standalone mode but with spring we do as there are many users that rely on this. So for the sake of having consistent api (use of AuditService from RuntimeEngine) I vote for merging this.

Salaboy:
+1 damn spring :)

jbpm-audit/src/main/java/org/jbpm/process/audit/JPAService.java
```diff
@@ -33,108 +34,134 @@
...
+	public <T> List<T> executeQuery(Query query, EntityManager em, Class<T> type) {
+		Object newTx = joinTransaction(em);
+		List<T> result;
+		try {
+			result = query.getResultList();
+		} finally {
+			closeEntityManager(em, newTx);
+		}
+		return result;
+	}
```
psiroky: Nit picking, but would it be possible to _not_ use tabs for indentation? We generally have 4 spaces as indent.

agiertli:
I confirm that these tests are failing even **without** my PR, which seems they are unrelated these changes. Also, none of them is utilizing AuditService in any way..

psiroky:
Yes, the test failures are not related to these changes.

mswiderski:
jenkins retest this

mswiderski:
@mrietveld does the explanation makes sense so we can merge it or would you prefer to not have it here (jbpm repo)?

mswiderski:
@krisv @mrietveld any objections? if not heard from you till end of today I'll merge it...

mrietveld:
@mswiderski Apologies for the late answer (I was on PTO monday and tuesday): the explanation makes sense.

PR #403, opened 2016-02-16
BZ-1304508 - NPE exception is thrown when using jbpm kie ProcessService to issue signal event, JBPM-4929 - Inconsistent state in ProcessServiceEJBImpl.startProcess()

mswiderski:
two commits in same PR as they do modify same parts of code base so it's easier to see whole change


mswiderski:
jenkins retest this

PR #404, opened 2016-02-17
BZ-1304508 - NPE exception is thrown when using jbpm kie ProcessService to issue signal event, JBPM-4929 - Inconsistent state in ProcessServiceEJBImpl.startProcess()

mswiderski:
back port of #403 to 6.4.x branch


mswiderski:
merged into 6.4.x

PR #406, opened 2016-02-23
BZ-1310820 - Can't signal sub process when using the EJB API

mswiderski:
(cherry picked from commit 6b3cb39b55a3cefeeac8120aa472356aee396259)

back port of #405


mswiderski:
merged into 6.4.x

PR #408, opened 2016-02-24
Update install.html jBPM User Guide URL

ceefour:


mswiderski:
ok to test

mswiderski:
+1 once the build completes (don't expect any failures based on the change :)) will merge it

thanks @ceefour !

PR #412, opened 2016-02-25
[6.3.x] test fixes

psiroky:


mswiderski:
I think we need this commit to be backported too: https://github.com/droolsjbpm/jbpm/commit/e462da6dcb580ce604a0d2d85037f47d7c24e99e

psiroky:
@mswiderski yeah, looks like that. Any idea why the tests are not failing currently (without the cherry-pick)?

mswiderski:
because ejb tests are not executed as part of the build since they require EAP instance to use

psiroky:
Ah, I see. Looks like Cargo could help in here. I think we should enable those tests by default?

I will cherry-pick the additional commit as well.

psiroky:
@mswiderski is it safe to take the whole change? (I see there are changes than just the deploy->install).

mswiderski:
it should be safe, though since it was not requested I don't think we should put that change in 6.3.x. Only the change in the tests as they might be used for ejb testing by QE...

psiroky:
I've included the EJB test fixes as part of the first commit.

mswiderski:
excellent, once the build completes I'll merge it, thanks a lot @psiroky

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/deploy/DeploymentDescriptorManagerTest.java
```diff
@@ -18,6 +18,10 @@
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
 
+import java.io.ByteArrayInputStream;
+import java.lang.reflect.Field;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
```
psiroky: These imports are result of bad merge, will remove them.

mswiderski:
@psiroky ejb tests are arquillian tests so not sure if we can use cargo with it. But it should be enough to have EAP on jenkins and point it via system property - at least of local ejb tests https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-impl/src/test/resources/arquillian.xml#L8

for ejb client test cargo could be of great use as we need to deploy application to remote server and then connect to it with ejb client

psiroky:
Ah, Arquillian :) I think we should be able to enable those in CI (even if stay disabled by default). If it is possible to enable these by just setting the property, that would be great. I've created https://issues.jboss.org/browse/JBPM-4968 to make sure I don't forget.

mswiderski:
merged into 6.3.x

PR #420, opened 2016-03-11
DROOLS-1089 - flush reentrant working memory actions when eager evalu…

mswiderski:
…ation is enforced - test case


mswiderski:
jenkins retest this

PR #424, opened 2016-03-16
BZ-1310439 - RuntimeManagerFactory may fail to initialize in OSGi

mswiderski:
depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/133

### DIFF
--- a/jbpm-flow-builder/src/main/java/org/jbpm/osgi/flow/compiler/Activator.java
+++ b/jbpm-flow-builder/src/main/java/org/jbpm/osgi/flow/compiler/Activator.java
@@ -19,6 +19,7 @@ import org.drools.compiler.compiler.ProcessBuilderFactory;
 import org.drools.compiler.compiler.ProcessBuilderFactoryService;
 import org.jbpm.process.builder.ProcessBuilderFactoryServiceImpl;
 import org.kie.api.Service;
+import org.kie.api.runtime.manager.RuntimeManagerFactory;
 import org.osgi.framework.BundleActivator;
 import org.osgi.framework.BundleContext;
 import org.osgi.framework.ServiceRegistration;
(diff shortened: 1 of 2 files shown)


jbpm-flow-builder/src/main/java/org/jbpm/osgi/flow/compiler/Activator.java
```diff
@@ -35,7 +36,8 @@ public void start(BundleContext bc) throws Exception {
         this.processBuilderReg = bc.registerService( new String[]{ ProcessBuilderFactoryService.class.getName(), Service.class.getName()},
                                                                    new ProcessBuilderFactoryServiceImpl(),
                                                                    new Hashtable() );
-        ProcessBuilderFactory.reInitializeProvider();
+        RuntimeManagerFactory.Factory.reset();
+        ProcessBuilderFactory.reInitializeProvider();
```
psiroky: Could we use the same method name, co be consistent? The `reInitializeProvider()` may not be ideal, but we could change that as well potentially.

mswiderski: not really as they do different things. RuntimeManagerFactory ... reset() does only reset of the flag and exception without reinitializing the factory impl itself

psiroky: Ah, thanks. I see now. However, that brings the next question -- should we unify the approach? It seems that the the end result should be the same (a way to re-try to load the provider class)

mswiderski: we could and actually that was my first attempt though when jbpm-flow or jbpm-flow-builder is started there might still not be all stuff available - in this case jbpm-runtime-manager so it just makes sure it resets and allows "clean start" for clients calls. While the other makes sense to eagerly init as it's possible.

psiroky: Just thinking out loud -- would it be better to add the call to `jbpm-runtime-manager`'s Activator then?

mswiderski: there is no such Activator class in jbpm-runtime-manager...

psiroky: Right, so I think we should create one. That would make it consist with the other jBPM modules. I can take a look at that.

PR #436, opened 2016-03-30
[6.4.x] BZ-1322398 - Performance regression in starting a process which conta…

mswiderski:
…ins a human task


mswiderski:
merged into 6.4.x

PR #446, opened 2016-04-11
JBPM-5056 - Timer boundary events do not support cron expression given as variable, JBPM-5057 - Terminate end node in adhoc subprocess

mswiderski:


mrietveld:
+1

PR #449, opened 2016-04-13
replaced MavenRepository.deployArtifact with MavenRepository.installA…

mswiderski:
…rtifact

@mariofusco could you double check if what I did does not break anything from kie ci point of view


mariofusco:
+1

PR #450, opened 2016-04-15
JBPM-5036 - Ability to specify priority for Asynchronous Tasks

mswiderski:


jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/InMemorySimpleExecutorTest.java
```diff
@@ -72,5 +72,17 @@ public void testCustomIncrementingRequestRetry() throws InterruptedException {
     public void testCustomIncrementingRequestRetrySpecialValues() throws InterruptedException {        
         super.testCustomIncrementingRequestRetrySpecialValues();
     }
+
+    @Test
+    @Ignore("It's only for JPA based as it in memory does not care about priorities")
+    public void testPrioritizedJobsExecution() throws InterruptedException {
+        super.testPrioritizedJobsExecution();
```
mrietveld: I probably am missing something, but it seems odd to ignore a test (in order to override the base test class?) and then _also_ have the test execute the overriden (parent) method.

mrietveld:
+1 (To be honest, I only skimmed most of it, but I saw nothing that caught my eye. )

PR #451, opened 2016-04-18
Fix random user generation in case-mgmt test coverage

MarianMacik:
Hello,

when looking through the case-mgmt test coverage I have found a small issue. Since nextInt(n) picks number from 0 (inclusive) to n (exclusive), -1 was not supposed to be there (the very last user would not be picked).

Marian


mrietveld:
+1 once tests succeed.

MarianMacik:
This issue exists in 6.3.x and 6.4.x too. Should I create two additional PRs for it?

mrietveld:
Yes, please. 

Thanks!

PR #456, opened 2016-04-21
[DROOLS-1129] consistent OSGi Bundle-SymbolicName for all jars

psiroky:
Depends on:
https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/183

### DIFF
--- a/jbpm-services/jbpm-kie-services/pom.xml
+++ b/jbpm-services/jbpm-kie-services/pom.xml
@@ -11,6 +11,10 @@
   <packaging>bundle</packaging>
   <name>jBPM :: KIE Services</name>
 
+  <properties>
+    <osgi.Bundle-SymbolicName>org.jbpm.kie.services.api;singleton:=true</osgi.Bundle-SymbolicName>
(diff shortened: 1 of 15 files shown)


jbpm-services/jbpm-kie-services/pom.xml
```diff
@@ -11,6 +11,10 @@
   <packaging>bundle</packaging>
   <name>jBPM :: KIE Services</name>
 
+  <properties>
+    <osgi.Bundle-SymbolicName>org.jbpm.kie.services.api.jpa;singleton:=true</osgi.Bundle-SymbolicName>
```
mswiderski: jpa should not be in the bundle name - it was a typo previously. So I think it would make sense to fix it no. I don't think services api is used anywhere in osgi

mswiderski:
+1 to merge

would be good to fix the wrong name in jbpm-kie-services but not a must

psiroky:
The test failure is not related (flaky timer test).

PR #458, opened 2016-04-22
[6.4.x] BZ-1310820 - Can't signal sub process when using the EJB API

mswiderski:
(cherry picked from commit 801df91557418ad14889edbec7ebefc907ab203f)


mswiderski:
needs to wait for 6.4.x bz with blockers/patch approved flag, will update it once it has it

PR #461, opened 2016-04-27
JBPM-5084: troubleshooting function for workitem: unit test

bxf12315:
@mswiderski 
I create unit test for  PR :https://github.com/droolsjbpm/drools/pull/744

But, I find i just can test DefaultWorkItemManager,can not test JPAWorkItemManager.
and How to test InfinispanWorkItemManager?

### DIFF
--- a/jbpm-test/src/test/java/org/jbpm/test/RetryWorkItemTest.java
+++ b/jbpm-test/src/test/java/org/jbpm/test/RetryWorkItemTest.java
@@ -1,0 +1,76 @@
+package org.jbpm.test;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.drools.core.process.instance.WorkItemManager;
+import org.drools.core.process.instance.impl.DefaultWorkItemManager;
+import org.drools.persistence.jpa.processinstance.JPAWorkItemManager;
+import org.jbpm.test.workitem.ExceptionWorkItemHandler;
+import org.jbpm.test.workitem.NoExceptionWorkItemHandler;
+import org.jbpm.workflow.instance.WorkflowRuntimeException;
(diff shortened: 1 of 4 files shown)


bxf12315:
@mswiderski 
How to test JPAWorkItemmanager?
I find the line never executed.https://github.com/droolsjbpm/jbpm/pull/461/files#diff-c8a0ebe8279172ed72b6bf171d965c9fR73

mswiderski:
could you please move the tests to jbpm-test-coverage instead of jbpm-test as jbpm-test is only test utilities (or support test classes) and should not contain actual tests of functionality etc of jbpm itself.

your test is running currently an in memory engine only - this is driven by the constructor of JbpmJUnitBaseTestCase where you can define if the data source and persistence should be enable for not. Since you don't use any constructor explicitly it will use the defaults.
You can use @Parametrized JUnit support to make it flexible and included only in one class

bxf12315:
@mswiderski 
I use  setupDataSource and sessionPersisence model.
https://gist.github.com/bxf12315/a8de14c2a82151f33c971bdeca237e42#file-gistfile1-txt-L28
But i find tow question:
1 code line :https://gist.github.com/bxf12315/a8de14c2a82151f33c971bdeca237e42#file-gistfile1-txt-L49
  if process throw exception , processInstance is null.
2 code line :https://gist.github.com/bxf12315/a8de14c2a82151f33c971bdeca237e42#file-gistfile1-txt-L57
  if use default model , here can get processInstance , but use  setupDataSource sessionPersisence model. also can not get processInstance.

How get a processInstance in the model?

mswiderski:
can this be closed as new PR was opened https://github.com/droolsjbpm/jbpm/pull/466?

bxf12315:
ok

PR #462, opened 2016-04-27
JBPM-5082: Updating global realm configuration to make the users and groups management work by default on the jbpm installer.

romartin:


mswiderski:
looks good to me, so +1 to merge

@krisv any comments from your side as you manage installer?

mswiderski:
is this still needed for 6.4.x?

I guess we don't plan any community updates and installer is not used in product...

romartin:
@mswiderski Thanks! mmm not sure right now, should be nice having on the community but if no updates are comming... @krisv what you dot think?

mswiderski:
closing this one for now, if we find a need for it to be merged we can always reopen.

PR #463, opened 2016-04-27
JBPM-5082: Updating global realm configuration to make the users and groups management work by default on the jbpm installer.

romartin:


mswiderski:
same here - looks good to me, so +1 to merge

@krisv any comments from your side as you manage installer?

romartin:
@mswiderski Thanks!

PR #468, opened 2016-05-13
JBPM-5116 Persistence tests doesn't use transactions for handling LOBs in Postgresql Plus

baldimir:
- Added com.edb driver handling to PersistenceUtil.useTransactions() method.

Master PR: https://github.com/droolsjbpm/jbpm/pull/467

This is needed for testing next planned 6.4.x builds on Postgresql Plus.


mswiderski:
+1 affects only tests

PR #471, opened 2016-05-22
Fix javadoc property name

jludvice:
I believe this property is called Body, not text.

email


mswiderski:
thanks @jludvice merged to master

PR #472, opened 2016-05-25
JBPM-5125 - Parent process is not notified about the subprocess compl…

mswiderski:
…etion if the subprocess was started in a previous version


mswiderski:
jenkins retest this

PR #474, opened 2016-05-25
JBPM-5111 - minor enhancements for jbpm document to allow pluggable D…

mswiderski:
…ocumentStorageService and smooth integration with kie server

### DIFF
--- a/jbpm-document/src/main/java/org/jbpm/document/marshalling/DocumentMarshallingStrategy.java
+++ b/jbpm-document/src/main/java/org/jbpm/document/marshalling/DocumentMarshallingStrategy.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Copyright (C) 2012 Red Hat, Inc. and/or its affiliates
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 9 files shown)


jbpm-document/src/main/java/org/jbpm/document/marshalling/DocumentMarshallingStrategy.java
```diff
@@ -13,29 +13,32 @@
...
+import org.jbpm.document.service.DocumentStorageServiceProvider;
 
 public class DocumentMarshallingStrategy extends AbstractDocumentMarshallingStrategy {
 
     private DocumentStorageService documentStorageService;
 
     public DocumentMarshallingStrategy() {
-        documentStorageService = new DocumentStorageServiceImpl();
+        documentStorageService = DocumentStorageServiceProvider.get().getStorageService();
     }
```
mrietveld: +1 for factory methods!

jbpm-document/src/main/java/org/jbpm/document/service/DocumentStorageService.java
```diff
@@ -64,4 +65,12 @@
...
      */
     boolean deleteDocument(Document document);
+    
+    /**
+     * Lists available document with paging support.
+     * @param page page to be displayed
+     * @param pageSize number of elements to return
+     * @return
+     */
+    List<Document> listDocuments(Integer page, Integer pageSize);
```
mrietveld: I'm guessing you're using an Integer because of CDI or serialization considerations?

mrietveld:
+1 Nice work.

jbpm-document/src/main/java/org/jbpm/document/service/DocumentStorageService.java
mswiderski: yes, as this is generic interface it can be implemented by various frameworks etc so I wanted to stay on the safe side so to say

PR #477, opened 2016-06-06
JBPM-5147: Provide a JPA Work Item Handler

jesuino:
JBPM-5147: Provide a JPA Work Item Handler

### DIFF
--- a/jbpm-workitems/pom.xml
+++ b/jbpm-workitems/pom.xml
@@ -219,6 +219,43 @@
       <artifactId>jboss-jms-api_1.1_spec</artifactId>
       <scope>provided</scope>
     </dependency>
+    
+    <!-- JPA -->
+    <dependency>
+      <groupId>org.hibernate.javax.persistence</groupId>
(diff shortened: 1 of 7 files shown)


jesuino:
Hope this is correct this time, @mswiderski  :)

mswiderski:
@jesuino looks nice and clean now, thanks. 

One thing though, looks like the handler is not thread safe as it keeps entity manager open across invocations of the executeWorkItem method. Entity manager is not thread safe thus it should not be shared. Could you change it so entity manager is created and then closed (but not cleared) within executeWorkItem method?

psiroky:
Sorry for nit-picking, but would it be possible to reformat the code to use spaces instead of tabs ?(to be consistent with the rest of the codebase)

jesuino:
Hello,

Thanks for both comments.

I am modifying the tests. When I close the em, it detach the entities, which makes most part of the current test obsolete. 

Sorry @psiroky  - Unfortunately my IDE configuration did not correctly replace the TAB for spaces. I will update all the Java source files.

PR #480, opened 2016-06-09
JBPM-5071 - Migrate InitiatorProviderProcessListener from workbench i…

cristianonicolai:
…nto jbpm-kie-services

Refactored/shared common code between InitiatorProviderProcessListener and IdentityProviderAwareProcessListener

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/IdentityProviderAwareProcessListener.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/IdentityProviderAwareProcessListener.java
@@ -15,69 +15,40 @@
 
 package org.jbpm.kie.services.impl;
 
-import org.jbpm.process.instance.impl.ProcessInstanceImpl;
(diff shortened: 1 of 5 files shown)


mswiderski:
@cristianonicolai can we simply merge these two listeners into one as they do pretty much the same - set user name of logged in user as process variable/meta data.

Would be good to have a test within process context as well, if possible

cristianonicolai:
@mswiderski merged both listeners into a single one. The side affect of that is IdentityProviderAwareProcessListener is registered by default into the ksession. I guess we're ok with that?

Another point is that after this change, some tests started to fail because `getProcessInstancesByVariableNameAndValue` wasn't taking into account the specific variable when checking the max id. That's why the change on `Servicesorm.xml`

mswiderski:
+1 though can't see whats the reason of build failure so let's retest

mswiderski:
Jenkins retest this

mswiderski:
looks like test failures are related to this pr, @cristianonicolai do you mind taking a look?

mswiderski:
Thanks @cristianonicolai

PR #481, opened 2016-06-12
JBPM-5147 -  Provide a JPA Work Item Handler

jesuino:
JBPM-5147 -  Provide a JPA Work Item Handler


mswiderski:
There is still sharing of entity manager as it's class field. Make sure you create em in the execute method and then pass that instance to do* methods instead of using class field. 

I'd recommend to do operations in try catch and close em in finały to avoid any leaks.

mswiderski:
jenkins retest this

mswiderski:
now this looks nice, thanks @jesuino 

+1

PR #482, opened 2016-06-14
Remove no longer needed -XX:MaxPermSize configs

psiroky:
* deprecated in JDK 8 (e.g. has no effect) and
   removed in JDK 9 (build fails)


mswiderski:
+1

PR #483, opened 2016-06-15
JBPM-5157 - Refactoring of LDAPUserGroupCallbackImpl and LDAPUserInfoImpl

livthomas:
Both LDAPUserGroupCallbackImpl and LDAPUserInfoImpl classes contained very long methods with many code duplicates which made it really hard to understand how exactly they worked.

I have refactored them and extracted the logic dealing directly with LDAP to LdapSearcher class as well as their common setup to AbstractLDAPUserGroupInfo. I have not changed the behavior of LDAPUserGroupCallbackImpl or LDAPUserInfoImpl. Both of them are covered by a lot of tests which as you can see have not been changed at all.

The only thing that has been changed is `ldap.roles.attr.id` configuration property in LDAPUserGroupCallbackImpl which is now `ldap.role.attr.id`. I think there was a typo in the previous version and since we are working on the next major version we do not have to care about backward compatibility too much.

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/identity/LDAPUserGroupCallbackImpl.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/identity/LDAPUserGroupCallbackImpl.java
@@ -17,54 +17,41 @@
 package org.jbpm.services.task.identity;
 
 
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Properties;
-
-import javax.naming.Context;
-import javax.naming.NamingEnumeration;
-import javax.naming.NamingException;
-import javax.naming.directory.Attribute;
-import javax.naming.directory.SearchControls;
 import javax.naming.directory.SearchResult;
-import javax.naming.ldap.InitialLdapContext;
 
 import org.kie.api.task.UserGroupCallback;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 /**
  * LDAP integration for Task Service to collect user and role/group information.
- * 
+ * <p>
  * Following is a list of all supported properties:
  * <ul>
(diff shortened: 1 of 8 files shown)


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/identity/LDAPUserGroupCallbackImpl.java
```diff
@@ -17,341 +17,116 @@
...
     public static final String ROLE_CTX = "ldap.role.ctx";
     public static final String USER_ROLES_CTX = "ldap.user.roles.ctx";
     public static final String USER_FILTER = "ldap.user.filter";
     public static final String ROLE_FILTER = "ldap.role.filter";
     public static final String USER_ROLES_FILTER = "ldap.user.roles.filter";
     public static final String USER_ATTR_ID = "ldap.user.attr.id";
-    public static final String ROLE_ATTR_ID = "ldap.roles.attr.id";
-    public static final String IS_USER_ID_DN = "ldap.user.id.dn";
+    public static final String ROLE_ATTR_ID = "ldap.role.attr.id";
+    public static final String USER_ID_DN = "ldap.user.id.dn";
```
triceo: My 2 cents: this change is not worth the backwards incompatibility it creates.

Anyway, if we want to keep the change, it needs to be clearly marked in the docs as requiring migration.

mswiderski: Agree that it is not worth to break the backward compatibility by changing the property name as we don't gain much from it anyway

livthomas: Thanks for the feedback. I have reverted this change in order not to break the backward compatibility.

PR #484, opened 2016-06-16
query-jpa: explicitly depend on jboss-jaxb-api_2.2_spec

psiroky:
* the API is exposed by Java 8 by default, but not by
   Java 9. Marking the dependency as provided to not
   change the current behavior


mswiderski:
+1 to merge

PR #485, opened 2016-06-17
Jbpm 5158

selrahal:
https://issues.jboss.org/browse/JBPM-5158


mswiderski:
Ok to test

mswiderski:
@selrahal not sure I fully understand the issue here but this fix seems not consistent with what ksession does as it does not seem to call dispose unless a dispose command is issued.

would it be possible to test the application with -Dorg.jbpm.rm.engine.eager=true to ensure that ksession is always initialized before taskService?

if you find some time please ping me on irc so we could quickly discuss your findings

livthomas:
@selrahal Can you please provide a reproducer in the form of a unit test which fails before and passes after your fix is applied? It would clearly show where the problem was.

mswiderski:
Closing this as it was fixed in https://github.com/droolsjbpm/droolsjbpm-integration/pull/503

PR #490, opened 2016-06-25
ENTESB-4967 Different behaviour of DocumentBuilderFactory in Fuse and EAP

cunningt:
https://issues.jboss.org/browse/ENTESB-4697


mswiderski:
Ok to test

PR #491, opened 2016-07-05
JBPM-5071 - support for query service to allow filtering by user's as…

mswiderski:
…signed projects

@cristianonicolai @nmirasch could you guys take a look - as we discussed that yesterday


cristianonicolai:
@mswiderski, looks good +1

PR #495, opened 2016-07-05
JBPM-5071 - support for query service to allow filtering by user's as…

mswiderski:
…signed projects including tasks

@cristianonicolai added support for tasks to be filtered by deployment as well for completeness and correctness with current workbench behavior


cristianonicolai:
@mswiderski looks good +1

PR #500, opened 2016-07-11
JBPM-5196 - Sorting direction does not work properly for task variabl…

mswiderski:
…es queries


mrietveld:
+1 when tests complete.

mswiderski:
jenkins retest this

mswiderski:
tests failures are not related to the changes - most likely caused by very slow builds that make the test timeout before they complete

PR #501, opened 2016-07-13
JBPM-5204 Redundant DB calls when Pessimistic locking is used

agiertli:
no unit test created - the tests for this functionality are already there, and they should be still passing after this change.


mswiderski:
+1 to merge

thanks @agiertli

PR #502, opened 2016-07-15
JBPM-5212 - It is not possible to store persistable data object as us…

mswiderski:
…er task attachment

@mrietveld could you please take a look?

tests remain the same as they were as it should not change the behavior. The root cause was that the marshaling was done outside of command which should not be the case


mrietveld:
+1 (once the test build completes, of course).

PR #503, opened 2016-07-19
Be resilent to complex content type

dmarrazzo:
Sometime REST services return content type with extra information.
E.g. application/json;charset=utf-8

### DIFF
--- a/jbpm-workitems/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
+++ b/jbpm-workitems/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
@@ -336,11 +336,11 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler {
     
     protected String transformRequest(Object data, String contentType) {
         try {
-            if (contentType.equals("application/json")) {
+            if (contentType.contains("application/json")) {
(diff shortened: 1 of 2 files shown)


mswiderski:
that's correct and this was already fixed in transformResult why do we need that in transformRequest which essentially depends on what is given in parameters? Do you have a case that sets ContentType parameter based on some other service response?

mswiderski:
regardless, if you could add test case similar as it was done here https://github.com/droolsjbpm/jbpm/commit/e8ef9b91eb18b5b06cc95d680b88be2d6b79cfc6 I will be more than happy to merge (and jira so we can reference it as reported issue)

thanks a lot @dmarrazzo

dmarrazzo:
It has more sense for the result that it's not under the user control. Anyway, now it's more consistent.
Is it possible to backport even on the 6.3 branch in order to include with the next fix?

mswiderski:
if you could add test case to make sure this works as expected then we should be able to take this up for 6.5 at least and when requested by PM could be back ported to 6.4 (6.3 product)

dmarrazzo:
I just changed existing test cases, because the simpler case (all lowercase without charset def) is logically contained in the more complex one.

mswiderski:
ok to test

mswiderski:
+1 to merge

PR #505, opened 2016-07-20
JBPM-5186 - Added indexes to update scripts (to improve query performance on VariableInstanceLog and NodeInstanceLog)

mrietveld:


mswiderski:
@mrietveld looks like the test failures are related to the change, could you please have a look?

mrietveld:
@mswiderski I'm not sure what to do - the test is actually expected to fail. 

The problem is that I've added these indexes to the create/install DDL SQL scripts -- but it's already after the 6.3 and 6.4 releases. 

I thus also added them to the update DDL SQL scripts, for those people who installed 6.3 or 6.4 before this point. 

I'll comment out the indexes, that's probably the best solution.

mswiderski:
+1 to comment out, so they are there with a note that they might already exist, thanks

mrietveld:
jenkins please retest

mrietveld:
False negative ( org.jbpm.test.functional.jobexec.AsyncTaskCallbackTest.testTaskCallback )

Jenkins, please retest

PR #510, opened 2016-07-28
- removed unused (duplicate) ProcessDescriptor.resolveUnqualifiedClasses method

mrietveld:
See org.jbpm.kie.services.impl.xml.ServicesProcessDataEventListener.resolveUnqualifiedClasses()

[Github link to original method](https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/xml/ServicesProcessDataEventListener.java#L215)

which is called [here](https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/xml/ServicesProcessDataEventListener.java#L276)

@mswiderski Would you mind taking a look?


mswiderski:
+1 to merge

PR #513, opened 2016-07-28
BPMSPL-295 - added support for json in wid configs

tsurdilo:
- added support to use json in wid configs
- WorkItemRepository supports now parsing both mvel (default) as well as json
- added to tests

### DIFF
--- a/jbpm-flow-builder/pom.xml
+++ b/jbpm-flow-builder/pom.xml
@@ -70,6 +70,16 @@
       <artifactId>mvel2</artifactId>
     </dependency>
     <dependency>
+      <groupId>com.fasterxml.jackson.core</groupId>
+      <artifactId>jackson-databind</artifactId>
+      <optional>true</optional>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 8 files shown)


manstis:
What is the impact on the WID editor in drools-wb? That has templates etc for MVEL style definitions; and supports .wid file extensions.

tsurdilo:
@manstis there is no impact as the default config is still mvel. I will look into updating the editor for json configs as well

mswiderski:
one thing that I am bit concerned is to add dependency of jackson to jbpm-flow-builder. Would it be possible to have it at least optional and only load given class (that depends on jackson) then - like we have it now within catch block?

psiroky:
I believe we should also use the Jackson 2.x (it has groupId com.fasterxml.jackson). Jackson 1.x is basically dead.

tsurdilo:
@mswiderski @psiroky thanks!! will update.

tsurdilo:
@mswiderski amended ( also included jackson updates as per @psiroky )

mswiderski:
+1 to merge

PR #514, opened 2016-08-01
BPMSPL-295 - added support for json in wid configs - adjusted osgi co…

mswiderski:
…nfiguration

@psiroky @mrietveld could you please guys take a quick look?
this is to fix broken tests in karaf/osgi environment

@tsurdilo fyi in case this needs to be back ported anywhere.

### DIFF
--- a/jbpm-flow-builder/pom.xml
+++ b/jbpm-flow-builder/pom.xml
@@ -27,6 +27,11 @@
         <artifactId>maven-bundle-plugin</artifactId>
         <configuration>
           <instructions>
+            <Import-Package>
+                com.fasterxml.jackson.annotation;resolution:=optional,
+                com.fasterxml.jackson.databind;resolution:=optional,
+                *
+            </Import-Package>
             <Bundle-Activator>org.jbpm.osgi.flow.compiler.Activator</Bundle-Activator>
           </instructions>
         </configuration>


mrietveld:
+1 -- sort of. 

On one hand, sure, this is the correct fix for the OSGi problem. 

On the other hand, I'm only noticing now that the fasterxml dep's have been added to jbpm-flow-builder. (Apologies for not seeing this earlier)

Given that the `WorkItemRepository` and `JsonWorkItemParser` are not used anywhere else in the module, I'm wondering if it's not possible to move them to 'jbpm-workitems', or at least a non-core module? 

It doesn't seem like either class belongs in `jbpm-flow-builder`... 

@mswiderski What do you think?

mswiderski:
I see this as a consequence of dealing with work item definitions in this module. I don't see it as perfect place for it (rather would like to move it out to separate module which would then provide work item definition, repository etc support).  Although this is bigger work than just a day or two thus though it will make more sense to put it here. And mark the jackson dependencies as optional so unless you really need to use WID with json format then you have to bring in the jackson dependencies yourself. 

I plan to do some work around work items before 7 goes out (mainly to separate the jbpm-workitems module into several that scoped to what they should be and not a one huge bucket of work items) so that should certainly be part of that work. 
Does this make sense?

mswiderski:
jenkins retest this

mrietveld:
@mswiderski yes -- thanks. 

+1 with the "promise" that it will be eventually moved somewhere else.. :)

psiroky:
+1 (based on the discussion above)

mswiderski:
jenkins retest this

PR #516, opened 2016-08-01
RHBPMS-3973: A ClassNotFound exception is shown while a Service task …

lazarotti:
…is being executed


mswiderski:
replaced by https://github.com/droolsjbpm/jbpm/pull/520 with test case

PR #517, opened 2016-08-01
RHBPMS-3973: A ClassNotFound exception is shown while a Service task is being executed

lazarotti:


mswiderski:
replaced by https://github.com/droolsjbpm/jbpm/pull/519 with test case

PR #530, opened 2016-08-08
Indentation fixes; removing no longer needed files and configuration

psiroky:
* adding default bitronix config files to fix tlog file locations
   (we want them in target/ not in current working dir)


mswiderski:
+1 to merge

PR #531, opened 2016-08-09
JBPM-5240 - Provide easy way to access task variables via TaskEvent f…

mswiderski:
…or process instance

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
@@ -23,18 +23,20 @@ import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
 import java.io.StringReader;
-import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.jbpm.services.task.events.DefaultTaskEventListener;
 import org.jbpm.services.task.exception.PermissionDeniedException;
 import org.jbpm.services.task.impl.factories.TaskFactory;
 import org.jbpm.services.task.impl.model.xml.JaxbContent;
 import org.jbpm.services.task.utils.ContentMarshallerHelper;
 import org.junit.Test;
+import org.kie.api.task.TaskEvent;
+import org.kie.api.task.TaskLifeCycleEventListener;
 import org.kie.api.task.model.Comment;
 import org.kie.api.task.model.Content;
 import org.kie.api.task.model.Group;
@@ -44,6 +46,7 @@ import org.kie.api.task.model.Status;
 import org.kie.api.task.model.Task;
 import org.kie.api.task.model.TaskSummary;
 import org.kie.api.task.model.User;
+import org.kie.internal.task.api.EventService;
 import org.kie.internal.task.api.TaskModelProvider;
 import org.kie.internal.task.api.model.AccessType;
 import org.kie.internal.task.api.model.ContentData;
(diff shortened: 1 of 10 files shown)


jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/ExternalTaskEventListener.java
```diff
@@ -54,29 +51,15 @@ public void processTaskState(Task task) {
...
-                }
-                results.put("ActorId", userId);
-                session.getWorkItemManager().completeWorkItem(task.getTaskData().getWorkItemId(), results);
-            } else {
-            	results.put("ActorId", userId);
-                session.getWorkItemManager().completeWorkItem(workItemId, results);
+            Map<String, Object> taskOutcome = task.getTaskData().getTaskOutputVariables();
+            if (taskOutcome != null) {
+                results.putAll(taskOutcome);
+//                results.put("Result", taskOutcome);
```
mswiderski: this is to be removed once all test go green - meaning it is not used anywhere

mswiderski:
jenkins retest this

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
```diff
@@ -2388,7 +2391,151 @@ public void testCompleteByActiveTasks() {
...
+                assertNull(event.getTask().getTaskData().getTaskOutputVariables());
+                
+                event.getTaskContext().loadTaskVariables(event.getTask());
+                
+                assertNotNull(event.getTask().getTaskData().getTaskInputVariables());
+                assertEquals(1, event.getTask().getTaskData().getTaskInputVariables().size());
+                assertTrue(event.getTask().getTaskData().getTaskInputVariables().containsKey("input"));
+                
+                assertNull(event.getTask().getTaskData().getTaskOutputVariables());
+            }
```
psiroky: These listener methods seem to be very similar. Could we extract the code into new sub-methods and reuse those? (I know these are "just" tests, but I believe even the tests should be kept as clean as possible).

mswiderski: yes, I thought about it a bit and couldn't find much better way which would not make it much less readable thus I stayed with duplicated as we can easily read it 

if you have any suggestion how to improve without loosing too much of readability then I am all in

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/jaxb/AbstractTaskSerializationTest.java
```diff
@@ -168,15 +170,15 @@ public void jaxbTaskTest() throws Exception {
...
+        ComparePair compare = new ComparePair(task, bornAgainTask, Task.class);        
         compare.recursiveCompare();
 
         assertNotNull(((InternalTask) xmlTask).getFormName());
         assertEquals(((InternalTask) xmlTask).getFormName(), ((InternalTask) bornAgainTask).getFormName());
 
         Task realTask = xmlTask.getTask();
         verifyThatXmlFieldsAreFilled(realTask, xmlTask, InternalTask.class, "deadlines");
-        verifyThatXmlFieldsAreFilled(realTask.getTaskData(), xmlTask.getTaskData(), TaskData.class);
+        verifyThatXmlFieldsAreFilled(realTask.getTaskData(), xmlTask.getTaskData(), TaskData.class, "taskInputVariables", "taskOutputVariables");
```
mrietveld: Thanks! :)

mrietveld:
+0 -- In other words, fine to merge, but I don't have a good feeling about the solution. 

The reason I'm adding +0 instead of +1 is that this is basically a "band-aid" solution to extract variables from the content. In other words, I feel that the solution is not very straight-forward or elegant. 

However, I can understand that there would need to be much deeper fixes to human-task to make it an elegant solution (schema changes, probably) and that it's not worth doing that for this feature. So, +0. :)

mrietveld:
(Re: "band-aid" -- we are basically storing a BLOB in the database, because we don't know what type of object we're storing. IMHO, given the demand for storing variables (unmarshalling a `Map` from this BLOB has been an issue for a couple years now), we should be storing those variables elsewhere, not in this BLOB. BLOB's are usually meant for things likes images, files and other large binary objects.. ;) ).

mswiderski:
correct, this is just a "helper" like solution to let uses simply pass task object to load variables to it instead of going through all low level services of the task to load it yourself. 
Agreed that we do need to rework task schema and the way we store it and I do hope we'll find the time for it before v7 goes out.

PR #533, opened 2016-08-11
JBPM-5252 - Node instance counter is not properly set after marshaling

mswiderski:


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
```diff
@@ -2601,4 +2601,44 @@ public void testMultipleTimerBoundaryEventCronCycleVariable() throws Exception {
...
+            countDownListener4.waitTillCompleted();
+            logger.debug("Fourth timer triggered");
+            
+            List<WorkItem> wi = handler.getWorkItems();
+            assertNotNull(wi);
+            assertEquals(3, wi.size());
+    
+            ksession.abortProcessInstance(processInstance.getId());
+        } finally {
+            System.clearProperty("jbpm.enable.multi.con");
```
psiroky: Just a minor thing -- but I believe we should reset the property to the original value instead of clearing it? (in case it would have been set to `false`, we would loose that setting).

mswiderski: clearing this property and removing it completely has exact same effect - as engine cares only about it when it's set to true

psiroky: Oh, OK. But in that case I think we could add a comment saying that, to make it clear also for future readers.

mswiderski: comment added

mswiderski:
jenkins retest this

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -93,9 +93,12 @@ public NodeContainer getNodeContainer() {
 	}
 
 	public void addNodeInstance(final NodeInstance nodeInstance) {
-	    long id;
-	    id = singleNodeInstanceCounter.getAndIncrement();
-		((NodeInstanceImpl) nodeInstance).setId(id);
+	    if (nodeInstance.getId() == 0) {
+            // assign new id only if it does not exist as it might already be set by marshalling 
+            // it's important to keep same ids of node instances as they might be references e.g. exclusive group
+    	    long id = singleNodeInstanceCounter.getAndIncrement();
```
mrietveld: +100 on adding the comment! Thanks.

mrietveld:
+1

PR #534, opened 2016-08-11
JBPM-5234 and JBPM-5255

mswiderski:
JBPM-5234 - Signalling container signals processes from other containers
JBPM-5255 - 'Process Not Found' when deploying more than one DeploymentUnit in BPM Suite 6.3 using PER REQUEST strategy


mrietveld:
+1 Nice work.

mswiderski:
merged manually to preserve both commits

PR #536, opened 2016-08-12
Fix indentation (tabs -> spaces) + enable checkstyle

psiroky:
* from now on the build will fail in case the source
   files use tabs for indentation

@mswiderski do you agree with merging this? It will likely create merge conflicts for existing changes, so it could wait until some bigger work is done (not sure if you or someone else has some bigger feature branch).


mswiderski:
ouch, I have a rather big one for case mgmt .... so would prefer to avoid too many merge conflicts... if possible

psiroky:
@mswiderski the other way around is to apply the same changes I did (I have scripts for that) on your branch -- then those merge conflicts should disappear. We can also wait till the changes are in (and would then just rerun the scripts on the new sources).

mswiderski:
I'd opt for delaying until the branch is merged - should be ready for first merge within a week. That way we scope it only to actual changes and then we clean up the formatting. Does that sound ok?

psiroky:
Yes, sounds good. I will close the PR now and re-create it once we agree to do these changes.

PR #541, opened 2016-08-16
JBPM-5236 - EJB Migration service - transactions fixed

ibek:


mswiderski:
+1 to merge when build completes, test will follow after migrating jbpm-container test into 6.5.x, thanks @ibek

PR #544, opened 2016-08-18
GUVNOR-2173,JBPM-5269 - Adding the jbpm-bpmn2-services module

mrietveld:
@mswiderski -- what do you think of this? 

See the following commit for use of this in designer: 
- https://github.com/droolsjbpm/jbpm-designer/pull/375/files?w=1#diff-63469f70be2a8501e45bd8454a0045e1R67

PR set: 
- https://github.com/droolsjbpm/droolsjbpm-build-bootstrap/pull/257
- https://github.com/droolsjbpm/jbpm/pull/544


mrietveld:
Jenkins retest please

mrietveld:
jenkins retest please (especially now that I added the necessary PR from `droolsjbpm-build-bootstrap`

mswiderski:
thanks for the link in designer as it makes it much easier to understand the flow.

Now I start to wonder why we need the BPMN2DefinitionServiceImpl at all, mainly because as you stated in the comment in designer it might fail and during authoring phase it can quite frequently which then leads to fallback using parser. 

What would you say to directly use parser approach only with customer listener developed in designer? That way we do have it done in single place and there is no need to rely on BPMN2DefinitionServiceImpl at all.

mrietveld:
@mswiderski Sounds reasonable, I'll try that. 

However, that will then basically duplicate the code that's now in `jbpm-bpmn2-services`. Is that okay?

mswiderski:
I believe it should be much less then what we need in services - if I got it right change impact mainly concert the data types (variables, imports, globals, interfaces, etc) used in process and does not really care if there are 10 user tasks or what user groups are assigned to the tasks.

Anyway I am ok with bit of duplication as it's for a good purpose :)

mrietveld:
Okay, that makes sense. 

I'll close this and do the agreed stuff in designer. 

Thanks!

PR #556, opened 2016-08-23
JBPM-5283 - Remove actual owner from roles allowed to activate tasks

livthomas:
I have removed unnecessary permission from `operations-dsl.mvel`.

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/resources/operations-dsl.mvel
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/resources/operations-dsl.mvel
@@ -119,19 +119,18 @@
                     allowed = [ Allowed.Owner, Allowed.BusinessAdministrator  ],      
                 newStatus = Status.Failed
             } ],
-    
     Operation.Activate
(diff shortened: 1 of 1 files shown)


mrietveld:
Please hold off on merging this -- while what Tomas is saying makes sense with regards to the spec, I think that the code-level interpretation of "Actual owner" may not match that (and may in some cases be the initiator?). I'd like to double check that today.

livthomas:
@mrietveld, thanks for looking at this. I ran into it when I was verifying the documentation and after a short discussion with @mswiderski decided to remove it. But I have not checked any internal interpretation of **_Actual owner_** so there is a chance it is done this way on purpose and might cause some problems. If this is a wrong step, we should also see it on failing tests.

livthomas:
It seems that there are no related test failures. Failing `ConcurrentHumanTaskTest` is passing on my machine. It would be nice if Jenkins can retest this.

mrietveld:
@livthomas Thanks for the additional info. 

+1

mswiderski:
thanks guys @livthomas @mrietveld

PR #564, opened 2016-09-01
[6.5.x] Fix compilation problems caused by changes in KIE API

livthomas:
After changes in KIE API introduced by DROOLS-1267, it is not possible to compile jBPM as well as KIE Server (and maybe other projects). I have implemented `submit` method added to KIE API by throwing `UnsupportedOperationException`. Not sure if it is the best approach but it seems like it is only Drools-related operation.


mswiderski:
+1 to merge

PR #568, opened 2016-09-02
[6.5.x] RHBPMS-1396, JBPM-5305 - Subprocess with Timers throws error while ex…

mswiderski:
…ecution

@psiroky @krisv do you mind?

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
@@ -18,6 +18,9 @@ package org.jbpm.runtime.manager.impl;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.locks.ReentrantLock;
 
 import org.drools.core.command.CommandService;
 import org.drools.core.command.SingleSessionCommandService;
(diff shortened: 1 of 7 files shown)


psiroky:
@mswiderski I can take a look at this, but it will likely take me quite some time to understand the issue and the fix as well...

mswiderski:
no problem, the main review should be around the locking mechanism used in case it might lead to deadlock or so. 
Feel free to ask any questions to clarify things

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
```diff
@@ -527,4 +541,67 @@ public TaskService initTaskService(Context<?> context, InternalRuntimeManager ma
...
+    
+    protected void createLockOnGetEngine(Context<?> context, RuntimeEngine runtime) {
+        if (!useLocking) {
+            logger.debug("Locking on runtime manager disabled");
+            return;
+        }
+        
+        if (context instanceof ProcessInstanceIdContext) {
+            Long piId = ((ProcessInstanceIdContext) context).getContextId();
+            if (piId != null) {
```
psiroky: What is the case when the pidId can be null?

mswiderski: when you start new process instance, then the context is without process instance id as it was not yet created

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/MultiInstanceCallActivityRuntimeManagerTest.java
```diff
@@ -0,0 +1,157 @@
...
+    public void testPerProcessInstanceMultiInstanceCallactivityCompleteAtTheSameTime() {
+        RuntimeEnvironment environment = createEnvironment();        
+        manager = RuntimeManagerFactory.Factory.get().newPerProcessInstanceRuntimeManager(environment, "first");        
+        assertNotNull(manager);
+
+        testMultiInstanceCallactivityCompleteAtTheSameTime();
+    }
+
+
+    public void testMultiInstanceCallactivityCompleteAtTheSameTime() {
```
psiroky: I would make this `private` as now it slightly looks like this could a be a test method with a missing `@Test` annotation.

psiroky:
+1 looks good (based on my still limited knowledge of the RuntimeEngine/Manager)

PR #569, opened 2016-09-02
Use default timeout for jbpm-bpmn2 tests to avoid build hanging

psiroky:
@mswiderski this does not fix the test, but it at least prevents the build from hanging when the tests goes havoc

Timeout 30s should I think be enough even for slower machines. If we have tests that take more than that, we should probably think about fixing them.


mswiderski:
+1 to merge, thanks @psiroky

PR #572, opened 2016-09-05
removed unused StatefulProcessSession and its dependencies (#448)

mswiderski:
(cherry picked from commit a0d03862909490c6b69e3e01ba6cc3ce29d533ab)

required to cleanup the session interface impl that is no longer used
https://github.com/droolsjbpm/jbpm/commit/9052ba25fe610eb8089a772d83e56f0f96c2732b


livthomas:
+1 (assuming it will not break any tests)

But I need to ask: Where was this used before and why is it no longer needed?

mswiderski:
it's been added long time ago to allow alternative (without drools) ksession if I remember but I have never seen it to be used to be honest

PR #577, opened 2016-09-07
RHBPMS-4241 - Fix NPE when deploymentId is null

cristianonicolai:


manstis:
@cristianonicolai I don't know the jbpm code so excuse my ignorance; but *why* would a ```deploymentId``` be ```null```? Isn't this type of fix a solution for a symptom and not a cause?

cristianonicolai:
@manstis sure, I meant to break this down into two separated PRs, see https://github.com/droolsjbpm/jbpm/pull/580. The later deals with the actual cause as you mentioned, based on this JIRA. My concern with fixing the null containerId even if that isn't really expected is that is a late change in the behavior of this API which could have impact into other areas.

mswiderski:
I actually don't see much value with this as the deploymentId must be given when using these methods as that's the whole point in this. Skipping when there is null will most likely hide the problem as users would expect to find stuff but instead of being informed about the wrong arguments they will get (in most of the cases) empty result. So instead of null check I'd prefer to throw exception saying that deploymentId is mandatory.

cristianonicolai:
@mswiderski I agree that throwing an exception that the deploymentId is required is the best way to handle this. The only point for me is that the empty or null results are the expected behavior already for 6.4.x. and 6.5.x versions, changing it there could have other side effects than the one found in this JIRA as other service calls were impacted by a previous commit. So perhaps we could handle it differently on master, using the required exception.

mswiderski:
@cristianonicolai fair enough, so let's merge this in to avoid issues 
+1

PR #579, opened 2016-09-07
RHBPMS-4241 - Avoid call to process service when task isn't associate…

cristianonicolai:
…d with a process


mswiderski:
+1 to merge when build completes

PR #584, opened 2016-09-08
JBPM-5294: Automatically reassigned tasks remain in task list even af…

nmirasch:
…ter being reassigned.
Changed the Preprocessor adding conditions to differentiate between the tasks without actual owner but allowed to any user connected role, and the tasks with actualowner set to connected user.


mswiderski:
test failure not related but let's jenkins retest this

mswiderski:
+1

PR #585, opened 2016-09-08
JBPM-5082: Added users management configurations for the jbpm installer.

romartin:
Hi @krisv , @mswiderski , @dgutierr ,

Can you please check if you agree on this change? If so, will create the PR for master as well. 

Currently passwords on the jbpm-installer are on clear text, but both add-user script and our users management stuff do encrypt the passwords. So adding users from the UI adds the hashed password on the properties file, but the default passwords are on clear text.

I see two options:
1- Adding a configuration flag on the users management stuff to indicate wether passwords should be encrypted and use this configuration to set no encryption on the jbpm-installer.
2.- Moving the jbpm-installer to the use of encrypted passwords, as both add-user script and users management stuff expects.

This PR is the fix for option 2 -> Added users management configurations for the jbpm installer and disabled use of UserRoles login module for the other domain security (use of encrypted password using same RealmDirect module).

Note that the UserRoles login module used on the security domain is replaced by the RealmDirect one, which by default expects the encrypted passwords. This way:
- The passwords are encrypted ( added comments in the password file with the clear text )
- The configuration is centralized on just the realm conftiguration ( no need to specify the file paths on the security domain section as well )

I have checked the workbench and the kieserver, both connectivities works fine and now the users created from the UI can log in fine. Not sure if this change could have more implications...

PS: https://issues.jboss.org/browse/JBPM-5082

Thanks!!

### DIFF
--- a/jbpm-installer/auth/users.properties
+++ b/jbpm-installer/auth/users.properties
@@ -1,9 +1,18 @@
-admin=admin
-krisv=krisv
-john=john
-mary=mary
-sales-rep=sales-rep
-katy=katy
-jack=jack
-salaboy=salaboy
-kieserver=kieserver1!
+# admin/admin
+admin=207b6e0cc556d7084b5e2db7d822555c
+# krisv/krisv
+krisv=7b21a03b9918f9c629a46e119a9b8714
+# john/john
+john=afda4373c6021f3f5841cd6c0a027244
+# mary/mary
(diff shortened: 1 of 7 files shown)


krisv:
+1

romartin:
PS: If this gets metged, will update docs as it's necessary!

mswiderski:
jenkins retest this

PR #586, opened 2016-09-09
RHBPMS-4249, JBPM-5319 - Transaction exception for Per Process Instan…, RHBPMS-4247, JBPM-5317 - Failed to upgrade process instances in Maria

mswiderski:
@sutaakar do you mind taking a look and if you have a change to double check it?

this PR consists of both 
- migration sql fix - split into two sql (collect nodes to update and then update)
- weblogic transaction issue - somehow transaction manager in weblogic couldn't deal with multiple transactions within same incoming request though making it as single works properly. Due to the issue in finding the cause - very few information about the cause and not being able to debug it was better to make it as single tx, and it will make it slightly faster as well as only one tx is started.


sutaakar:
changes looks good
Tried 6.5.x with weblogic transaction issue locally and it looks fixed.
Will try migration issue too.....

sutaakar:
Migration issue looks fixed too, checked on MariaDB.

mswiderski:
@sutaakar thanks for quick verification and review

mswiderski:
merged manually to prevent separate commits

manstis:
@mswiderski I assume you merged manually to *keep* separate commits; not *prevent* them!

mswiderski:
@manstis of course, that's what happens when there are too many ideas on how to write something... I should stick to simple and not try to be smarter than...

PR #587, opened 2016-09-09
[6.5.x] RHBPMS-4249, JBPM-5319 - Transaction exception for Per Process Instan…, RHBPMS-4247, JBPM-5317 - Failed to upgrade process instances in Maria

mswiderski:
@sutaakar and here is back port to 6.5.x


mswiderski:
jenkins retest this

PR #590, opened 2016-09-11
RHBPMS-4252, JBPM-5320 - Signal not getting invoked externally throug…

mswiderski:
…h rest API when it is in Error Scenario

@krisv @psiroky do you mind taking a look?

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
@@ -69,6 +69,7 @@ import org.jbpm.workflow.core.node.EndNode;
 import org.jbpm.workflow.core.node.EventNode;
 import org.jbpm.workflow.core.node.EventSubProcessNode;
 import org.jbpm.workflow.core.node.EventTrigger;
+import org.jbpm.workflow.core.node.FaultNode;
 import org.jbpm.workflow.core.node.HumanTaskNode;
 import org.jbpm.workflow.core.node.RuleSetNode;
 import org.jbpm.workflow.core.node.Split;
@@ -749,6 +750,7 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
     }
 
     private void postProcessNodes(RuleFlowProcess process, NodeContainer container) {
+        List<String> eventSubProcessHandlers = new ArrayList<String>();
         for (Node node: container.getNodes()) {
             
             if (node instanceof StateNode) {
@@ -806,7 +808,9 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
                                             exceptionHandler.setAction(action);
                                             exceptionHandler.setFaultVariable(faultVariable);
                                             if (faultCode != null) {
-                                            	exceptionScope.setExceptionHandler(type.replaceFirst(replaceRegExp, ""), exceptionHandler);
+                                                String trimmedType = type.replaceFirst(replaceRegExp, "");
+                                            	exceptionScope.setExceptionHandler(trimmedType, exceptionHandler);
+                                            	eventSubProcessHandlers.add(trimmedType);
                                             } else {
                                             	exceptionScope.setExceptionHandler(faultCode, exceptionHandler);
                                             }
@@ -860,6 +864,16 @@ public class ProcessHandler extends BaseAbstractHandler implements Handler {
                 	throw new IllegalArgumentException("Event node '" + node.getName() + "' [" + node.getId() + "] has no incoming connection");
                 }
             }  
+        }
+        
+        // process fault node to disable termnate parent if there is event subprocess handler
+        for (Node node: container.getNodes()) {
+            if (node instanceof FaultNode) {
+                FaultNode faultNode = (FaultNode) node;
+                if (eventSubProcessHandlers.contains(faultNode.getFaultName())) {
+                    faultNode.setTerminateParent(false);
+                }
+            }
         }
     }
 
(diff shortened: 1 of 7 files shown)


jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
```diff
@@ -85,6 +86,8 @@
 import org.xml.sax.Attributes;
 import org.xml.sax.SAXException;
 
+import oracle.net.aso.f;
```
livthomas: Do we really need this import? Or has it been added accidentally?

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/JbpmBpmn2TestCase.java
```diff
@@ -144,7 +144,7 @@
     private RequirePersistence testReqPersistence;
 
     @Rule
-    public Timeout globalTimeout = Timeout.seconds(30);
+    public Timeout globalTimeout = Timeout.seconds(300);
```
livthomas: Is this really necessary? Cannot we override the rule with higher timeout in the test class where it is really needed? Waiting 5 minutes for every single test is quite a lot of time and if something goes wrong with more tests, it will take a while to finish all the tests in this module.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
mswiderski: That's by mistake, will remove it

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/JbpmBpmn2TestCase.java
mswiderski: That needs to be reverted as it was only increased when debugging

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
```diff
@@ -806,9 +808,12 @@ private void postProcessNodes(RuleFlowProcess process, NodeContainer container)
                                             exceptionHandler.setAction(action);
                                             exceptionHandler.setFaultVariable(faultVariable);
                                             if (faultCode != null) {
-                                            	exceptionScope.setExceptionHandler(type.replaceFirst(replaceRegExp, ""), exceptionHandler);
+                                                String trimmedType = type.replaceFirst(replaceRegExp, "");
+                                            	exceptionScope.setExceptionHandler(trimmedType, exceptionHandler);
+                                            	eventSubProcessHandlers.add(trimmedType);
                                             } else {
                                             	exceptionScope.setExceptionHandler(faultCode, exceptionHandler);
+                                            	eventSubProcessHandlers.add(faultCode);
```
livthomas: Why do you add `faultCode` which is `null` to the list?

krisv:
@mswiderski do we really need to add a listener to NodeInstanceContainer for this?  CompositeNode/Instance already supports passing events to the child nodes, so I was expecting a simple super.signalEvent(..) in EventSubProcessInstance (which you are calling now as well just through a listener), did you try that?

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
mswiderski: that's a valid comment, thanks @livthomas

mswiderski:
@krisv yes, I did try that and the issue is that it won't work as regular listener as it uses the dummy one and then it should go into checking if the node accepts the event while in case of event subprocess node instance is checking the start events only to instantiate. 
As you know there are two options:
- signal active node instances via concrete listeners https://github.com/droolsjbpm/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java#L490-L502
- signal event nodes when their definition is aware of the signal https://github.com/droolsjbpm/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java#L505

So with event subprocess we cannot rely on second approach so I made the NodeInstanceContainer to allow to register event listeners too.

krisv:
@mswiderski  shouldn't we then call super.acceptsEvent in EventSubProcess as well?  I'm afraid we'll otherwise miss other events (we're only fixing it for EventNodeInstance but other signals would still get lost)?

mswiderski:
@krisv we cannot add super.acceptEvent in EventSubProcessNode as that will essentially cause the event sub process to be created again. signalEvent in subprocess is mainly dedicated for starting instance of it and then the listener should be used trigger individual instances of the active event subprocess instance.
Though a good comment about other types of signals that we might lose, will revisit it and see if it can be improved in anyway.

mswiderski:
@krisv I changed the approach by implementing some additional logic to handle event subprocess node instead of registering event listen from node instance container, do you mind taking a look?

krisv:
@mswiderski this is kinda what I had in mind: this way the container will always accept events for itself or any of the nodes it contains, and delegate the signals to the sub node instances where appropriate.  The EventNodeSubProcess is just special in that it also creates a new instance of itself incase its event is triggered (but other than that still calls super.signalEvent or super.acceptsEvent):
https://github.com/krisv/jbpm/commit/9d51d89bd3dc7c7b6346cad46127723c14fa6382
Wdyt?

mswiderski:
@krisv it looks more compact so +1 for your version. I'll close this one so you can submit your PR, is that ok?

mswiderski:
@krisv closing this so please drop your version of the fix as PR

PR #591, opened 2016-09-12
RHBPMS-4259 - ServiceTaskHandler logs error messages for web service …

tkobayas:
…namespace/servicename attempts

I chose INFO rather than DEBUG because there could be a case that the log is important. I also added INFO log "WS Client is created for ..." so that users can be sure that the client is eventually created.


mswiderski:
+1 to merge

mswiderski:
@tkobayas if you'd like to include this in 6.4 product, please create PR for 6.5 as well, thanks!

tkobayas:
Sure, sent https://github.com/droolsjbpm/jbpm/pull/593

PR #592, opened 2016-09-12
Skip findbugs for jbpm-installer

psiroky:
* there are no java classes in src/main/java
   which could be analyzed and the plugin just
   fails becuase of that


cristianonicolai:
+1

PR #595, opened 2016-09-13
RHBPMS-4249, JBPM-5319 - Transaction exception for Per Process Instan…

mswiderski:
…ce deployment in WebLogic - fix for getting transaction manager

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
@@ -303,7 +303,7 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
     		return (TransactionManager) txm;
     	}
     	
-    	return TransactionManagerFactory.get().newTransactionManager();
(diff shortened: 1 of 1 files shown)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -303,7 +303,7 @@ protected TransactionManager getTransactionManager(Environment env) {
     		return (TransactionManager) txm;
     	}
     	
-    	return TransactionManagerFactory.get().newTransactionManager();
+    	return TransactionManagerFactory.get().newTransactionManager(env);
```
mariofusco: @mswiderski This fixes the osgi/blueprint test cases but spring integration is still broken. Indeed when it runs inside spring the findTransactionManager returns an instance of org.springframework.orm.jpa.JpaTransactionManager that doesn't implement javax.transaction.TransactionManager thus causing a ClassCastException here https://github.com/droolsjbpm/drools/blob/master/drools-persistence-jpa/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java#L86

PR #596, opened 2016-09-13
Fixed Timestamp parameters for LocalTaskAuditTest.testDueDateUpdateSa…

baldimir:
Just a quick fix in test - providing same time as arguments of the test method. It called getToday() twice, which can differ in milliseconds sometimes, because it is called twice.


mswiderski:
+1

PR #603, opened 2016-09-19
Change Cargo profile location in container tests

MarianMacik:
Hi,

I have made a slight change in jBPM container tests to be able to run these tests with our (QA) Cargo automation.

Thanks,

Marián


mswiderski:
@MarianMacik could you please prepare PR for master as well?

MarianMacik:
I will create the PR for master as well once I have completely debugged our automatization for 6.5.x. I think probably another PR for 6.5.x will be needed... Anyway, I will try to finish this migration (both 6.5.x and master) by the end of the week. ;)

PR #604, opened 2016-09-21
Override xnio versions because of productization kie-platform-bom

MarianMacik:
Hi,

I had to override xnio versions because otherwise these tests would not run properly with productized binaries on EAP 7. This change won't probably be needed on master.

Marián

### DIFF
--- a/jbpm-container-test/jbpm-container-test-suite/pom.xml
+++ b/jbpm-container-test/jbpm-container-test-suite/pom.xml
@@ -669,6 +669,17 @@
             <artifactId>jboss-logging</artifactId>
             <version>3.2.1.Final</version>
           </dependency>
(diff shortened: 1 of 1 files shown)


psiroky:
We definitely need this kind of override for WF10/EAP7, because the default version is aligned to EAP 6 (and it needs to stay that way).

@MarianMacik would it be possible to extract the version into property ,e.g. `version.org.jboss.xnio.wildfly10` and possibly put that directly into `kie-parent`? That way we can reuse the property for all WF10/EAP7 test modules.

MarianMacik:
Yes, it is not a problem. I will put it into kie-parent in a minute.

MarianMacik:
Thanks for merging the new property, I will change this PR so it uses the new property as well.

MarianMacik:
I have pushed a new version which uses the new property. All checks have passed.

PR #616, opened 2016-09-27
[JBPM-5343,RHBPMS-4244] fix JPA annotations to make the entities work…

psiroky:
… with both Hibernate 4 and Hibernate 5

 * there is a new profile for jbpm-installer called 'hibernate5' which
   makes it very easy to run the tests with Hibernate 5 (instead of
   the default Hibernat 4) - 'mvn clean install -Dhibernate5'
 * the change in EmailNotificationImpl should be fully backwards
   compatible as the entity name still stays the same, but is now
   explicitly set, intead of relying on hibernate to deduce the
   name from the class name

This now works on H2 (which weren't before the change). I've also tried manually on MariDB. Unfortunately the tests fail there, because the create-drop is not really working as expected.

@mswiderski, @baldimir could you please take a look?

### DIFF
--- a/jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/EmailNotificationImpl.java
+++ b/jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/EmailNotificationImpl.java
@@ -32,8 +32,8 @@ import org.kie.internal.task.api.model.EmailNotificationHeader;
 import org.kie.internal.task.api.model.Language;
 import org.kie.internal.task.api.model.NotificationType;
 
-
-@Entity
+@Entity(name = "Notification") // need to use `name = "Notification"` to make the DDL scripts work with both Hibernate 4 and Hibernate 5
+// see https://issues.jboss.org/browse/RHBPMS-4244 for more info
 @DiscriminatorValue("EmailNotification")
 public class EmailNotificationImpl extends NotificationImpl implements org.kie.internal.task.api.model.EmailNotification{
 
--- a/jbpm-installer/pom.xml
+++ b/jbpm-installer/pom.xml
@@ -46,11 +46,7 @@
       <scope>test</scope>
     </dependency>
 
-    <dependency>
-      <groupId>org.hibernate.javax.persistence</groupId>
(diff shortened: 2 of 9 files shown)


jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/EmailNotificationImpl.java
```diff
@@ -32,8 +32,8 @@
 import org.kie.internal.task.api.model.Language;
 import org.kie.internal.task.api.model.NotificationType;
 
-
-@Entity
+@Entity(name = "Notification") // need to use `name = "Notification"` to make the DDL scripts work with both Hibernate 4 and Hibernate 5
+// see https://issues.jboss.org/browse/RHBPMS-4244 for more info
```
psiroky: This is the actual fix. Rest is just test code + configuration to make it run with Hibernate 5.

mswiderski: Why adding jira as comment? Won't this be available in history as part of commit message

mswiderski:
Jenkins retest this

baldimir:
+1, thanks @psiroky. Nice work.

mswiderski:
Can you please submit same for master?

jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/EmailNotificationImpl.java
psiroky: I don't have reasonably good explanation :) Sometimes when I do something non-obvious (which in this case is naming the entity by the parent class name) I tend to also add a JIRA link to make it more visible. But I may just be overthinking this...it's true that the JIRA number will always be part of the git history.

manstis:
Personally I favour helpful comments where something is not obvious and/or
links to fuller explanation (e.g. JIRA). When looking at chunks of code I
tend not to (naturally) think "oh, let's find the commit" (and in some
environments this  may be infeasible).

PR #617, opened 2016-09-29
[6.5.x] RHBPMS-4319 - Nested sub-processes fail with SessionNotFoundException…

mswiderski:
… when aborted with error

@krisv and one more please


krisv:
+1

PR #628, opened 2016-10-07
JBPM-4646: fix case definition sorting

sutaakar:
Placed case definition sorting before paging - results have to be sorted first to provide correct input for paging.
Introduced ascending/descending switch.
Removed redundant whitespaces from test.


sutaakar:
jenkins retest this

sutaakar:
@mswiderski Can you please look on this?

mswiderski:
nice work @sutaakar thanks!

PR #629, opened 2016-10-07
JBPM-4646 - Case management - support for reopening the case, refacto…

mswiderski:
…r after move of CaseData and CaseAssignment, support for new methods of ProcessContext

depends on https://github.com/droolsjbpm/drools/pull/953

@krisv can you take a look?


krisv:
+1

PR #633, opened 2016-10-11
JBPM-5363 - Evaluation of variable expressions in subprocesses

mswiderski:
@krisv it was relatively easy to fix and might become handy, wdyt?


krisv:
+1

PR #634, opened 2016-10-13
[6.5.x] Remove unused property for Maven resolver

MarianMacik:
Hi,

I have noticed, that one property in container tests is not used anymore, so I got rid of it.

Thanks


mswiderski:
jenkins retest this

PR #636, opened 2016-10-14
[DROOLS-1327] test that uses Drools declared types in sequence flow e…

DuncanDoyle:
…xpression.


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/DroolsDeclaredTypeSequenceFlowExpressionTest.java
```diff
@@ -0,0 +1,60 @@
+package org.jbpm.bpmn2;
```
mariofusco: @mswiderski @DuncanDoyle Please add the missing header asap. We need to merge the fix in order to have a chance to include this in 6.5.0.Final

PR #639, opened 2016-10-14
RHBPMS-4370 - Remote EJB client freezes due to locking of PPI runtime…

mswiderski:
… manager on server side.

@rsynek here is the fix for the ejb locking issue


mswiderski:
Jenkins retest this

PR #642, opened 2016-10-17
JBPM-5370 - Administration service for processes and tasks - services…

mswiderski:
… api and kie server

depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/177

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
@@ -393,7 +393,7 @@ public abstract class NodeInstanceImpl implements org.jbpm.workflow.instance.Nod
     	org.jbpm.workflow.instance.NodeInstance nodeInstance = (org.jbpm.workflow.instance.NodeInstance)
     		((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer())
             	.getNodeInstance(getNode().getNodeContainer().getNode(nodeId));
-    	triggerNodeInstance(nodeInstance, null);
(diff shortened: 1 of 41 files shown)


mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
anyone?

mswiderski:
@krisv any comments or good to be merged? starting to get stale a bit here...

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
```diff
@@ -393,7 +393,7 @@ public void triggerNode(long nodeId) {
     	org.jbpm.workflow.instance.NodeInstance nodeInstance = (org.jbpm.workflow.instance.NodeInstance)
     		((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer())
             	.getNodeInstance(getNode().getNodeContainer().getNode(nodeId));
-    	triggerNodeInstance(nodeInstance, null);
+    	triggerNodeInstance(nodeInstance, org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
```
krisv: @mswiderski what is the reason for changing this?  I would expect the argument to indicate why the node was triggered, in this case the node is triggered explicitly by the user, not through an incoming connection, so wouldn't null be more adequate (to distinguish from incoming connection)?

mswiderski: @krisv the reason is that some of the NodeInstances (e.g. SubProcessNodeInstance or TimerNodeInstance) throw exception if the connection is different than the default. So instead of fixing all the node instance to check if it's null I thought it's better to simply provide the default connection when retriggering.

PR #645, opened 2016-10-18
[6.5.x] RHBPMS-4371 - Running process instances aftert migration to 6.4.0 not…

mswiderski:
… working


mswiderski:
jenkins retest this

etirelli:
+1. @mswiderski tells me it is not possible to catch this error without an integration test without different versions that we aren't able to do at the moment.

PR #647, opened 2016-10-18
[6.5.x] JBPM-5374 - Process instance doesn't exit an Embedded SubProcess in an AdHoc process

mswiderski:


livthomas:
+1 ([RHBPMS-4378](https://issues.jboss.org/browse/RHBPMS-4378) seems to be acked)

mswiderski:
merged manually

PR #648, opened 2016-10-19
JBPM-4646 - Case management - simplified use for case role assignment…

mswiderski:
…s, assign correlation key for each process instance, store case data to audit log upon case instance start

depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/178

@krisv here are the changes we talked about


mswiderski:
jenkins retest this

krisv:
+1

PR #650, opened 2016-10-20
[6.4.x] Backports for  6.3.4

mswiderski:
RHBPMS-4208 - [GSS] (6.3.z) Unable to get log tables precision up to milliseconds when using MariaDB/MySQL
RHBPMS-4262 - [GSS] (6.3.z) Signal not getting invoked externally through rest API when it is in Error Scenario


mswiderski:
@krisv here are the back ports for 6.3.4

krisv:
+1

PR #657, opened 2016-10-31
JBPM-5239 Non-blocking command executor

nikolobok:
Non blocking AbstractAvailableJobsExecutor with command callback using CommandAsync interface.
Interfaces is in droolsjbpm-knowledge project (JBPM-5239)  
PR set:
https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/184
https://github.com/droolsjbpm/jbpm/pull/657

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/AsyncCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/AsyncCommand.java
@@ -1,0 +1,33 @@
+package org.jbpm.executor.commands;
+
+import org.jbpm.executor.ExecutorServiceFactory;
+import org.kie.api.executor.CommandAsync;
+import org.kie.api.executor.CommandContext;
+import org.kie.api.executor.ExecutionResults;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Simple async command to call external service without blocking executor thread
+ * for example webservice with callback or SOAP over JMS
+ * Just for demo purpose.
+ *
+ */
+public class AsyncCommand implements CommandAsync {
+    private static final Logger logger = LoggerFactory.getLogger(AsyncCommand.class);
+
+    public void execute(CommandContext ctx, Long requestId) {
+        logger.info("Command executed on executor with data {}", ctx.getData());
+        new Thread(new Runnable() {
+            @Override
+            public void run() {
+                try {
+                    Thread.sleep(100);
+                } catch (InterruptedException e) {
+                    e.printStackTrace();
+                }
+                ExecutorServiceFactory.newExecutorService().scheduleResponse(requestId, new ExecutionResults());
+            }
+        }).start();
+    }
+}
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
@@ -31,14 +31,7 @@ import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.jbpm.executor.entities.ErrorInfo;
 import org.jbpm.executor.entities.RequestInfo;
 import org.jbpm.executor.impl.event.ExecutorEventSupport;
-import org.kie.api.executor.Command;
-import org.kie.api.executor.CommandCallback;
-import org.kie.api.executor.CommandContext;
-import org.kie.api.executor.ExecutionResults;
-import org.kie.api.executor.ExecutorQueryService;
-import org.kie.api.executor.ExecutorStoreService;
-import org.kie.api.executor.Reoccurring;
-import org.kie.api.executor.STATUS;
+import org.kie.api.executor.*;
 import org.kie.internal.runtime.manager.InternalRuntimeManager;
 import org.kie.internal.runtime.manager.RuntimeManagerRegistry;
 import org.slf4j.Logger;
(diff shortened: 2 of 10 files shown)


nikolobok:
PR set:
https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/184
https://github.com/droolsjbpm/jbpm/pull/657

mswiderski:
@nikolobok thanks for sending the PR it makes it easier to discuss what is the actual need. First thing that comes to my mind is that I don't see clear benefits of moving the execution of the command to another thread.. as it's already running in separate on - managed by jbpm executor. 

Next, looks like we don't need another type of command but more another type of results and then logic around handling the results. The implementation of a command can still be the same and even spin another thread to do the processing and then directly return AyncExecutionResults that would work similar to Future where it can either block when asked for results or use callback to notify when it's done. This will then better fit the design behind executor as the only part that has to be changed is how executor deals with response/results.

nikolobok:
I`ve assumed using the callback style to delegate execution to different system, in my case it should be soap over jms (request/response) service, this unit of work not managed by jbpm executor moreover it could be a different machine.
That is why it should not block any thread in jbpm jvm and that is why we can not use Future.
May be it could be done by using CompletableFuture. 
Anyway it needs different Command interface and other logic would be the same as in PR except not calling scheduleResponse() directly from user code.
Is Java 8 requirement would be affordable for JBPM project? If so I will try to code it using CompletableFuture.

nikolobok:
I have looked more about CompletableFuture, so it could not be used in this case. Because here we should completely break execution flow, and answer could be received on different server.

mswiderski:
java 8 is the minimum for version 7 (master).

Sorry, I still don't see this use case... so what you try to do is not block any thread on jbpm then you should not use executor at all. You should have your own handler that simply sends the message to jms/soap and then waits for signal when the job is completed. That signal should carry on all the data it has. What's the point of using executor here?

nikolobok:
@mswiderski The main point do not overload bpmn diagram with technical data about implementing service call.
When we call request-response SOAP web service with regular HTTP transport, we only have one activity in bpmn diagram, why it should be different when we change transport to jms, not interface nor business operation.
Secondly I believe signals are good for one-way conversation or 	
separated  interfaces one for request and other for response so they show a real business mean about waiting and receiving some data. But I want encapsulate two-way service call regardless of it`s implementation.

mswiderski:
@nikolobok that's all good and I agree with what you described - don't change the process when technical/implementation details changed. And that's why there is work item handlers that can be replaced without changing the process at all. 
What I don't get why you would like to change the way jbpm executor works? Wouldn't using work item handler impl be a better fit here?

nikolobok:
@mswiderski I believe it can`t be done without patching executor, of course we could implement something that will replace executor but why if we already have working code, that we can extend with backward compatibility. So executor is default way to implement async execution for jbpm, only I have done is append some new states to executor jobs, I have not changed default (blocking) way executor works, there is no any need to change existing code.

mswiderski:
if you see executor is valid for you use case you could use it without any modifications. What you actually do in the pull request is to schedule "response" once the job is completed with execution. 

But why do you need to change it to "response"? why simply not to schedule another command that will get the the data (input) and copy to results? 

Because what you have done in schedule Response is pretty much scheduling another command execution - but instead of giving a new request you update the state which means (from executor point of view) second execution of the same job - that is against the idea behind the executor. 

So instead of extending the executor itself I'd recommend to give it a try with using its capabilities to achieve what you need - I am quite confident you can already do that with command chaining.

nikolobok:
@mswiderski Well I have thought about it, but in this case we will need somehow handle all CommandContext  over invocation of remote service, and it is not possible in general, and some other troubles arise. 
But when we switching status to WAITING_ASYNC in case of new CommandAsync job implementation, it does not pass status DONE and we could transfer over invocation only correlation id of request, then receive it, find job in WAITING_ASYNC status, and pass it to executor to complete it to status DONE.
So we does not pass same job twice, we just making delay in it`s execution, and we have all needed data to correctly manage this job when it waiting an answer, it could be killed for example.
When there is error in request it will be handled as usual, by current code, tried X times, and then break.
If we mark it as recur it will be correctly repeated as new job, with new id, by existing code.
All will be as usual, except of non-blocking pause of execution in case of CommandAsync.

nikolobok:
I understand your attention to retain executor logic stable and not change it.
And in case of this PR the default life cycle did not changed, it just extends in case of using this new async command feature, in all other cases it keeps previous behavior.

mswiderski:
actually it does change the behavior slightly (and that's why I mentioned command chaining) as the scheduleResponse calls the event listener job scheduled event so it's actually marks is as new command - in the way executor treats the commands.

What I am trying to say is that this behavior can be achieved without changing the life cycle of the executor. You can do all that stuff with custom commands and callbacks. When you async command completes you can call another command explicitly or via callback. Because the changes in this PR directly do that via additional methods - waiting async is considered as queued or retrying so why not simply use these states to perform the completion?

nikolobok:
@mswiderski This behavior is only when we use new CommandAsync interface, and yes event listener will be fired in this case, but why not? anyway I can change it. BUT in case of regular Command executor job life cycle did not changed at all, event fired only once.
As about do this stuff by custom work item handler it can`t be done, because executor will complete work item after command returned result and will move process forward to next item. But it is not what should be in case of two-way service call from business (BPMN) point of view.

mswiderski:
> This behavior is only when we use new CommandAsync interface, and yes event listener will be fired in this case, but why not?

because that would mean there is a new command scheduled while it is not, so it's abusing the executor mechanism to let the command be re-excuted to return results.

> As about do this stuff by custom work item handler it can`t be done, because executor will complete work item after command returned result and will move process forward to next item. But it is not what should be in case of two-way service call from business (BPMN) point of view.

it can be done, as your command is in full control over the execution and if you don't want to complete the work item simply remove callbacks from the command context and then none will be executed.

PR #661, opened 2016-11-04
JBPM-5396 RHBPMS-4443 SQL upgrade scripts contain numeric(19,0) datatype instead of bigint

baldimir:
- Replaced numeric(19,0) with bigint.


mswiderski:
jenkins retest this

PR #663, opened 2016-11-04
[DROOLS-1353] contextual conversation

mariofusco:


mariofusco:
Merged with https://github.com/droolsjbpm/jbpm/pull/673

PR #665, opened 2016-11-07
JBPM-4646 - Case management - improved handling of missing comments

sutaakar:
If user tried to update or delete not existing comment then NoSuchElementException was thrown.

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
@@ -36,6 +36,7 @@ import java.util.UUID;
 
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
 import org.jbpm.casemgmt.api.CaseActiveException;
+import org.jbpm.casemgmt.api.CaseCommentNotFoundException;
 import org.jbpm.casemgmt.api.CaseNotFoundException;
 import org.jbpm.casemgmt.api.model.AdHocFragment;
 import org.jbpm.casemgmt.api.model.CaseDefinition;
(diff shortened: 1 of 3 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -955,6 +956,56 @@ public void testCaseWithComments() {
...
+        units.add(deploymentUnit);
+        Map<String, Object> data = new HashMap<>();
+        CaseFileInstance caseFile = caseService.newCaseFileInstance(deploymentUnit.getIdentifier(), USER_TASK_STAGE_AUTO_START_CASE_P_ID, data);
+
+        String caseId = caseService.startCase(deploymentUnit.getIdentifier(), USER_TASK_STAGE_AUTO_START_CASE_P_ID, caseFile);
+        assertNotNull(caseId);
+        assertEquals(FIRST_CASE_ID, caseId);
+        try {
+            caseService.removeCaseComment(FIRST_CASE_ID, "not-existing-comment");
+            fail("Not existing comment cannot be deleted.");
```
jhrcek: Seeing this assertion message in failing test doesn't make it immediately clear why the test failed.
Something like "Removing non-existent comment from a case should throw CaseCommentNotFoundException" would make the failure much more meaningful

sutaakar: Good point, will update it.

mswiderski: maybe it's just me but what's there right now for me does say all that it should - when you find this error in your tests suite and the name of the method you have all that is needed, don't you think so?

PR #666, opened 2016-11-07
[6.4.x] JBPM-5398 - Transaction issues with Timer catch event

mswiderski:
@sutaakar here is a change we talked about that should fix the problem with kie server tests

### DIFF
--- a/jbpm-bpmn2/src/test/resources/BPMN2-BusinessRuleTaskWithDataInput.drl
+++ b/jbpm-bpmn2/src/test/resources/BPMN2-BusinessRuleTaskWithDataInput.drl
@@ -21,7 +21,7 @@ import org.jbpm.bpmn2.objects.Person;
 rule "Person rule" ruleflow-group "MyRuleFlow"
 no-loop
     when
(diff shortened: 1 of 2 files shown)


mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

jbpm-bpmn2/src/test/resources/BPMN2-BusinessRuleTaskWithDataInput.drl
```diff
@@ -21,7 +21,7 @@ import org.jbpm.bpmn2.objects.Person;
 rule "Person rule" ruleflow-group "MyRuleFlow"
 no-loop
     when
-        $p : Person()
+        $p : Person(name == null)
```
psiroky: This should not be needed. The issue seems to be with no-loop + eager evaluation directly in Drools. I believe @mariofusco fixed it for 6.5.x and master.

mswiderski:
won't be included in 6.4.x

PR #670, opened 2016-11-08
Modify DefaultSignalManager for concurrency

jcavieres:
There is a concurrency problem in the class org.jbpm.process.instance.event.DefaultSignalManager.
It is not threadSafe because the variable processEventListeners is of type HashMap and because the creation of processEventListeners and eventListeners is not synchronized. This proposed change solves that problem.


jcavieres:
@mswiderski please give "ok to test" to this new PR

mswiderski:
ok to test

mswiderski:
@jcavieres can you open jira and update commit message so we have the reference?

jcavieres:
Done
https://issues.jboss.org/browse/JBPM-5400

mswiderski:
merged manually with https://github.com/droolsjbpm/jbpm/commit/69057f7e9e3a421ce880e774387bb1b932c45608

thanks @jcavieres

PR #673, opened 2016-11-10
[DROOLS-1353] contextual conversation

mswiderski:


jbpm-services/jbpm-shared-services/pom.xml
```diff
@@ -53,6 +53,10 @@
       <artifactId>logback-classic</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.as</groupId>
+      <artifactId>jboss-as-arquillian-protocol-jmx</artifactId>
```
mswiderski: I don't think is needed so feel free to remove it. Or I'll remove it on Monday

PR #675, opened 2016-11-14
[6.4.x] RHBRMS-2621 - [GSS] (6.3.z) Errors in Phreak under heavy and multi th…

mswiderski:
…readed load

changed default behavior as requested in RHBRMS-2621
@krisv can you take a quick look?


krisv:
+1

PR #676, opened 2016-11-15
JBPM-4646 - Case management - merge case milestones from instance and…

mswiderski:
… definition to provide complete view

@krisv @cristianonicolai here is the change for milestones


krisv:
That was fast ;)
+1

PR #677, opened 2016-11-15
Memory Leak in DefaultSignalManager

jcavieres:
Every time a process is started a node is added to the map processEventListeners and it's never removed. This PR fixes that.


jcavieres:
A issue was created for this:
https://issues.jboss.org/browse/JBPM-5413

mswiderski:
please rebase with master branch as it has conflicts now

jcavieres:
fixed

mswiderski:
ok to test

mswiderski:
Would it be possible to add a test case for it to make sure it won't happen again?

jcavieres:
I found the problem using a memory analyzer. I don't know how to put that in a test case.

mswiderski:
Just check the DefaultSignalManager if the event listener was removed

jcavieres:
Maybe it's a silly question, but how do I get the DefaultSignalManager
instance from the KieSession used to start the process in the test case?

JCS

On Thu, Nov 17, 2016 at 11:07 AM, Maciej Swiderski <notifications@github.com

mswiderski:
you can try with this:
```
((InternalProcessRuntime)((InternalKnowledgeRuntime)(ksession)).getProcessRuntime()).getSignalManager()
```

manstis:
That looks easily accessible.. just rolls off the keyboard ;-)

jcavieres:
The problem is "processEventListeners.size()" can't be accessed because
processEventListeners is private .

private Map<String, List<EventListener>> processEventListeners

mswiderski:
try just registering event listener for given type then call it so it gets invoked then remove it and send a signal again to check that the listener was not invoked. Does it make sense?

mswiderski:
@manstis it should not be really easy accessible from outside, but anyway would be good to test it does work as expected :)

@jcavieres any updates?

jcavieres:
@mswiderski your idea doesn't work because with or without the leak the
result is the same. The reason is this piece code in internalSignalEvent():

List<EventListener> eventListeners = processEventListeners.get(type);
if (eventListeners != null) {
                for (EventListener eventListener: eventListeners) {
                    eventListener.signalEvent(type, event);
                }
}

With the memory leak, the second time signalEvent is called there isn't any
*eventListener* element in the "for" loop, so signalEvent is never executed
.

Without the memory leak, the second time signalEvent is called
*evenListeners* (with the final *s*) is null, so signalEvent is never
executed.







On Fri, Nov 25, 2016 at 10:36 AM, Maciej Swiderski <notifications@github.com

mswiderski:
@jcavieres sorry for the delay, finally got some time to look at it. 

would it be possible to get the process you use to reproduce this problem to be attached to jira? I am having hard time understanding where is the memory leak. Especially after your last comment - there there is either empty list or null.

jcavieres:
Hi, it's been a long time and I don't have the process available anymore,
but the leak is not hard to find:

In the original *DefaultSignalManager.java* there is an element added in
"processEventListeners.put(type, eventListeners)" that is never removed.
My change adds that missing processEventListeners.remove(type) command;

The "eventListeners == null" part is for accessing the synchronized part of
the code as little as possible .

Regards

JCS


On Tue, Feb 14, 2017 at 11:05 AM, Maciej Swiderski <notifications@github.com

PR #680, opened 2016-11-17
Proposed fix for [DROOLS-1364]

eliasisrael:
Proxy KieSession objects in InMemorySessionFactory so that they can
be removed from the internal sessions collection on dispose/destroy


eliasisrael:
Please note the second commit, which fixes a small issue with the original proposal.

mswiderski:
I'd prefer to avoid any kind of proxies and instead extend the SessionFactory interface to provide method to remove the ksession instead.

eliasisrael:
I understand your reasoning on this and probably would have approached it in a similar fashion except that the code included a request for a proxy, and it would have been a much bigger change to hook into a listener somewhere (which, right now, I'm not even sure exists for the needed event).

mswiderski:
closing as this should be addressed differently and for the case describe din DROOLS-1364 - runtime manager should not be used

PR #681, opened 2016-11-17
[JBPM-5285] Custom JNDI name for AvailableJobsExecutor in ExecutorServiceFactory

alberto-brigandi:
This PR allows to define the JNDI lookup name of AvailableJobsExecutor implementation in ExecutorServiceFactory, thus allowing to load a customized implementation

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/ExecutorServiceFactory.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/ExecutorServiceFactory.java
@@ -51,7 +51,7 @@ public class ExecutorServiceFactory {
 	private final static String mode = System.getProperty( "org.jbpm.cdi.executor.mode", "singleton" );
 	
 	private final static String availableJobsExecutorName = System.getProperty(
-	    "org.jbpm.cdi.available.jobs.executor.name", "java:module/AvailableJobsExecutor" );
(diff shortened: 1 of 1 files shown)


mswiderski:
ok to test

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/ExecutorServiceFactory.java
```diff
@@ -49,6 +49,9 @@
 public class ExecutorServiceFactory {
 
 	private final static String mode = System.getProperty( "org.jbpm.cdi.executor.mode", "singleton" );
+	
+	private final static String availableJobsExecutorName = System.getProperty(
+	    "org.jbpm.cdi.available.jobs.executor.name", "java:module/AvailableJobsExecutor" );
```
livthomas: Would it be possible to change this system property to something easier and self-explanatory? Since there is already `org.jbpm.cdi.executor.mode` used in this class, I would suggest using `org.jbpm.cdi.executor.jndi` for this new property.

mswiderski: +1 at changing the propert name to be aimpler

mswiderski:
@alberto-brigandi do you mind changing the property name?

mswiderski:
jenkins retest this

mswiderski:
thanks @alberto-brigandi

PR #683, opened 2016-11-21
JBPM-5415 : KModuleDeploymentService doesn't add multiple @Remotable …

tkobayas:
…classes in a dependent jar


jbpm-services/jbpm-kie-services/src/test/resources/multi-remotable/pojo-pom.xml
```diff
@@ -0,0 +1,18 @@
...
+
+ <groupId>org.test</groupId>
+ <artifactId>jbpm-kie-services-filter-test-pojo</artifactId>
+ <version>1.0.0-SNAPSHOT</version>
+
+ <dependencies>
+  <dependency>
+   <groupId>org.kie</groupId>
+   <artifactId>kie-api</artifactId>
+   <version>6.4.0.Final</version>
```
psiroky: I believe we should not depend on `6.4.0.Final` here, even if just for tests. Could we use property + maven-resources-plugin filtering?

PR #687, opened 2016-11-21
JBPM-5416 - Update workitem configuration to allow for default values…

tsurdilo:
… of some / all parameters


mswiderski:
looks good to me

jenkins retest this

mswiderski:
@tsurdilo seems there is a test failure related to this change, can you please have a look?

tsurdilo:
@mswiderski amended and added the choice to use enum type or string like we talked about

PR #688, opened 2016-11-22
[DROOLS-1353] define ExecutableRunner interface

mariofusco:


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
```diff
@@ -118,8 +118,7 @@ public void dispose() {
     private TimerManager getTimerManager(KieSession ksession) {
         KieSession internal = ksession;
         if (ksession instanceof CommandBasedStatefulKnowledgeSession) {
-            internal =  ((RegistryContext) ((CommandBasedStatefulKnowledgeSession) ksession)
-                    .getCommandService().getContext()).lookup( KieSession.class );
+            internal = ( (SingleSessionCommandService) ( (CommandBasedStatefulKnowledgeSession) ksession ).getRunner() ).getKieSession();;
```
rsynek: One ';' more than necessary. 

Secondly, there is a lot of type casting in this PR, I wonder if it would be possible to redesign the API to allow accessing  these types without casting - or at least to have it on a single place.

PR #691, opened 2016-11-28
JBPM-4646 - Case management - boundary events are not ad hoc fragments

mswiderski:
@krisv here is the fix for boundary events to not be seen as ad hoc fragments


krisv:
+1, thanks Maciej !

PR #696, opened 2016-12-07
Add revapi filter

MarianMacik:
Hello,

I have made a small change in a revapi config before upgrading it to 0.8.0. Please do not merge it yet.

Thanks,

Marian

### DIFF
--- a/jbpm-services/jbpm-services-api/pom.xml
+++ b/jbpm-services/jbpm-services-api/pom.xml
@@ -49,40 +49,6 @@
       <plugin>
         <groupId>org.revapi</groupId>
         <artifactId>revapi-maven-plugin</artifactId>
-        <configuration>
-          <oldArtifacts>
-            <artifact>${project.groupId}:${project.artifactId}:${version.org.kie.latestFinal.release}</artifact>
-          </oldArtifacts>
-          <!-- By default revapi will check the oldArtifact against the currently executed build -->
-          <analysisConfigurationFiles>
-            <configurationFile>
-              <path>src/build/revapi-config.json</path>
-            </configurationFile>
-          </analysisConfigurationFiles>
-          <reportSeverity>nonBreaking</reportSeverity>
-        </configuration>
-        <!-- Running two executions is a workaround to make sure we get a HTML report in case revapi finds
-             some incompatible changes. The "check" goal will simply fail the whole build before it could get
-             to the report. To make sure we always get a HTML report, the "report" goal needs to be executed
-             before the "check" goal.
-             Once https://github.com/revapi/revapi/issues/11 is fixed it should be possible to use single execution. -->
-        <executions>
-          <execution>
-            <id>check</id>
-            <goals>
-              <goal>check</goal>
-            </goals>
-            <phase>verify</phase>
-          </execution>
-          <execution>
-            <!-- report can be found in ${build.directory}/site/revapi-report.html -->
-            <id>report</id>
-            <goals>
-              <goal>report</goal>
-            </goals>
-            <phase>package</phase>
-          </execution>
-        </executions>
       </plugin>
     </plugins>
   </build>
--- a/jbpm-services/jbpm-services-api/src/build/revapi-config.json
+++ b/jbpm-services/jbpm-services-api/src/build/revapi-config.json
@@ -1,6 +1,17 @@
 {
   "revapi": {
-    "_comment": "Changes between 6.4.0.Final and master. These changes are desired and thus ignored. They should be removed when 7.0.0.Final is available.",
+    "java": {
+      "filter": {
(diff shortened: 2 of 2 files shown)


MarianMacik:
jenkins retest this

MarianMacik:
@psiroky feel free to merge now.

jbpm-services/jbpm-services-api/pom.xml
```diff
@@ -60,6 +60,7 @@
             </configurationFile>
           </analysisConfigurationFiles>
           <reportSeverity>nonBreaking</reportSeverity>
+          <failSeverity>potentiallyBreaking</failSeverity>
```
psiroky: Is this something we could move to the kie-parent? Or do we expect to have different failSeverity in different modules (and similarly for other config options).

jbpm-services/jbpm-services-api/src/build/revapi-config.json
```diff
@@ -42,6 +53,18 @@
         "old": "field org.jbpm.services.api.query.QueryResultMapper<T>.COLUMN_ORGANIZATIONAL_ENTITY",
         "new": "field org.jbpm.services.api.query.QueryResultMapper<T>.COLUMN_ORGANIZATIONAL_ENTITY",
         "justification": "Fix for some of dbs with query aliases"
+      },
+      {
+        "code": "java.field.enumConstantOrderChanged",
+        "old": "field org.jbpm.services.api.query.model.QueryDefinition.Target.CUSTOM",
+        "new": "field org.jbpm.services.api.query.model.QueryDefinition.Target.CUSTOM",
+        "justification": "ADD YOUR EXPLANATION FOR THE NECESSITY OF THIS CHANGE"
```
psiroky: Should we fix the justification before the merge?

MarianMacik:
jenkins retest this

MarianMacik:
Failed tests have nothing to do with the revapi plugin.

jbpm-services/jbpm-services-api/src/build/revapi-config.json
mswiderski: is this really relevant? I mean changing order in enum?

though I agree we should not commit with "ADD YOUR EXPLANATION.." thingy

triceo: @mswiderski As to the relevance, unfortunately, iteration order of enums is part of the API. 
+1 on not committing without justifications, although I think they are better filled in by the committers of such breaking changes.

mswiderski: can we add justification to be:
included more out of the box targets for query definition that are more specific than custom and thus placed higher in the enum.

MarianMacik:
jenkins retest this

jbpm-services/jbpm-services-api/src/build/revapi-config.json
```diff
@@ -42,6 +53,18 @@
...
+        "code": "java.field.enumConstantOrderChanged",
+        "old": "field org.jbpm.services.api.query.model.QueryDefinition.Target.CUSTOM",
+        "new": "field org.jbpm.services.api.query.model.QueryDefinition.Target.CUSTOM",
+        "justification": "Included more out of the box targets for query definition that are more specific than custom and thus placed higher in the enum"
+      },
+      {
+        "code": "java.field.serialVersionUIDUnchanged",
+        "old": "field org.jbpm.services.api.query.model.QueryParam.serialVersionUID",
+        "new": "field org.jbpm.services.api.query.model.QueryParam.serialVersionUID",
+        "justification": "ADD YOUR EXPLANATION FOR THE NECESSITY OF THIS CHANGE"
```
psiroky: Do we still wan to wait for this explanation? @MarianMacik I remember we talked about this, but we never came to a conclusion as I am not sure what to put here. But I guess anything is better than "ADD YOUR...". Worst case we could add at least something like "We don't care about that" or "We don't really know why is this being reported" :)

MarianMacik: I have done a little bit of a research and this means that the serial version UID stayed unchanged even though the default serial version UID generated by a compiler would be different. I think it might be a good "warning" in the future, but I agree that in this case it probably does not need a change since only a couple of static methods were added. I would agree to add justification like "Not an issue, only static methods were added etc." What do you think?

mswiderski: +1 from me on this comment/justification

psiroky: +1, sounds good. Thanks @MarianMacik.

MarianMacik:
Jenkins retest this

PR #699, opened 2016-12-07
Don't start/stop container with -DskipTests

psiroky:
Before the change, when running `mvn clean install -DskipTests` the container would still get started/stopped, which takes considerable amount of time and is not needed in such case.

@rsynek could you please quickly check this?


mswiderski:
@rsynek any comments? or shall I merge this?

rsynek:
+1 to merge, thanks!

PR #700, opened 2016-12-07
[6.5.x] RHBPMS-4518: Add engine-level reproducer

livthomas:
This adds engine-level reproducer for [RHBPMS-4518](https://issues.jboss.org/browse/RHBPMS-4518). Although it is a bug in the engine it mostly influence Workbench remote APIs since signalling there is done through `SignalEventCommand`.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2014 - 2016 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


livthomas:
The reproducer is `testExecuteSignalEventCommandWithPerProcessInstanceStrategy` which fails but it seems it also affects other tests. I will try to fix it.

livthomas:
The tests have been updated and only the reproducer should fail now.

mswiderski:
@livthomas I know this is reproducer but specific case that refers to lie-remote but from jbpm services point of view it's wrong usage. You should use https://github.com/droolsjbpm/jbpm/blob/master/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/ProcessService.java#L127

taking into consideration RHBPMS-4518 was nack'ed and won't be fixed I opt at closing this. wdyt?

livthomas:
@mswiderski, I know there is a better way to send a signal using `ProcessService` and I have just tried to reproduce the problem in the underlying functionality of KIE remote service. Since the [RHBPMS-4518](https://issues.jboss.org/browse/RHBPMS-4518) has not been approved, we can close this.

PR #705, opened 2016-12-16
JBPM-5472: static inner classes in CompositeNode and ForEachNode

marianbuenosayres:
We're working on serialization mechanisms for the process definition, and to have them work we need to be able to recreate the inner classes of ForEachNode and CompositeNode to allow serialization /deserialization of the object to be possible from outside the class.

### DIFF
--- a/jbpm-flow/src/test/java/org/jbpm/workflow/core/node/NodeInnerClassesTest.java
+++ b/jbpm-flow/src/test/java/org/jbpm/workflow/core/node/NodeInnerClassesTest.java
@@ -1,0 +1,63 @@
+package org.jbpm.workflow.core.node;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.drools.core.process.core.datatype.impl.type.ObjectDataType;
+import org.jbpm.process.core.context.variable.Variable;
+import org.jbpm.process.test.TestProcessEventListener;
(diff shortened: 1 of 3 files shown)


mswiderski:
Could you please create test case and jira for this? Without test this can later on be removed during refactoring so we need to have this covered

marianbuenosayres:
I'll create a test case and jira. Also, is there a way of gaining access ot the jenkins server logs? I tried running the tests locally and couldn't see any errors, so it would be useful to fix them if required.

marianbuenosayres:
Added a test and this is the JIRA ticket created for this fix: JBPM-5472

jbpm-flow/src/test/java/org/jbpm/workflow/core/node/NodeInnerClassesTest.java
```diff
@@ -0,0 +1,63 @@
+package org.jbpm.workflow.core.node;
```
mswiderski: missing license header, could you please add it?

mswiderski:
if you could just update commit message with jira issue that would be excellent

marianbuenosayres:
I cannot change the commit message already, but I can create a new branch replicating the changes on a new branch, create a new pull request, and close this one. 

The new PR is https://github.com/droolsjbpm/jbpm/pull/706

PR #711, opened 2016-12-23
[6.5.x] RHBPMS-4548 - RuntimeDataService 'getTaskAssignedAsPotentialOwner' ru…

mswiderski:
…ns single threaded and therefore doesn't scale


mswiderski:
jenkins retest this

PR #713, opened 2017-01-03
Add a profile to turn-off container setup by cargo.

rsynek:
Gives a possibility to override cargo execution and run tests from QE environment.


jbpm-container-test/jbpm-remote-ejb-test/jbpm-remote-ejb-test-suite/pom.xml
```diff
@@ -463,7 +463,35 @@
...
+            <executions>
+              <!-- do nothing, container is not managed by cargo -->
+              <execution>
+                <id>start-container</id>
+                <phase>none</phase>
+              </execution>
+              <execution>
+                <id>stop-container</id>
+                <phase>none</phase>
+              </execution>
```
psiroky: I know this is what I originally did in kie-server tests, but I believe nowadays there is a better way. Simple setting property `cargo.maven.skip` to `true` will disable all Cargo related executions (which we want in this case). WDYT?

rsynek: As we discussed, setting `eap7.download.url` default value to a valid URL makes it possible to skip the cargo execution using `cargo.maven.skip` without cargo complaining about invalid URL.

PR #715, opened 2017-01-03
Revert "Fix tests in CI by using local maven repo in target/"

psiroky:
This reverts commit 1bdde2350e7cff1fcb8c42398750228764b03a5c.

This is no longer needed because the CI machines always start with
empty ~/.m2/repository, so there is no longer a possibility to
use outdated/wrong SNAPSHOTs from there.

Also, this custom settings.xml makes it harder to run release
builds as it only consideres the Public JBoss Nexus when
downloading artifacts. For release builds we need also the
staging repositories, so some of the tests are failing as they
can not find those not yet released versions in Pubblic Nexus.


psiroky:
The test failures seem to be related, I'll need to take a closer look.

mswiderski:
@psiroky looks ok to me, is this read to be merged?

psiroky:
Yes, this is ready now.

PR #720, opened 2017-01-06
Fix POMs after IP BOM 7.0.0.CR6 upgrade

psiroky:


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-client/pom.xml
```diff
@@ -20,17 +20,6 @@
...
-    <!-- uncomment to test with application deployed on JBoss EAP 6.1 -->
-    <!-- 
-    <dependency>
-      <groupId>org.jboss.as</groupId>
-      <artifactId>jboss-as-ejb-client-bom</artifactId>
-      <version>7.2.0.Final</version>
-      <optional>true</optional>
-      <type>pom</type>
-    </dependency>
-    -->
```
mswiderski: Why removing this? I usually use this when I need to work on remote ejb debugging so it's quite handy

psiroky: I must have missed the comment, so just removed the commented code. However, we currently have separate test suite for ejb (https://github.com/droolsjbpm/jbpm/tree/master/jbpm-container-test/jbpm-remote-ejb-test), so we need to figure how many tests are being duplicated. 

Also, if we wanna keep it like this we need to upgrade to WFLY10/EAP7, we are not supporting AS7/EAP6 anymore. Also I think it may be better to add this so some specific profile (even though having too many profiles is bad as well) to avoid commented code.

psiroky: https://github.com/droolsjbpm/jbpm/tree/master/jbpm-container-test/jbpm-remote-ejb-test

mswiderski: +1 to upgrade to wildfly but I'd prefer to keep this as it is simple to use when working on code changes while the container tests are good in builds

PR #723, opened 2017-01-11
Reproducer for Case file conditional event fails other cases

ibek:


ibek:
The tests will be failing till this issue is fixed - https://issues.jboss.org/browse/JBPM-5541

mswiderski:
reproducer included in https://github.com/droolsjbpm/jbpm/pull/724 though it's still not decided if that is going to be solved or how.

PR #724, opened 2017-01-12
JBPM-5541 - Case file conditional event fails other cases

mswiderski:
depends on https://github.com/droolsjbpm/drools/pull/1060

@krisv here is a change to use entry points for case file that is named same as case definition (ad hoc process id) that way case definitions conditions are isolated to its own type/definition so they don't interfere.

The problem here is that all conditions (drools rules) are compiled into kbase and shared globally for all process definitions. This is not really a case mgmt issue but overall problem though it will be mostly visible in cases where conditions will usually refer to CaseData (aka CaseFile).

To be able to support layered case instances - case definitions + subprocesses to access easily same file the entry point (process id of the case def) is stored as process variable. That way it is always given to child processes and can be used to look up the right entry point to find case file. 

What do you think? I couldn't find any other applicable solution than this... so open for any ideas.

Only other alternative is to put the requirement on end users to keep in mind that when there are multiple case definitions with conditions (which most likely be rather normal case) then conditions must be distinct like using case definition id as part of the conditions. Not always possible and easy to mis... especially that it can't be caught on compile time but on runtime.

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
@@ -30,6 +30,7 @@ import java.util.Collections;
 public abstract class CaseCommand<T> implements ExecutableCommand<T> {
 
     private static final long serialVersionUID = 4116744986913465571L;
+    protected static final String ENTRY_POINT_VAR_NAME = "_EntryPoint_";
     
     private CaseEventSupport emptyCaseEventSupport = new CaseEventSupport(Collections.emptyList());
 
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/org/jbpm/casemgmt/demo/insurance/insurance-rules.drl
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/org/jbpm/casemgmt/demo/insurance/insurance-rules.drl
@@ -6,7 +6,7 @@ import org.jbpm.casemgmt.demo.insurance.ClaimReport;
 rule "classify-claim" ruleflow-group "claims"
 
 when 
-    $caseData : CaseData()
+    $caseData : CaseData() from entry-point "insurance-claims.CarInsuranceClaimCase" 
     $claim : ClaimReport( ) from $caseData.getData("claimReport")      
 then 
     System.out.println("Classified claim " + $claim);
--- a/jbpm-flow-builder/src/main/java/org/jbpm/compiler/ProcessBuilderImpl.java
+++ b/jbpm-flow-builder/src/main/java/org/jbpm/compiler/ProcessBuilderImpl.java
@@ -15,6 +15,16 @@
  */
 package org.jbpm.compiler;
 
+import java.io.IOException;
+import java.io.Reader;
+import java.io.StringReader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import javax.xml.parsers.FactoryConfigurationError;
+
 import org.drools.compiler.builder.impl.KnowledgeBuilderConfigurationImpl;
 import org.drools.compiler.builder.impl.KnowledgeBuilderImpl;
 import org.drools.compiler.compiler.BaseKnowledgeBuilderResultImpl;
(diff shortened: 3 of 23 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
```diff
@@ -30,6 +30,7 @@
 public abstract class CaseCommand<T> implements ExecutableCommand<T> {
 
     private static final long serialVersionUID = 4116744986913465571L;
+    protected static final String ENTRY_POINT_VAR_NAME = "_EntryPoint_";
```
krisv: Have some doubts about using a variable for storing this, as this will result in an internal implementation detail be very visible in the process instance details.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/org/jbpm/casemgmt/demo/insurance/insurance-rules.drl
```diff
@@ -6,7 +6,7 @@ import org.jbpm.casemgmt.demo.insurance.ClaimReport;
 rule "classify-claim" ruleflow-group "claims"
 
 when 
-    $caseData : CaseData()
+    $caseData : CaseData() from entry-point "insurance-claims.CarInsuranceClaimCase"
```
krisv: Doesn't this introduce backwards compatibility issues, and pushes the burden partially to the user as well (just in a different location, now the rules are specifying the entrypoint rather than the rules having to make sure they are independent from other cases if necessary.

jbpm-flow-builder/src/main/java/org/jbpm/compiler/ProcessBuilderImpl.java
```diff
@@ -426,7 +438,7 @@ private String createSplitRule(Process process,
         		connection.getToType() + "\"  @Propagation(EAGER) \n" +
         	"      ruleflow-group \"DROOLS_SYSTEM\" \n" +
         	"    when \n" +
-        	"      " + constraint + "\n" +
+        	"      " + constraint + getEntryPoint(constraint, process) +  "\n" +
```
krisv: Same concern around backwards compatibility, doesn't this also make it difficult to actually try and share data across instances, which might be useful in some cases as well.  Or are we going to consider this an anti-pattern since this will result in different behaviour depending on session strategy?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/SubProcessNodeInstance.java
```diff
@@ -201,6 +204,9 @@ public void internalTrigger(final NodeInstance from, String type) {
                 // remove foreach input variable to avoid problems when running in variable strict mode
                 parameters.remove(getSubProcessNode().getMetaData("MICollectionInput"));
             }
+            if (getVariable(ENTRY_POINT_VAR_NAME) != null) {
```
krisv: This seems to happen at the end, so seems to override any value provided in mapping?  Should we only do this if this param isn't set yet?  Also would suggest just using getVariable(..) once and reus second time for efficiency.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
mswiderski: Same here but couldn't find anything better

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/org/jbpm/casemgmt/demo/insurance/insurance-rules.drl
mswiderski: Yes, it does make it backward incompatible as it introduces complete isolation bwteen the processes

jbpm-flow-builder/src/main/java/org/jbpm/compiler/ProcessBuilderImpl.java
mswiderski: It's only kbase separation as process instance is still mainly based on runtime strategy so I'd say it's an antipattern

mswiderski:
In general my main concern is that this is too much constrained solution as it separates the hard way so to say. I think to close this and put the need of separation on users to create rules in the same kjar to be unique when this is desired. With improvement to include case def id (process id) to be available in CaseData/CaseFile type so it can be easily used in rules to separate them. Wdyt?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
krisv: Alternative would be to make metadata persisted (like it is at definition side).

mswiderski:
closing in favor of https://github.com/droolsjbpm/jbpm/issues/729

PR #726, opened 2017-01-18
JBPM-5591 - Pluggable task assignments

mswiderski:
depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/201

@krisv main work done in here... and description and some details can be found at http://mswiderski.blogspot.com/2017/01/distribute-tasks-wisely-pluggable-task.html


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/AssignmentServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+    @Override
+    public void assignTask(Task task, TaskContext context) {
+        
+        assignTask(task, context, null);
+    }    
+
+    @Override
+    public void assignTask(Task task, TaskContext context, String excludedUser) {
+        if (!isEnabled()) {
+            logger.debug("AssignmentService is not enabled - to enable it set system property 'org.jbpm.task.assignment.enabled' to true");
```
rsynek: The property name should be defined as a constant - it's been already used twice.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/AssignmentServiceImpl.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+    
+    public AssignmentServiceImpl(AssignmentStrategy strategy) {        
+        this.strategy = strategy;
+    }
+    
+    @Override
+    public void assignTask(Task task, TaskContext context) {
+        
+        assignTask(task, context, null);
```
rsynek: Not sure if empty string is an option here, but passing null to some method brings a risk of having NPE somewhere else.

mswiderski: that's actually what is expected, if not checked then it should result in NPE on this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/AssignmentServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+import org.kie.internal.task.api.TaskModelFactory;
+import org.kie.internal.task.api.TaskModelProvider;
+import org.kie.internal.task.api.assignment.Assignment;
+import org.kie.internal.task.api.assignment.AssignmentStrategy;
+import org.kie.internal.task.api.model.InternalTaskData;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class AssignmentServiceImpl implements AssignmentService {
```
livthomas: Unit tests for this class are missing.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LifeCycleLocalCommandBasedTest.java
```diff
@@ -0,0 +1,55 @@
...
+import org.jbpm.services.task.LifeCycleBaseTest;
+import org.junit.After;
+import org.junit.Before;
+import org.kie.internal.task.api.InternalTaskService;
+
+import bitronix.tm.resource.jdbc.PoolingDataSource;
+
+
+
+public class LifeCycleLocalCommandBasedTest extends LifeCycleBaseTest {
```
livthomas: What is the purpose of having this test class here? It seems like it has been copied from `org.jbpm.services.task.LifeCycleLocalCommandBasedTest` without any change.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/AssignmentServiceProvider.java
```diff
@@ -0,0 +1,56 @@
...
+        this.assignmentService = assignmentService;
+        
+    }
+    
+    public static AssignmentService get() {
+        return Holder.INSTANCE.getAssignmentService();
+    }
+    
+    public static AssignmentService get(AssignmentStrategy strategy) {
+        Holder.INSTANCE.setAssignmentService(new AssignmentServiceImpl(strategy));
```
livthomas: Isn't this a little bit counterintuitive? It is `get` method which changes the internal state of the object.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/AssignmentServiceImpl.java
```diff
@@ -0,0 +1,104 @@
...
+    private static final String ENABLED_PROPERTY = "org.jbpm.task.assignment.enabled";
+    
+    private boolean enabled = Boolean.parseBoolean(System.getProperty(ENABLED_PROPERTY, "false"));
+
+    private AssignmentStrategy strategy;
+    private AssignmentServiceRegistry registry = AssignmentServiceRegistry.get();
+    private TaskModelFactory taskModelFactory = TaskModelProvider.getFactory();
+    
+    public AssignmentServiceImpl() {
+        this.strategy = registry.getStrategy(System.getProperty("org.jbpm.task.assignment.strategy", "PotentialOwnerBusyness"));
```
livthomas: Shouldn't the strategy identifier be referenced by its `IDENTIFIER` constant (which might be made public)?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/AssignmentService.java
```diff
@@ -0,0 +1,31 @@
...
+import org.kie.api.task.TaskContext;
+import org.kie.api.task.model.Task;
+
+public interface AssignmentService {
+    
+    boolean isEnabled();
+
+    void assignTask(Task task, TaskContext context);
+    
+    void assignTask(Task task, TaskContext context, String excludedUser);
```
mswiderski: these methods serve different use cases so I'd like to keep them separated on the api level.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/AssignmentServiceImpl.java
mswiderski: this class is tested via TaskService for different methods that should invoke the service. It would be rather complex to come up with all plumbing without the task service and is not meant to be used as standalone any way

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/AssignmentServiceImpl.java
mswiderski: very good point

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LifeCycleLocalCommandBasedTest.java
mswiderski: a copy paste used to test the all tests build, will remove it, good catch

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/AssignmentServiceProvider.java
mswiderski: that is mainly for the testing purpose so I'll make this a protected/package level and rename to override instead, thanks

mswiderski:
@livthomas updates done

mswiderski:
Jenkins retest this

krisv:
Overall +1

Two open questions (can be addressed later):
- is it sufficient to make this a system property?  this will mean all kjars will be using the same rules (this means I guess that either assignment strategy should be the same or the rules would have to make the distinction somehow).  Wondering if a more project-based configuration (deployment descriptor) might be possible as well, although I realize this might not be trivial.. Maybe add some additional logic where assignment service can cache multiple kbases, pluggable strategy to select a kbase based on deployment id?
- I think we should make some more out-of-the-box strategies that people can use, like round-robin, geography or skill-based, etc.

Wdyt?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/PotentialOwnerBusynessAssignmentStrategy.java
```diff
@@ -0,0 +1,140 @@
...
+        List<OrganizationalEntity> potentialOwners = new ArrayList<>(task.getPeopleAssignments().getPotentialOwners());
+        Set<String> resolvedUsers = new TreeSet<>(Collections.reverseOrder());
+        
+        potentialOwners.stream().filter(po -> po instanceof User).forEach(po -> resolvedUsers.add(po.getId()));
+       
+        UserInfo userInfo = (UserInfo) ((org.jbpm.services.task.commands.TaskContext)context).get(EnvironmentName.TASK_USER_INFO);
+        if (userInfo != null) {
+            logger.debug("Groups going to be resolved by {}", userInfo);
+            potentialOwners.stream().filter(po -> po instanceof Group).forEach(po -> {
+                Iterator<OrganizationalEntity> usersOfGroup = userInfo.getMembersForGroup((Group)po);
```
mswiderski: No that I am aware of

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/BusinessRuleAssignmentStrategy.java
```diff
@@ -0,0 +1,126 @@
...
+                QueryResults results = kieSession.getQueryResults(queryName);
+                results.forEach(row -> 
+                    assignments.add((Assignment)row.get("assignment")));
+            } else {
+                logger.debug("No query defined, retrieving all facts of type Assignment");
+                Collection<Assignment> results = (Collection<Assignment>) kieSession.getObjects(new ClassObjectFilter(Assignment.class));
+                
+                assignments.addAll(results);
+            }
+            logger.debug("Rule evaluation completed with selected assignemtns of {}", assignments);
```
mswiderski: Will fix

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/BusinessRuleAssignmentStrategy.java
```diff
@@ -0,0 +1,126 @@
...
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Assignment apply(Task task, TaskContext context, String excludedUser) {
+        if (!active) {
+            logger.debug("{} strategy is not active", this);
+            return null;
+        }
+        logger.debug("Using rules to assign actual owner to task {}", task);
+        KieSession kieSession = this.kieContainer.newKieSession();
```
mswiderski: I considered this at the beginning but that would require batch command to be used to isert, fire all reles etc so decided to go with default KieSession instead

mswiderski:
Yes , the strategy by design is global but you have always an option to create strategy that includes other strategies sort of container approach. When it comes to per kjar that can easily be done with rules and filter by deployment that is available on inserted task instance. 

More strategies out of the box - fully agree. We just need to identify which one should be there and start adding them where one should be based on optaplanner as that can later on be used for further extensions. Round robin strategy sounds really similar the the potential owner busyness so not sure that will have much value. geography sounds ok but what should this be based on? Like variable on a task or? If so then best to do this is actually via rules again...

PR #727, opened 2017-01-19
Findbugs

rsynek:
Introducing findbugs and fixing issues the tool found. Rank 6 in the configuration of findbugs maven plugin seems to provide good results in terms of no false positives have been found.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,7 @@
     <maven.jdbc.driver.class>org.h2.Driver</maven.jdbc.driver.class>
     <maven.jdbc.db.name/>
     <maven.jdbc.db.port/>
-    <maven.jdbc.db.server/>     
+    <maven.jdbc.db.server/>
     <maven.jdbc.driver.jar/>
     <maven.jdbc.username>sa</maven.jdbc.username>
     <maven.jdbc.password>sasa</maven.jdbc.password>
@@ -88,7 +88,7 @@
   </modules>
 
   <build>
-     <pluginManagement>
+    <pluginManagement>
       <plugins>
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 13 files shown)


pom.xml
```diff
@@ -98,9 +98,33 @@
               <additionalClasspathElement>${maven.jdbc.driver.jar}</additionalClasspathElement>
             </additionalClasspathElements>
           </configuration>
-        </plugin> 
+        </plugin>
+        <plugin>
+          <groupId>org.codehaus.mojo</groupId>
```
triceo: My $0.02: this configuration should be centralized across the entire KIE Group. Individual projects would then simply opt-in by enabling it. This will make it much easier to propagate this change in the future, and make the enforcement rules consistent across the entire codebase.

rsynek: That's a step number two in my opinion - I am not sure if this particular setting, which seems to be completely fine for jBPM, will be equally suitable for all other projects. Once we go over more repositories in KIE Group, we will have better overview and for sure this piece of configuration will end up in kie-parent.

rsynek:
Jenkins retest this. Unable to see this failing locally, let's give it a second try.

mswiderski:
@rsynek could you please rebase with master as this is now with conflict. Plus squash all commits into single one as I don't see much value of having them separated

rsynek:
@mswiderski done, tests seem to be ok now.

psiroky:
Thanks @rsynek!

One thing to consider: FindBugs considerably slows down the build. Do we want to keep it enabled by default? So far the Jenkins jobs were executing the checks on demand via `findbugs:findbugs`. We can still do that for PR builds, but locally people need `mvn clean install findbugs:findbugs` instead of just `mvn clean install` to get the results.

Just for comparison (on my machine):
* `mvn clean install -T4 -DskipTests` takes ~4 minutes
* `mvn clean install -T4 -DskipTests -Dfindbugs.skip=true` takes ~ 1min 30secs

PR #729, opened 2017-01-23
JBPM-5541 - Case file conditional event fails other cases

mswiderski:
depends on https://github.com/droolsjbpm/droolsjbpm-knowledge/pull/202


mswiderski:
@krisv this is the alternative which I opt at merging rather than https://github.com/droolsjbpm/jbpm/pull/724, wdyt?

krisv:
+1, developer has some responsibility here, we can request feedback on how to simplify this even further, but forcing the separation at engine level (as the previous PR was doing) doesn't seem ideal as it seems to introduce issues as well

mswiderski:
depends on https://github.com/droolsjbpm/drools/pull/1072

mswiderski:
jenkins retest this

PR #734, opened 2017-01-27
JBPM-4646 - Case management - case instance security

mswiderski:


mswiderski:
jenkins retest this

mswiderski:
@krisv if you have some time. This introduces default security for case instances that only involved actors (assigned as group or individual) will be able to perform operations on a case instance. In addition this allows to configure what roles can interact with case instances - by default only chance and destroy case operations are restricted to:
- owner
- admin

roles which are later on equipped with actual assignment per case instance. Any role can be defined for the operations though this relies on case role being set in case definition

mswiderski:
@cristianonicolai @nmirasch feel free to review add your comments or questions as well

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/audit/CaseInstanceAuditEventListener.java
```diff
@@ -0,0 +1,270 @@
...
+import org.jbpm.shared.services.impl.commands.PersistObjectCommand;
+import org.jbpm.shared.services.impl.commands.QueryStringCommand;
+import org.jbpm.shared.services.impl.commands.UpdateStringCommand;
+import org.kie.internal.runtime.Cacheable;
+import org.kie.internal.task.api.TaskModelProvider;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+public class CaseInstanceAuditEventListener implements CaseEventListener, Cacheable {
```
cristianonicolai: @mswiderski, perhaps you could create a default listener to implement all methods and only override the ones you need here.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/audit/CaseRoleAssignmentLog.java
```diff
@@ -0,0 +1,170 @@
...
+
+import org.kie.api.task.model.Group;
+import org.kie.api.task.model.OrganizationalEntity;
+import org.kie.api.task.model.User;
+
+@Entity
+@SequenceGenerator(name="caseRoleAssignLogIdSeq", sequenceName="CASE_ROLE_ASSIGN_LOG_ID_SEQ", allocationSize=1)
+public class CaseRoleAssignmentLog implements Serializable {
+    	
+	private static final long serialVersionUID = 5757703502285633210L;
```
cristianonicolai: some minor formatting issues in here...

cristianonicolai:
@mswiderski looks good to me, just some minor comments.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/audit/CaseInstanceAuditEventListener.java
mswiderski: I was actually thinking about this and maybe better would be to actually create all the methods in the interfaces as default instead, wdyt?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/audit/CaseRoleAssignmentLog.java
mswiderski: will fix, thanks

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/audit/CaseInstanceAuditEventListener.java
cristianonicolai: yeah, that would be even better :)

PR #740, opened 2017-02-07
JBPM-5657: Added testTransitionBetweenStagesInCase

ibek:
Reproducer for https://issues.jboss.org/browse/JBPM-5657

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImplTest.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImplTest.java
@@ -38,6 +38,8 @@ import org.jbpm.casemgmt.api.model.CaseRole;
 import org.jbpm.casemgmt.api.model.CaseStage;
 import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
 import org.jbpm.casemgmt.api.model.instance.CaseInstance;
+import org.jbpm.casemgmt.api.model.instance.CaseStageInstance;
+import org.jbpm.casemgmt.api.model.instance.StageStatus;
 import org.jbpm.casemgmt.impl.util.AbstractCaseServicesBaseTest;
 import org.jbpm.kie.services.impl.KModuleDeploymentUnit;
 import org.jbpm.services.api.model.DeploymentUnit;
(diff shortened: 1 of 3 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImplTest.java
```diff
@@ -807,4 +810,68 @@ public void testUserTasksInCaseAdStakeholder() {
...
+            assertEquals("Task 1", adhocTasks.iterator().next().getName());
+            
+            Collection<NodeInstanceDesc> activeNodes = caseRuntimeDataService.getActiveNodesForCase(caseId, new QueryContext(0, 10));
+            assertNotNull(activeNodes);
+            assertEquals(1, activeNodes.size());
+            assertEquals("Stage One", activeNodes.iterator().next().getName());
+            
+            caseService.addDataToCaseFile(caseId, "customData", "nextStagePlease");
+            
+            stage = caseRuntimeDataService.getCaseInstanceStages(caseId, true, new QueryContext(0, 1));
```
mswiderski: @ibek the problem here is that you narrow it down to only one item and thus it fetches to little data from db so it cannot properly filter the data. Is this intentional:
```
new QueryContext(0, 1)
```

once this is changed to 
```
new QueryContext(0, 10)
```

all works as expected and the test pass.

ibek: I cannot check it right now - I will look at it tomorrow ... however even though the range is extended . new QueryContext(0, 10) - the method shouldn't return both stages as active if it's the case why it passes.... the reason I use new QueryContext(0, 1) is that I want to get only one stage because I expect only one stage to be active.

ibek: @mswiderski It is exactly how you describe it with only one result which makes the test completely pass ... quite strange - couldn't the filtering happen in the db before it fetches the data?

mswiderski: its actually fails because paging is applied on db and thus returns first found row. That is because NodeInstanceLog has two rows per each node instance - entry and exit thus filtering on db is expensive. So taking into consideration stages there won't be many in given case - like not 10000 etc then better is to do filtering in the code. The fix for this is here https://github.com/droolsjbpm/jbpm/pull/741 including your test case so I will close this PR.

mswiderski:
replaced by https://github.com/droolsjbpm/jbpm/pull/741

PR #743, opened 2017-02-09
JBPM-5667: Second stage of a Case verifies its completion condition w…

ibek:
…ith old value of case file property


mswiderski:
test case incorporated in https://github.com/droolsjbpm/jbpm/pull/745, thanks @ibek

PR #745, opened 2017-02-09
JBPM-4646 - Case management - added support for taking completed node, JBPM-5667 - Second stage of a Case verifies its completion condition

mswiderski:


ibek:
wonderful, thank you @mswiderski 
+1 for merge

PR #747, opened 2017-02-13
Fix container test suite dependencies

rsynek:


psiroky:
Looks good. Is this the only dependency we need to add? There are more of these `idep` modules https://github.com/droolsjbpm/jbpm/tree/master/jbpm-container-test/jbpm-in-container-test/jbpm-container-integration-deps ( I don't really know these tests, so it may be that we really need just the one, but better to be sure)

rsynek:
Adding other modules the jbpm-container-test-suite depends on. As the jbpm-remote-ejb-test-suite already defines dependencies on KJAR modules, it's instability might be caused by runtime scope of these dependencies (parallel build might decide to postpone them). Changing them to compile scope.

mswiderski:
looks good to me, is this ready to be merged?

PR #750, opened 2017-02-16
JBPM-5692 - Addition of sub process ID to /nodes endpoint for SubProc…

mswiderski:
…essNode, JBPM-5703 - CaseStageInstace doesn't return any active nodes

@ibek as we discussed active nodes are now part of the stage
@tsurdilo here is the jbpm commit for image links

reason why these two jiras are in single PR/commit is the changes on same db table/entity to reduce changes to same parts of code


ibek:
reviewed, +1 for merge

PR #759, opened 2017-02-22
DROOLS-1446 - ruleunit implementation - jbpm adjustments

mswiderski:
@mariofusco @tarilabs could you please take a look at this guys? I don't know if I got the adjustments right...


psiroky:
+1 to merge when build is green. This is needed to avoid errors on downstream repos, so I am in favor of merging this even without a review from @mariofusco. Adjustments can be done later. Having a working build should be a priority.

PR #762, opened 2017-02-24
RHBPMS-4622 - LogCleanupCommand executor command should not clean up, RHBPMS-4543 - Error during async work item handler complete may wrong

mswiderski:


mswiderski:
thanks @lanceleverich for review

to provide clarification:
for LogCleanupCommand the change was to always include subquery that checks the process instance log table to only allow deletion of data belonging to not active process instances.
for the executor fix it is now separated the command execution from singling the engine. If the engine failed after the signal, job is marked as done but put back to retrying mode, though when the job fires again executor will check if the job has already been executed - based on results being already there. If so it will only attempt to signal the engine again instead of calling command.

PR #765, opened 2017-02-27
JBPM-5727 - SQLGrammarException is thrown when getTasksAssignedAsPote…

mswiderski:
…ntialOwner() method is called

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
@@ -16,14 +16,14 @@
 
 package org.jbpm.kie.services.impl;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.kie.api.command.Command;
 import org.kie.internal.command.ProcessInstanceIdCommand;
+import org.kie.internal.identity.IdentityProvider;
 
-/**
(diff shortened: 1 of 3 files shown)


mswiderski:
@MarianMacik here is the compensation for the issue with found

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
```diff
@@ -42,4 +42,16 @@ public static Long getProcessInstanceId(Command<?> command) {
         return null;
 		
 	}
+	
+	// to compensate https://hibernate.atlassian.net/browse/HHH-8091 add empty element to the roles in case it's empty
+	public static List<String> getAuthenticatedUserRoles(IdentityProvider identityProvider) {
+        List<String> roles = identityProvider.getRoles();
```
lanceleverich: There should be a null check on the identityProvider, prior to calling its getRoles method. Maybe something like...
`List<String> roles = (identityProvider != null) ? identityProvider.getRoles():null;`

mswiderski: in general the identityProvider should not be null in real deployments but I agree to apply the change - with small update - if it is null return empty list instead of null. thanks @lanceleverich

mswiderski:
change applied

PR #767, opened 2017-03-02
[6.5.x] RHBPMS-4638 - SQLGrammarException when retrieving user tasks

mswiderski:
@sutaakar do you mind verifying this on 6.5 bits before merging?

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
@@ -16,30 +16,42 @@
 
 package org.jbpm.kie.services.impl;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import org.kie.api.command.Command;
 import org.kie.internal.command.ProcessInstanceIdCommand;
+import org.kie.internal.identity.IdentityProvider;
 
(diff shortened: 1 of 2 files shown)


mswiderski:
jenkins retest this

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/CommonUtils.java
```diff
@@ -42,4 +42,16 @@ public static Long getProcessInstanceId(Command<?> command) {
         return null;
 		
 	}
+	
+	// to compensate https://hibernate.atlassian.net/browse/HHH-8091 add empty element to the roles in case it's empty
+	public static List<String> getAuthenticatedUserRoles(IdentityProvider identityProvider) {
+        List<String> roles = identityProvider != null ? identityProvider.getRoles() : new ArrayList<>();
```
sutaakar: Seems to be a compilation issue here, new ArrayList<>() should be new ArrayList<String>()

mswiderski:
should be fixed now. not sure why I got jenkins report saying there is only test failure... or I read it wrong. thanks @sutaakar

mswiderski:
jenkins retest this

sutaakar:
@mswiderski Yes, it fix the issue.

PR #772, opened 2017-03-07
distribution: remove no longer needed installer assembly

psiroky:
The jbpm-installer is properly mavenized, so there is no need
to keep this one around.

@krisv could you please take a look? Is this what you had in mind?


krisv:
+1

PR #773, opened 2017-03-08
JBPM-5441 - When a timer is triggered the user unknown does not have …

mswiderski:
…permission to access asset which requires a specific role

@ibek if you find a minute...


ibek:
looks great, thank you for fixing it Maciej

so generally every async activity bypass the security right? Since it's always planned by model (not requested by any user) it should be alright to do so.

mswiderski:
that is correct, async operations will bypass security check if user info can't be retrieved, that's why the check is at the end of the security verification.

PR #790, opened 2017-03-24
Round robin assignment strategy [JBPM-5841]

lanceleverich:
Adds round-robin assignment strategy to the growing set of human task auto-assign strategies.


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,118 @@
...
+        OrganizationalEntity owner = mappedQueue.take();
+        return new Assignment(owner.getId());
+    }
+    
+    protected boolean queueContainsUser(CircularQueue<OrganizationalEntity> queue, OrganizationalEntity oe) {
+    	return queue.contains(oe);
+    }
+
+    protected String getQueueName(Task task) {
+        return task.getTaskData().getDeploymentId()+"_"+task.getName();
```
mswiderski: I'd recommend to add process id to the name as well as there might be tasks with same name in different processes but still in the same kjar

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,118 @@
...
+        return IDENTIFIER;
+    }
+
+    @Override
+    public Assignment apply(Task task, TaskContext taskContext, String s) {
+        List<OrganizationalEntity> excluded = ((InternalPeopleAssignments)task.getPeopleAssignments()).getExcludedOwners();
+        UserInfo userInfo = (UserInfo) ((org.jbpm.services.task.commands.TaskContext)taskContext).get(EnvironmentName.TASK_USER_INFO);
+
+        // Get the the users from the task's the potential owners
+        List<OrganizationalEntity> potentialOwners = task.getPeopleAssignments().getPotentialOwners().parallelStream()
```
mswiderski: List of potential owners is usually small so what would be the advatage of using parallelStream?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,118 @@
...
+                .filter(oe -> oe instanceof User && !excluded.contains(oe))
+                .collect(Collectors.toList());
+
+        // Get the users belonging to groups that are potential owners
+        task.getPeopleAssignments().getPotentialOwners().parallelStream().filter(oe -> oe instanceof Group)
```
mswiderski: Same here, parallelStream would most likely be slower than regular one for such small list

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,118 @@
...
+            potentialOwners.forEach(po -> {
+            	if (!queueContainsUser(queue,po)) {
+            		queue.add(po);
+            	}
+            });
+            queue.removeIf(oe -> !potentialOwners.contains(oe));
+            mappedQueue = queue;
+        } else {
+            CircularQueue<OrganizationalEntity> queue = new CircularQueue();
+            potentialOwners.forEach(po -> {queue.add(po);});
```
mswiderski: Can't we use queue.addAll here?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,118 @@
...
+        // If a queue already exists for this task then its contents should be synchronized with the
+        // current list of potential owners
+        if (circularQueueMap.containsKey(queueName)) {
+            final CircularQueue<OrganizationalEntity> queue = circularQueueMap.get(queueName); 
+            potentialOwners.forEach(po -> {
+            	if (!queueContainsUser(queue,po)) {
+            		queue.add(po);
+            	}
+            });
+            queue.removeIf(oe -> !potentialOwners.contains(oe));
```
mswiderski: Are the above queue modifications thread safe? Asking as this will be executed by multiple threads in parallel as the strategy is shared for all incoming requests. What would happen when there are two concurrent requests fir the same task but they have different potential owner lists?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private class CircularQueue<T> extends LinkedBlockingQueue<T> {
+        @Override
+        public T take() {
+            T headValue = null;
+            try {
+                headValue = super.take();
+                super.offer(headValue);
+            } catch (InterruptedException e) {
+                e.printStackTrace();
```
mswiderski: Maybe use logger instead system.err

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/RoundRobinAssignmentTests.java
```diff
@@ -0,0 +1,137 @@
...
+        this.taskService = (InternalTaskService) HumanTaskServiceFactory.newTaskServiceConfigurator()
+                .entityManagerFactory(emf)
+                .getTaskService();
+
+    }
+
+    @After
+    public void clean() {
+        System.clearProperty("org.jbpm.task.assignment.enabled");
+        AssignmentServiceProvider.clear();
```
mswiderski: Maybe clear the other system property as well that is set in setup method

mswiderski:
Ok to test

mswiderski:
@lanceleverich nice to see this pull request, added some comments to clarify

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
domhanak: +1 for logger usage

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
domhanak: Maybe use sequential stream, this adds another unnecessary layer of parallelism.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
domhanak: Maybe use sequential stream, this adds another unnecessary layer of parallelism.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
domhanak: +1

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/RoundRobinAssignmentStrategy.java
```diff
@@ -0,0 +1,130 @@
...
+                                potentialOwners.add(user);
+                            }
+                        });
+                    }
+                });
+        String queueName = getQueueName(task);
+        CircularQueue<OrganizationalEntity> mappedQueue = synchronizedQueue(circularQueueMap.get(queueName),potentialOwners);
+        // if the queue did not already exist then it will be added to the map
+        // otherwise the queue that was in the map will be updated
+        circularQueueMap.put(queueName,mappedQueue);
```
mswiderski: I'd recommend to move this into the synchronizedQueue method to avoid potential race condition in case multiple threads would come in to get the queue as they might get new queues from synchronizedQUeue method and then replace each other on this line

mswiderski:
@lanceleverich there is only one last comment to fix and this will be ready to be merged too.

PR #791, opened 2017-03-27
JBPM-5853 added null checking for the itemSubjectRef during TaskHandl…

marianbuenosayres:
…er reading. Defaulting to 'Object' data type in case no type is found. This is added due to the same BPMN file (BPMN2-BrokenStructureRef.bpmn2) not throwing any errors in the jBPM 6 version.


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/BrokenStructureRefTest.java
```diff
@@ -0,0 +1,27 @@
+package org.jbpm.bpmn2;
```
mswiderski: @marianbuenosayres could you please add license header?

PR #792, opened 2017-03-28
JBPM-5816 - Allow changing case owner

cristianonicolai:
Ensure that all case instances have a case role even if done defined in the model.
 Updated queries to retrieve owner from assignment instead of using audit change log user.

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
@@ -73,13 +73,13 @@ import org.kie.api.KieServices;
 import org.kie.api.command.KieCommands;
 import org.kie.api.runtime.manager.RuntimeManager;
 import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.api.runtime.query.QueryContext;
 import org.kie.api.task.model.Group;
 import org.kie.api.task.model.OrganizationalEntity;
 import org.kie.api.task.model.User;
 import org.kie.internal.KieInternalServices;
 import org.kie.internal.identity.IdentityProvider;
 import org.kie.internal.process.CorrelationKeyFactory;
-import org.kie.api.runtime.query.QueryContext;
 import org.kie.internal.runtime.manager.context.CaseContext;
 import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
 import org.kie.internal.task.api.TaskModelFactory;
@@ -164,9 +164,13 @@ public class CaseServiceImpl implements CaseService {
             ((CaseFileInstanceImpl)caseFile).setCaseId(caseId);
             logger.debug("CaseFile {} was given, associating it with case {}", caseFile, caseId);
         }
-        
-        ((CaseFileInstanceImpl)caseFile).assignOwner(newUser(identityProvider.getName()));
-        
+
+        //If owner is provided in the case file use that, otherwise default to current logged in user.
+        boolean hasOwner = ((CaseFileInstanceImpl) caseFile).getAssignments().stream().anyMatch(role -> role.getRoleName().equals(AuthorizationManager.OWNER_ROLE));
+        if(hasOwner == false){
+            ((CaseFileInstanceImpl)caseFile).assignOwner(newUser(identityProvider.getName()));
+        }
+
         processService.execute(deploymentId, CaseContext.get(caseId), new StartCaseCommand(caseId, deploymentId, caseDefinitionId, caseFile, processService));
         
         return caseId;
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
@@ -16,8 +16,6 @@
 
 package org.jbpm.casemgmt.impl.model.instance;
 
-import static java.util.stream.Collectors.toMap;
-
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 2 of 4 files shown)


cristianonicolai:
@mswiderski please take a look if this seems reasonable.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
```diff
@@ -164,9 +164,13 @@ public String startCase(String deploymentId, String caseDefinitionId, CaseFileIn
             ((CaseFileInstanceImpl)caseFile).setCaseId(caseId);
             logger.debug("CaseFile {} was given, associating it with case {}", caseFile, caseId);
         }
-        
-        ((CaseFileInstanceImpl)caseFile).assignOwner(newUser(identityProvider.getName()));
-        
+
+        //If owner is provided in the case file use that, otherwise default to current logged in user.
+        boolean hasOwner = ((CaseFileInstanceImpl) caseFile).getAssignments().stream().anyMatch(role -> role.getRoleName().equals(AuthorizationManager.OWNER_ROLE));
```
mswiderski: any reason why this is here and not in CaseFileInstance.assignOwner?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
```diff
@@ -218,11 +220,12 @@ public void setupRoles(Collection<CaseRole> roles) {
...
-        boolean hasDefinedOwner = roles.containsKey(AuthorizationManager.OWNER_ROLE);
-        if (!hasDefinedOwner && !roles.isEmpty()) {
-            this.roles.put(AuthorizationManager.OWNER_ROLE, new CaseRoleInstanceImpl(AuthorizationManager.OWNER_ROLE, 1));
-            assign(AuthorizationManager.OWNER_ROLE, actualOwner);
+        CaseRoleInstance caseRoleInstance = this.roles.get(AuthorizationManager.OWNER_ROLE);
+        if (caseRoleInstance == null) {
+            caseRoleInstance = new CaseRoleInstanceImpl(AuthorizationManager.OWNER_ROLE, 1);
+            this.roles.put(AuthorizationManager.OWNER_ROLE, caseRoleInstance);
         }
+        ((CaseRoleInstanceImpl)caseRoleInstance).addRoleAssignment(actualOwner);
```
mswiderski: this might throw exception of there are already owner assigned, is that what we want?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
```diff
@@ -14,8 +14,8 @@
         log.processInstanceDescription, <!-- case description -->            
         log.processId, <!--case definition  -->
         log.status, <!-- case status -->
-        log.externalId, <!-- deployment id -->      
-        log.identity, <!-- case owner -->
+        log.externalId, <!-- deployment id -->
+        (select r.entityId from CaseRoleAssignmentLog r where r.processInstanceId = log.processInstanceId and r.roleName = 'owner' and r.type = 1), <!-- case owner -->
```
mswiderski: I believe this will only work for case definitions that have owner defined or any other role defined as then the owner is added automatically. But if case definition is not using roles at all that will be then empty, isn't it?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -1916,4 +1917,134 @@ public void testTriggerNotExistingAdHocFragment() {
         }
     }
 
+    @Test
+    public void testStartEmptyCaseUsingLoggedInOwner() {
+        assertNotNull(deploymentService);
+        DeploymentUnit deploymentUnit = new KModuleDeploymentUnit(GROUP_ID, ARTIFACT_ID, VERSION);
+
+        deploymentService.deploy(deploymentUnit);
+        units.add(deploymentUnit);
```
rsynek: Adding a deployment unit is a step of preparing a test environment, thus is should be called from a `@Before` method - that will reduce duplication, as every test method in this PR starts with this block of code.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
cristianonicolai: @mswiderski the idea is that owner role is always added to the case instance, independently if it was defined in the model.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
cristianonicolai: Yep, like any other role assignment, if you define more users/groups than the allowed cardinality and exception is thrown.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
cristianonicolai: This is only checking if there is a owner already set in case file when starting the case instance. I guess this is only relevant in here whereas the assignOwner is used every time there is an assignment.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
cristianonicolai: @rsynek done

rsynek: @cristianonicolai looks good, thanks!

PR #794, opened 2017-04-01
Container test fixes

psiroky:
@rsynek could you please review? This should fix the test failures seen in the nightly builds.


rsynek:
+1, thanks for fixing it

PR #795, opened 2017-04-03
Run container tests only when container profile is used

psiroky:
@MarianMacik could you please take a look?


MarianMacik:
Everything seems to be OK. Good job spotting this! Thanks!

PR #796, opened 2017-04-03
JBPM-5468 - Enable multiple column sorting on advanced queries

markcoble:
Enhancing the advanced queries provided by REST and JMS to support a full SQL ORDER BY clause.  This ensures the sort order can be selected per column of ordering.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
@@ -57,6 +57,7 @@ import org.jbpm.services.api.query.model.QueryParam;
 import org.jbpm.shared.services.impl.QueryManager;
 import org.jbpm.shared.services.impl.TransactionalCommandService;
 import org.jbpm.shared.services.impl.commands.QueryNameCommand;
+import org.kie.api.runtime.query.AdvancedQueryContext;
 import org.kie.api.runtime.query.QueryContext;
 import org.kie.internal.identity.IdentityProvider;
 import org.slf4j.Logger;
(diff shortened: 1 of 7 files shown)


jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/query/QueryService.java
```diff
@@ -79,7 +80,7 @@
      * @return results mapped to objects from raw data set
      * @throws QueryNotFoundException in case there is no such query registered
      */
-    <T> T query(String queryName, QueryResultMapper<T> mapper, QueryContext queryContext, QueryParam...filterParams) throws QueryNotFoundException;
+    <T> T query(String queryName, QueryResultMapper<T> mapper, AdvancedQueryContext queryContext, QueryParam...filterParams) throws QueryNotFoundException;
```
mswiderski: I'd prefer not to change the API here as the AdvancedQueryContext will only be used when there is a need to pass the sql like oder by clause which might not be so often. In all other cases it would be enough to rely on default QueryContext and by that we'll not break backward compatibility. Moreover this PR will have far less changes :)

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
```diff
@@ -232,20 +223,24 @@ public void unregisterQuery(final String uniqueQueryName) throws QueryNotFoundEx
...
         }
-        
-        if (queryContext.getOrderBy() != null) {
-            String[] oderByItems = queryContext.getOrderBy().split(",");
-            
-            for (String orderBy : oderByItems) {
-                logger.debug("Applying order by {} and ascending {}", orderBy, queryContext.isAscending());
-                builder.sort(orderBy.trim(), queryContext.isAscending()?"asc":"desc");
+
+        // process the ORDER BY clause into order by and sort order pairs
```
mswiderski: if we will use the QueryContext as method signature then we could just add a check here to handle orderByClause in the actual impl is AdvancedQueryContext

tsurdilo:
Can we set the pr title to include the Jira title, for example:
"JBPM-5468 - Enable multiple column sorting on advanced queries"

PR #797, opened 2017-04-04
RHBPMS-1328 - [RFE] business rule task supports boundary error event

tsurdilo:


jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -65,83 +75,112 @@
...
+
+                    removeEventListeners();
+                    retractFacts();
+                    triggerCompleted();
+                }
+            }
+        } catch (Exception e) {
+            if(!handleException(e, false)) {
+                throw new WorkflowRuntimeException(this, getProcessInstance(), "Unable to execute Action: " + e.getMessage(), e);
+            }
```
mswiderski: can we move this logic to handlerException and in case it's not handled simply throw exception?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -65,83 +75,112 @@
...
+        if(!handled && ExceptionUtils.getRootCause(e) != null) {
+            return handleException(ExceptionUtils.getRootCause(e), handled);
+        }
+        return handled;
+    }
+
+    private ExceptionScopeInstance getExceptionScopeInstance(Throwable e) {
+        String exceptionName = e.getClass().getName();
+        Context exceptionContext = getContextContainer().getDefaultContext(ExceptionScope.EXCEPTION_SCOPE).resolveContext(exceptionName);
+        return exceptionContext == null ? null : (ExceptionScopeInstance) getContextInstance(exceptionContext);
```
mswiderski: any reason why we don't use the same approach as for other node instances:

```
String exceptionName = e.getClass().getName();
ExceptionScopeInstance exceptionScopeInstance = (ExceptionScopeInstance) resolveContextInstance(ExceptionScope.EXCEPTION_SCOPE, exceptionName);
```

tsurdilo: unless structureRef is used for the error node in bpmn2 resolveContextInstance in this case will always be null. I can update jBPM Designer to set the structureRef attribute (it does not do it atm) and we would also need to make this change in Stunner once boundary events are enabled there. This approach would work with no changes to the editor so I used that. Let me know what you think.

mswiderski: this method should be responsible for dealing with that: https://github.com/kiegroup/jbpm/blob/master/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java#L420 meaning if there is no structureRef given it should fallback to the default error handler. We need to investigate that as I'd prefer to have the backend processing aligned for all node instance types.

tsurdilo:
@mswiderski updated - thanks.

tsurdilo:
@mswiderski fixed the context resolution and updated tests. context resolution works same across board now. thanks!

mswiderski:
nice work @tsurdilo, thanks!

PR #800, opened 2017-04-06
JBPM-5869 - Boundary Event Signal does not transalate variables while…

tsurdilo:
… registring as Signal Event Listener


tsurdilo:
jenkins please retest this

PR #801, opened 2017-04-07
JBPM-5610: Not able to exclude multiple owners for HumanTask - Tests

domhanak:
**Add tests with groups and excluded owners**

Adds 2 tests methods that test behavior of excluding user/users
from group which is set as potential owner of the task.


livthomas:
I do not think it is a good idea to reference our internal JIRA in the commit message.

mswiderski:
fully agree with @livthomas

domhanak:
@livthomas @mswiderski Yes, I am sorry for that, reworded the commit message.

domhanak:
Should I change the PRQ name too?

mswiderski:
thanks @domhanak

PR #803, opened 2017-04-10
Avoid NPE while using document.toString();

ngs-mtech:
[JBoss JIRA] (JBPM-5878) DocumentImpl toString() generates NPE when no modification date is available


mswiderski:
do you mind adding a simple test case to make sure it will not popup again?

PR #804, opened 2017-04-11
6.4.x Solves (JBPM-5878) DocumentImpl toString() generates NPE when no modification date is available

ngs-mtech:
[JBoss JIRA] (JBPM-5878) DocumentImpl toString() generates NPE when no modification date is available

Including a poor man test to avoid reiteration.


mswiderski:
could you please move this to master branch instead. 6.4.x branch is not under development any more so fixes can go into master only.

PR #806, opened 2017-04-11
JBPM-5880 - 'async' as a default option for all tasks

mswiderski:
@krisv here is the option to turn all nodes into async mode via deployment descriptor. This means we can have it globally by using server level deployment descriptor or per kjar deployed.

in deployment descriptor it needs to be given as environment entry named "AsyncMode" and the value is "true" or "false"


mswiderski:
jenkins retest this

krisv:
Looks good.  Are we going to add it to the deployment descriptor screen as well?

mswiderski:
Didn't plan it as you can already add environment entries via the editor

PR #809, opened 2017-04-13
JBPM-5886: Replace joda-time with Java 8 time implementation

sutaakar:


mswiderski:
@sutaakar looks like compilation errors related tot his change, do you mind taking a look?

sutaakar:
Sure, should be ok, I hope.

sutaakar:
@mswiderski I added there several checks for numeric values, please look on them. I think they could cover all cases where users could use milliseconds as values.

mswiderski:
jenkins retest this

mswiderski:
@sutaakar looks good. Just one question - why do we need these checks? Is this improvement on top of what was there or this was covered somehow differently previously?

mswiderski:
thanks @sutaakar, do we need this for 7.0.x branch as well? I guess that will help with WebLogic deployments, right?

sutaakar:
@mswiderski Going to add test coverage for this too. I think we can cherrypick it to 7.0.x too if it is still opened for PRs.

mswiderski:
I'd say it's worth adding to 7.0 as it might be a blocker for WLS deployments of Kie Server, isn't that the case? if it is not then we can skip the back port

sutaakar:
Yes, it would be a blocker for WLS in 7.0.

PR #812, opened 2017-04-13
[7.0.x] JBPM-4646 - case management - case file data audit trail

mswiderski:


mswiderski:
@tsurdilo here is the back port to 7.0x for the case file data audit

PR #813, opened 2017-04-13
[7.0.x] JBPM-5891 - Process validation should not trigger on empty package name

tsurdilo:


tsurdilo:
@psiroky amended with test fix per your find. Will do a new commit for master as it's already been merged there.

PR #814, opened 2017-04-13
JBPM-5891 - Process validation should not trigger on empty package name

tsurdilo:


tsurdilo:
jenkins please retest this

jbpm-flow/src/test/java/org/jbpm/ruleflow/core/validation/RuleFlowProcessValidatorTest.java
```diff
@@ -125,4 +125,27 @@ public void testDynamicNodeValidationInDynamicProcess() throws Exception {
...
+    @Test
+    public void testNoPackageName() throws Exception {
+        RuleFlowProcess process = new RuleFlowProcess();
+        process.setId("org.drools.core.process");
+        process.setName("No Package Name Process");
+        process.setDynamic(true);
+
+        ProcessValidationError[] errors = validator.validateProcess(process);
+        assertNotNull(errors);
+    }
```
psiroky: I am maybe overlooking something, but it seems that these tests are not really testing the fix? They would pass even when the fix would be removed? Should we perhaps assert specific content of the `errors` array?

PR #821, opened 2017-04-19
[6.5.x] Fix Byteman test for IBM Java

MarianMacik:
Hi, @mswiderski,

I noticed that this new Byteman test hasn't been passing with IBM Java, but I have found a way to boot Byteman even with this version of Java. What do you think? Should I create the same workaround for other modules where Byteman is used?

Thanks,

Marian

### DIFF
--- a/jbpm-runtime-manager/pom.xml
+++ b/jbpm-runtime-manager/pom.xml
@@ -204,7 +204,40 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
           <forkMode>always</forkMode>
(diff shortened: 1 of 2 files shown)


MarianMacik:
jenkins retest this

MarianMacik:
jenkins retest this

MarianMacik:
Failed tests are not connected to this change.

MarianMacik:
jenkins retest this

mswiderski:
@MarianMacik thanks, could you please prepare PR for master and 7.0.x so we have it aligned

MarianMacik:
Sure, I will prepare it.

PR #822, opened 2017-04-19
JBPM-5370: Add test cases covering add/removepotential owners with groups

domhanak:
One case covers scenario where we cant to forward task to multiple PO(group)
One case covers adding PO to non-existent task and checks for exception
One case covers Add/RemoveExcludedOwners with groups

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/admin/UserTaskAdminServiceImplTest.java
+++ b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/admin/UserTaskAdminServiceImplTest.java
@@ -26,6 +26,7 @@ import static org.kie.scanner.MavenRepository.getMavenRepository;
 import java.io.File;
 import java.io.FileOutputStream;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
@@ -37,12 +38,14 @@ import org.jbpm.kie.services.test.KModuleDeploymentServiceTest;
 import org.jbpm.kie.test.util.AbstractKieServicesBaseTest;
 import org.jbpm.kie.test.util.CountDownListenerFactory;
 import org.jbpm.services.api.ProcessInstanceNotFoundException;
+import org.jbpm.services.api.TaskNotFoundException;
 import org.jbpm.services.api.admin.TaskNotification;
 import org.jbpm.services.api.admin.TaskReassignment;
 import org.jbpm.services.api.admin.UserTaskAdminService;
 import org.jbpm.services.api.model.DeploymentUnit;
 import org.jbpm.services.task.exception.PermissionDeniedException;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.kie.api.KieServices;
@@ -110,6 +113,7 @@ public class UserTaskAdminServiceImplTest extends AbstractKieServicesBaseTest {
         units.add(deploymentUnit);
         // set user to administrator so it will be allowed to do operations
         identityProvider.setName("Administrator");
+        identityProvider.setRoles(Arrays.asList(""));
     }
 
     @After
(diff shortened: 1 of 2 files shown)


domhanak:
@mswiderski Could you please take a look at this?
_testAddRemoveExcludedOwnersAsGroup()_ is not passing, could confirm it is correct, please?

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/admin/UserTaskAdminServiceImplTest.java
```diff
@@ -161,6 +165,72 @@ public void testAddPotentialOwnersNotBusinessAdmin() {
...
+        assertEquals(0, tasks.size());
+
+        userTaskAdminService.removeExcludedOwners(task.getId(), factory.newGroup("IT"));
+        tasks = runtimeDataService.getTasksAssignedAsPotentialOwner("salaboy", new QueryFilter());
+        assertEquals(1, tasks.size());
+
+        try {
+            userTaskService.claim(task.getId(), "salaboy");
+        } catch (Exception ex) {
+            Assert.fail("salaboy should be able to claim task here. + " +  ex.getMessage());
```
triceo: Any chance we could not hide the stack trace?

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/admin/UserTaskAdminServiceImplTest.java
```diff
@@ -161,6 +165,72 @@ public void testAddPotentialOwnersNotBusinessAdmin() {
...
+        userTaskService.release(task.getId(), "salaboy");
+
+        identityProvider.setRoles(Arrays.asList("IT"));
+        userTaskAdminService.addExcludedOwners(task.getId(),true, factory.newGroup("IT"));
+        tasks = runtimeDataService.getTasksAssignedAsPotentialOwner("salaboy", new QueryFilter());
+        assertEquals(0, tasks.size());
+
+        userTaskAdminService.removeExcludedOwners(task.getId(), factory.newGroup("IT"));
+        tasks = runtimeDataService.getTasksAssignedAsPotentialOwner("salaboy", new QueryFilter());
+        assertEquals(1, tasks.size());
```
triceo: `AssertJ` will make this much nicer, including the failure message - `Assertions.assertThat(tasks).hasSize(1);`
This could be done on many places in this PR.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/admin/UserTaskAdminServiceImplTest.java
```diff
@@ -161,6 +165,72 @@ public void testAddPotentialOwnersNotBusinessAdmin() {
...
+    public void testAddPotentialOwnersToNonExistentTask() {
+        processInstanceId = processService.startProcess(deploymentUnit.getIdentifier(), "org.jbpm.writedocument");
+        assertNotNull(processInstanceId);
+
+        List<TaskSummary> tasks = runtimeDataService.getTasksAssignedAsPotentialOwner("salaboy", new QueryFilter());
+        assertEquals(1, tasks.size());
+        TaskSummary task = tasks.get(0);
+
+        userTaskService.release(task.getId(), "salaboy");
+        userTaskAdminService.addPotentialOwners(15456, false, factory.newUser("john"));
```
triceo: I'd rather we use `Assertions` methods to catch the exception from a specific place, as opposed to catching the exception from anywhere at all using `@Test`.

triceo: Also FYI JUnit 5 no longer supports `@Test(exception = ...)`. :-)

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/admin/UserTaskAdminServiceImplTest.java
```diff
@@ -110,6 +113,7 @@ public void prepare() {
         units.add(deploymentUnit);
         // set user to administrator so it will be allowed to do operations
         identityProvider.setName("Administrator");
+        identityProvider.setRoles(Arrays.asList(""));
```
triceo: `Collections.singletonList("");`

domhanak:
added requested changes

domhanak:
will squash if above changes are approved

mswiderski:
@domhanak that test is actually not correct mainly because of this https://github.com/kiegroup/jbpm/pull/822/files#diff-3b5dd0ea0416ad31bc7230ec703276eaR215 which means you search for tasks that are eligible for salaboy and the IT group. the task you just released is still eligible for user salaboy and that's why it's found.
Moreover exclude owners makes sense for users as excluding groups seems like overkill as why would you assign task to a group that you next want to exclude completely. It serves the purpose of assigning to group but actually removing certain users of that group e.g. those who already made decisions in previous steps.

domhanak:
@mswiderski Thanks for clarification, I will remove it.

mswiderski:
@triceo anything else needed or we can merge this one?

mswiderski:
thanks @domhanak

PR #826, opened 2017-04-19
[6.5.x] RHBPMS-4702 - EJBTimerScheduler.getTimerByName throws NoSuchObjectLoc…

mswiderski:
…alException in WAS

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -125,16 +125,20 @@ public class EJBTimerScheduler {
 	
 	public TimerJobInstance getTimerByName(String jobName) {
 		for (Timer timer : timerService.getTimers()) {
-			Serializable info = timer.getInfo();
-			if (info instanceof EjbTimerJob) {
-				EjbTimerJob job = (EjbTimerJob) info;
-				
-				EjbGlobalJobHandle handle = (EjbGlobalJobHandle) job.getTimerJobInstance().getJobHandle();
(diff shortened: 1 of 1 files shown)


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -125,16 +125,20 @@ public boolean removeJob(JobHandle jobHandle) {
...
+    				EjbTimerJob job = (EjbTimerJob) info;
+    				
+    				EjbGlobalJobHandle handle = (EjbGlobalJobHandle) job.getTimerJobInstance().getJobHandle();
+    				if (handle.getUuid().equals(jobName)) {
+    					logger.debug("Job  {} does match timer and is going to be returned", jobName);
+    					return handle.getTimerJobInstance();
+    				}
+    			}
+		    } catch (NoSuchObjectLocalException e) {
+                logger.debug("Timer info for {} was not found ", timer);
```
lanceleverich: Should this be logged with a level of WARN, instead of DEBUG? Or is this such a common occurrence that the user/administrator doesn't really need to know when it happens?

tsurdilo: I think we could look over all log statements in this class  as part of another jira and fix in master branch and then propagate to other branches if changes are needed. Feel free to raise a Jira for this please but i think here its fine as it mimics master atm.

mswiderski: in general it should not make administrators to be worried and that's why it's in debug as it is then consistent with the other log of sam exception.

PR #829, opened 2017-04-20
[7.0.x] Fix Byteman test for IBM Java

MarianMacik:
Hi, @mswiderski,

this is the same PR as #821, but for 7.0.x.

Thanks,

Marian


MarianMacik:
jenkins retest this

PR #830, opened 2017-04-20
JBPM-5905 - Transaction exception for Singleton container deployment …

mswiderski:
…in WebLogic

@sutaakar do you mind taking a quick look?
I tested this on WebLogic 12.2.1


sutaakar:
looks good, +1

PR #832, opened 2017-04-21
[7.0.x] JBPM-5897 - Fails to start a process with CorrelationKey which is unique but includes correlation properties of an existing correlation key

mswiderski:
@tsurdilo here is a back port from master to 7.0.x


mswiderski:
Jenkins retest this

PR #833, opened 2017-04-24
Jbpm 5842

lanceleverich:
Added LoadBalanceAssignmentStrategy with the first load calculator (TaskCountLoadCalculator)


jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LoadBalanceAssignmentStrategyTest.java
```diff
@@ -0,0 +1,248 @@
...
+        		+ "name = 'MultiUserWithRemoveLoadBalanceTask2'})";
+		
+        tasks.put("Bobba Fet",createForCompletionTask(taskString, "Bobba Fet", 4, "Bobba Fet","Darth Vader","Luke Cage","Tony Stark"));
+        tasks.put("Darth Vader",createForCompletionTask(taskString, "Darth Vader", 4, "Bobba Fet","Darth Vader","Luke Cage","Tony Stark"));
+        tasks.put("Luke Cage",createForCompletionTask(taskString, "Luke Cage", 4, "Bobba Fet","Darth Vader","Luke Cage","Tony Stark"));
+        tasks.put("Tony Stark",createForCompletionTask(taskString, "Tony Stark", 4, "Bobba Fet","Darth Vader","Luke Cage","Tony Stark"));
+        assertNumberOfNonCompletedTasks("Darth Vader", 1);
+        assertNumberOfNonCompletedTasks("Bobba Fet", 1);
+        assertNumberOfNonCompletedTasks("Luke Cage", 1);
+        assertNumberOfNonCompletedTasks("Tony Stark", 1);
```
triceo: Could we have these people as constants, please? They're supposed to be the same thing, so let's make that absolutely clear.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/LoadCalculator.java
```diff
@@ -0,0 +1,13 @@
+package org.jbpm.services.task.assignment;
+
+import java.util.Collection;
+import java.util.List;
+
+import org.kie.api.task.TaskContext;
+import org.kie.api.task.model.User;
+
+public interface LoadCalculator {
```
mswiderski: would be good to add some java docs as it seems end users can implement this interface to provide custom calculation mechanism. Maybe we should move this to kie-internal somewhere next to https://github.com/kiegroup/droolsjbpm-knowledge/tree/master/kie-internal/src/main/java/org/kie/internal/task/api/assignment

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/LoadCalculatorImpl.java
```diff
@@ -0,0 +1,16 @@
+package org.jbpm.services.task.assignment.impl;
+
+import org.jbpm.services.task.assignment.LoadCalculator;
+
+public abstract class LoadCalculatorImpl implements LoadCalculator {
```
mswiderski: a minor comment ... maybe rename this class to AbstractLoadCalculator instead of LoadCalculatorImpl as (at least to me) it indicates actual impl

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
```diff
@@ -0,0 +1,96 @@
...
+    private static final String SINGLE_USER_QUERY = 
+            "select new Integer(count(t)) "
+            + "from TaskImpl t "
+            + "where t.taskData.actualOwner.id = :owner and t.taskData.status in ('Reserved', 'InProgress', 'Suspended') ";
+    private static final String MULTI_USER_QUERY = 
+            "select new org.jbpm.services.task.assignment.impl.AssignmentImpl(t.taskData.actualOwner.id, count(t)) "
+            + "from TaskImpl t "
+            + "where t.taskData.actualOwner.id in (:owners) and t.taskData.status in ('Reserved', 'InProgress', 'Suspended') "
+            + "group by t.taskData.actualOwner "
+            + "order by count(t) asc, t.taskData.actualOwner.id asc";
```
mswiderski: what's the point of having these two queries? Can't we only use the MULTI_USEER_QUERY and in case of single user just give a list with one user in it? Anyway the list is processed in both cases so we could reuse bit of code

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
```diff
@@ -0,0 +1,96 @@
...
+    
+    
+    public TaskCountLoadCalculator() {
+    	super(IDENTIFIER);
+    }
+
+    Function<AssignmentImpl, String> assignKey = (assignment) -> { return assignment.getUser(); };
```
mswiderski: do we need a function for just one liner?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
```diff
@@ -0,0 +1,96 @@
...
+		Collection<UserTaskLoad> userTaskLoads = new ArrayList<>();
+		List<String> userIds = users.stream().map(user -> {return user.getId();}).collect(Collectors.toList());
+		
+        TaskPersistenceContext persistenceContext = ((org.jbpm.services.task.commands.TaskContext)context).getPersistenceContext();
+        Map<String, Object> params = new HashMap<>();
+        params.put("owners", userIds);
+        
+        logger.debug("DB query to be used for finding assignments :: '{}'", getMultiUserQuery());
+        List<AssignmentImpl> assignments = persistenceContext.queryStringWithParametersInTransaction(getMultiUserQuery(), params, ClassUtil.<List<AssignmentImpl>>castClass(List.class));
+        Map<String,AssignmentImpl> assignmentMap = assignments.stream().collect(Collectors.toMap(assignKey,(assign)->assign));
```
mswiderski: I believe the assignKey function could be replaced with method reference

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/LoadBalanceAssignmentStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+                .forEach(oe -> {
+                    Iterator<OrganizationalEntity> groupUsers = userInfo.getMembersForGroup((Group)oe);
+                    if (groupUsers != null) {
+                        groupUsers.forEachRemaining(user -> {
+                            if (user != null && !excluded.contains(user) && !potentialOwners.contains(user)) {
+                                potentialOwners.add(user);
+                            }
+                        });
+                    }
+                });
```
mswiderski: above couple of lines (collecting users and filtering them by excluded owners) seems to be repeated from other strategies maybe it would make sense to reuse them via abstract strategy or some utility

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/LoadBalanceAssignmentStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+        logger.info("Asking the load calculator [{}] for task loads for the users {}",calculator.getIdentifier(),potentialOwners);
+        List<User> users = potentialOwners.stream().map(entityToUser).collect(Collectors.toList());
```
mswiderski: why do we need to rewrite the list here? potentialOwners already include users only so it's safe to either cast it to User or simply use OrgEntity.getId()

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/LoadBalanceAssignmentStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+        Collection<UserTaskLoad> loads = calculator.getUserTaskLoads(users, taskContext);
+        UserTaskLoad lightestLoad = loads.stream().min(UserTaskLoad::compareTo).orElse(null);
+		return lightestLoad != null ? new Assignment(lightestLoad.getUser().getId()):null;
```
mswiderski: instead of orElse(null) maybe better is to use Optional.isPresent as condition instead of null check

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LoadBalanceAssignmentStrategyTest.java
```diff
@@ -0,0 +1,252 @@
...
+    private static final String LUKE_CAGE = "Luke Cage";
+    private static final String TONY_STARK = "Tony Stark";
+
+
+	@Before
+	public void setUp() throws Exception {
+        System.setProperty("org.jbpm.task.assignment.enabled", "true");
+        System.setProperty("org.jbpm.task.assignment.strategy", "LoadBalance");
+        System.setProperty("org.jbpm.task.assignment.loadbalance.calculator","org.jbpm.services.task.assignment.impl.TaskCountLoadCalculator");
+        System.setProperty("org.jbpm.task.assignment.loadbalance.entry.timetolive", "10"); // this has to be low in order that we update the load balances
```
mswiderski: this (org.jbpm.task.assignment.loadbalance.entry.timetolive) does not seem to be used anywhere in the implementation, then could be removed from here?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/LoadBalanceAssignmentStrategy.java
```diff
@@ -0,0 +1,73 @@
...
+	};
+	
+
+	@Override
+	public String getIdentifier() {
+		return IDENTIFIER;
+	}
+
+	@Override
+	public Assignment apply(Task task, TaskContext taskContext, String excludedUser) {
```
livthomas: `excludedUser` parameter seems to be ignored in this method.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LoadBalanceAssignmentStrategyTest.java
```diff
@@ -0,0 +1,252 @@
...
+    private static final String DARTH_VADER = "Darth Vader";
+    private static final String BOBBA_FET = "Bobba Fet";
+    private static final String LUKE_CAGE = "Luke Cage";
+    private static final String TONY_STARK = "Tony Stark";
+
+
+	@Before
+	public void setUp() throws Exception {
+        System.setProperty("org.jbpm.task.assignment.enabled", "true");
+        System.setProperty("org.jbpm.task.assignment.strategy", "LoadBalance");
```
livthomas: I think setting this system property is not necessary since the strategy is later set using `AssignmentServiceProvider` and so the system property is ignored.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
lanceleverich: The queries are different, in that the single user query only requires retrieving the one Integer, rather than retrieving multiple Assignment objects. That said, the use case for the single user may not be realistic, as the loads should be getting calculated for all potential owners, each time.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
lanceleverich: The assignKey references the earlier defined Function<AssignmentImpl,String> assignKey...
And yes, it could be replaced with a method, but why would a method be preferable to using the Function?
I would be more inclined to change from using the external Function to using an inline function.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
lanceleverich: The one-liner Function makes for easier reading in the lambda referenced in your next comment.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/LoadCalculatorImpl.java
lanceleverich: I will do this

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/LoadCalculator.java
lanceleverich: This is a good idea, but is it necessary for wrapping up this JIRA?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/LoadBalanceAssignmentStrategy.java
lanceleverich: I will see how I can make this work. A challenge is that the data returned from the queries is different; it isn't just user data.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LoadBalanceAssignmentStrategyTest.java
lanceleverich: It is used in the TotalCompletionTimeLoadCalculator, which is part of my latest push.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/strategy/LoadBalanceAssignmentStrategy.java
lanceleverich: Fixed in latest push.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/assignment/LoadBalanceAssignmentStrategyTest.java
lanceleverich: The AssignmentServiceProvider uses the system property to determine which strategy to use. The test has to set the property or the AssignmentServiceProvider will use its default value.

livthomas: But there is the following line later in the `setUp()` method:

```java
AssignmentServiceProvider.override(new LoadBalanceAssignmentStrategy());
```

And it basically sets the default assignment strategy so the system property is ignored in `AssignmentServiceImpl`.

lanceleverich: Good point. To be honest, I probably copied that portion of code from the setup for BusinessRuleAssignmentTest. But since the LoadBalanceAssignmentStrategy doesn't look at the system property, it isn't necessary. On the other hand, I don't believe that this setting of a system property, within the context of the unit test, is going to present any issues. I would rather see the property reflect the value that is being used; especially if later we determine that there is a need to validate/verify that the active strategy is the one we think it is.

manstis: Sorry.. none of my business really.. but in other tests in KIE that set System Properties we usually set as needed in ```@BeforeClass``` and restore in ```@AfterClass```.. can we guarantee other test classes in the same suite/module/JVM don't rely on System Properties being set differently?!

lanceleverich: @manstis No problem with interjecting your thoughts here. I wondered, briefly, about why the author of the earlier tests had not made use of the `@BeforeClass` and `@AfterClass`. I didn't pursue it, and essentially copied the structure that had been used. I don't mind going back and adjusting the structure to the file in question, and in the other tests for the same module/package. While I'm doing that, I can also remove the setting of any extraneous system properties.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/LoadCalculator.java
mswiderski: we can leave it for now, please just add license header as it seems to be missing here

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/assignment/impl/TaskCountLoadCalculator.java
mswiderski: fair enough

jbpm-human-task/jbpm-human-task-audit/pom.xml
```diff
@@ -154,6 +154,10 @@
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-audit</artifactId>
     </dependency>
+    <dependency>
+    	<groupId>com.google.guava</groupId>
+    	<artifactId>guava</artifactId>
```
livthomas: Why is Guava dependency needed here?

lanceleverich:
The TotalCompletionTimeLoadCalculator uses guava Cache and CacheBuilder so
that it doesn't have to calculate average duration of task every time.

mswiderski:
@lanceleverich could you please resolve conflicts as this can't be merged

tsurdilo:
@lanceleverich let me know if i could help you with resolving the git conflicts on your local branch and squash the commits. Would be nice to get this in since the associated Jira is set to resolved already. Thanks.

lanceleverich:
Sorry, I meant to take care of this last night and got too engrossed in
testing/debugging standalone controller. I will fix the collisions when I
return from my doctor appointment.

PR #835, opened 2017-04-26
JBPM-5841: Round Robin Assignment Strategy

livthomas:
I have added more unit tests and fixed ignored excluded users (from the parameter and from from excluded groups).

I am not sure about one unit test. If you have a task with no potential owners or not existing user/group, it will end up waiting for the queue forever. @lanceleverich, is this expected behavior? See the ignored test.


mswiderski:
looks good to me, @lanceleverich could you please review?

PR #838, opened 2017-05-02
Update jbpm-in-container-test to support WebSphere 9

MarianMacik:
Hi, @mswiderski,

I have updated jbpm-in-container tests so they now support WebSphere 9. There should not be any errors in the tests since WebSphere is only run in our Central-CI or locally, of course.

Thanks,

Marian

PR #839, opened 2017-05-02
[7.0.x] Update jbpm-in-container-test to support WebSphere 9

MarianMacik:
Hi, @mswiderski,

this is the same as #838 but for 7.0.x branch.

Thanks,

Marian


MarianMacik:
Fixed.

PR #841, opened 2017-05-04
Change revapi-config.json comment to be more general

MarianMacik:
Hi, @mswiderski,

this is only a minor change to the revapi-config.json file, since these files are no more only on master branch, but on 7.0.x as well.

Marian


MarianMacik:
jenkins retest this

mswiderski:
Jenkins retest this

PR #843, opened 2017-05-08
Add tests for 'async' as a default option for all tasks

mswiderski:
@MarianMacik see the second commit here to find out the adjustment of the test to work with signal. In general it was singling to early - before the async managed to trigger the signal node.


MarianMacik:
+1

MarianMacik:
jenkins retest this

MarianMacik:
The fixed test failed previously on assert, because there were only 2 completed executor requests found. But on my machine it works as it should, i.e. there should be 3 completed commands (for intermediate catch event, script task and end node).

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

PR #845, opened 2017-05-11
JBPM-5974: Extend tests triggering case ad-hoc tasks

livthomas:
I have extended case management tests which cover triggering ad-hoc tasks. This is needed in order to properly test [BPMSPL-317](https://issues.jboss.org/browse/BPMSPL-317).

I have also fixed `processInstanceId` parameter in the JavaDoc of `triggerAdHocFragment` from `CaseService`.

It seems that the files were not properly formatted so there are many whitespace changes. You can look at this commit without whitespace changes [here](https://github.com/kiegroup/jbpm/commit/8327a326fc14395c0bac232146213e7b940c1a12?w=1).

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
@@ -16,20 +16,12 @@
 
 package org.jbpm.casemgmt.impl;
 
-import static java.util.stream.Collectors.toMap;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
-import static org.kie.scanner.MavenRepository.getMavenRepository;
-
 import java.io.File;
 import java.io.FileOutputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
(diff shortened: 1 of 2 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -1779,6 +1757,119 @@ public void testTriggerNotExistingAdHocFragment() {
...
+            List<TaskSummary> tasks = runtimeDataService.getTasksAssignedAsPotentialOwner("john", new QueryFilter());
+            assertThat(tasks).isNotNull().isEmpty();
+
+            caseService.triggerAdHocFragment(caseId, "Task 1", null);
+            tasks = runtimeDataService.getTasksAssignedAsPotentialOwner("john", new QueryFilter());
+            assertThat(tasks).isNotNull().hasSize(1);
+            assertTask(tasks.get(0), "john", "Task 1", Status.Reserved);
+
+            caseService.triggerAdHocFragment(caseId, "Task 2", null);
+            // the task from the next stage should not be triggered
```
livthomas: @mswiderski I am not sure if the current behavior here is correct. Should not we throw `AdHocFragmentNotFoundException` if you try to trigger an ad-hoc fragment from the stage which is not active?

mswiderski:
jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
mswiderski: it could actually, feel free to add that or we can tackle it via another jira

PR #847, opened 2017-05-15
RHBPMS-4729 QuartzSchedulerService increments NEXT_FIRE_TIME by timeC…

tkobayas:
…ycle regardless of current time when calling rescheduleJob()

Please review, thanks!

Some points:

- It calls trigger.setNextFireTime() to update org.drools.core.time.impl.IntervalTrigger/CronTrigger not only SimpleTrigger. If it updates SimpleTrigger only, IntervalTrigger/CronTrigger would stay old so this logic would be executed every time.

- I'm not sure what value is good for "org.jbpm.timer.reschedule.delay" default. I set "500ms" because there was 70ms gap in a known issue case (RHBPMS-4717).

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
@@ -48,6 +48,7 @@ import org.slf4j.LoggerFactory;
 import java.io.NotSerializableException;
 import java.io.Serializable;
 import java.util.Collection;
+import java.util.Date;
 import java.util.concurrent.Callable;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
@@ -66,6 +67,8 @@ public class QuartzSchedulerService implements GlobalSchedulerService {
     private static final Integer START_DELAY = Integer.parseInt(System.getProperty("org.jbpm.timer.delay", "2"));
     private static final Integer FAILED_JOB_RETRIES = Integer.parseInt(System.getProperty("org.jbpm.timer.quartz.retries", "5"));
     private static final Integer FAILED_JOB_DELAY = Integer.parseInt(System.getProperty("org.jbpm.timer.quartz.delay", "1000"));
+
+    private static final Integer RESCHEDULE_DELAY = Integer.parseInt(System.getProperty("org.jbpm.timer.quartz.reschedule.delay", "500"));
 
     private AtomicLong idCounter = new AtomicLong();
     private TimerService globalTimerService;
@@ -156,8 +159,16 @@ public class QuartzSchedulerService implements GlobalSchedulerService {
         jobq.setRequestsRecovery(true);
         jobq.getJobDataMap().put("timerJobInstance", timerJobInstance);
             
+        // Amend nextFireTime not to schedule older than now + RESCHEDULE_DELAY
+        Date nextFireTime = timerJobInstance.getTrigger().hasNextFireTime();
+        Date threshold = new Date(System.currentTimeMillis() + RESCHEDULE_DELAY);
+        if (nextFireTime.before(threshold)) {
+            logger.debug("nextFireTime [" + nextFireTime + "] is older than (now + RESCHEDULE_DELAY). Amending it to [" + threshold + "]");
+            nextFireTime = threshold;
+        }
+
         // Define a Trigger that will fire "now"
-        org.quartz.Trigger triggerq = new SimpleTrigger(quartzJobHandle.getJobName()+"_trigger", quartzJobHandle.getJobGroup(), timerJobInstance.getTrigger().hasNextFireTime());
+        org.quartz.Trigger triggerq = new SimpleTrigger(quartzJobHandle.getJobName()+"_trigger", quartzJobHandle.getJobGroup(), nextFireTime);
             
         // Schedule the job with the trigger
         try {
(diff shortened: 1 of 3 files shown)


jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
```diff
@@ -156,6 +161,19 @@ public void internalSchedule(TimerJobInstance timerJobInstance) {
...
+        Trigger trigger = timerJobInstance.getTrigger();
+        Date nextFireTime = trigger.hasNextFireTime();
+        if (nextFireTime.before(new Date())) {
+            logger.debug("nextFireTime [" + nextFireTime + "] is older than now. Amending it to (now + RESCHEDULE_DELAY " + RESCHEDULE_DELAY + "ms)");
+            nextFireTime = new Date(System.currentTimeMillis() + RESCHEDULE_DELAY);
+            if (trigger instanceof IntervalTrigger) {
+                ((IntervalTrigger)trigger).setNextFireTime(nextFireTime);
+            } else if (trigger instanceof CronTrigger) {
+                ((CronTrigger)trigger).setNextFireTime(nextFireTime);
+            }
```
tsurdilo: Should add log if trigger is instance of another trigger impl type or null.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
```diff
@@ -355,4 +355,128 @@ public void testTimerRequiresRecoveryFlagSet() throws Exception {
...
+        countDownListener.waitTillCompleted(4000);
+
+        manager.disposeRuntimeEngine(runtime);
+
+        assertEquals(5, timerExporations.size());
+
+        manager.close();
+
+        if (propBkFilename != null) {
+            System.setProperty("org.quartz.properties", propBkFilename);
```
tsurdilo: Why set this property at end of test when it will be cleaned in @After? Just wondering

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
```diff
@@ -355,4 +355,128 @@ public void testTimerRequiresRecoveryFlagSet() throws Exception {
...
```
tsurdilo: Same comment here as before

tsurdilo:
@tkobayas thanks for PR - very nice! Could you look at the comments and let me know what you think. thanks.

mswiderski:
@tkobayas thanks for the PR and fix.

one comment, I'd prefer to not update drools/jbpm triggers and let the logic only set the Quartz trigger if the next schedule time is in the past. This is again the same principal to avoid missing timer execution. Let's say we have 5 timer executions missed while server is down (of the same timer instance). If we update the trigger it might result in firing only once instead of five times. So in some cases this might be an issue.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
tkobayas: Thanks, yes, I'll fix it.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
tkobayas: Oops, you are right. I'll fix it.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
tkobayas: Thanks, I'll fix it.

tkobayas:
@tsurdilo Thank you for the review. I'll fix them and push again.

@mswiderski Thank you for the suggestion. I got your point. But how about a case like "1 min timeCycle StartEvent" and you stop a server and restart it after 1 hour. You will get 60 process instances starting in a row. Shall I add a system property to switch the behaviour?

mswiderski:
@tkobayas I'd say that this is expected behaviour - fire 60 timers. For instance this is important to capture that the work was not done in given timer expirations, mainly from KPI/audit point of view. as there might be different logic if given amount of tries exceeded.
Not fully convinced we should have a switch for this...but I am ready to be convinced :)

tkobayas:
@mswiderski I'm convinced :)

@tsurdilo @mswiderski I pushed the revised commit which doesn't modifies drools IntervalTrigger/CronTrigger. So only one unit test method now. Please review gain, thanks!

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
```diff
@@ -156,8 +159,16 @@ public void internalSchedule(TimerJobInstance timerJobInstance) {
         jobq.setRequestsRecovery(true);
         jobq.getJobDataMap().put("timerJobInstance", timerJobInstance);
             
+        // Amend nextFireTime not to schedule older than now + RESCHEDULE_DELAY
+        Date nextFireTime = timerJobInstance.getTrigger().hasNextFireTime();
+        Date threshold = new Date(System.currentTimeMillis() + RESCHEDULE_DELAY);
+        if (nextFireTime.before(threshold)) {
```
tkobayas: In the previous commit, it was nextFireTime.before(now). But now I make sure that the nextFireTime will be later than (now+RESCHEDULE_DELAY).

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
```diff
@@ -305,9 +305,6 @@ public void afterNodeLeft(ProcessNodeLeftEvent event) {
         
         countDownListener.waitTillCompleted(3000);
         assertEquals(2, timerExporations.size());
-
-        
-        manager.close();
```
tkobayas: This is a unit test which I wrote in the past. I removed the line because I noticed that manager is closed by GlobalTimerServiceBaseTest.cleanup().

PR #848, opened 2017-05-16
[6.5.x] RHBPMS-4697 - Fails to start a process with CorrelationKey which is unique but includes correlation properties of an existing correlation key RHBPMS-4693 - Slow JpaProcessPersistenceContext.getProcessInstanceByCorrelationKey() with MySQL

mswiderski:


mswiderski:
@tsurdilo, I believe we need to merge this one (https://github.com/kiegroup/droolsjbpm-integration/pull/960) to otherwise droolsjbpm-integration will fail

tsurdilo:
@mswiderski thanks! will do when jenkins finishes.

PR #849, opened 2017-05-16
JBPM-5851 - Technical Error Events - enhanced DB filter to include ro…

mswiderski:
…llback exception as well that is used when running on Oracle and Postgres

@MarianMacik here is a fix for the test we talked about on Oracle and Postgres. I verified it on Postgres but didn't check it on Oracle, do you mid double check it?


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

MarianMacik:
@mswiderski it works with Oracle 12c too, so I think it is fixed now. Once you merge this I will run the whole test suite to be 100% sure.

Thanks!

mswiderski:
@MarianMacik thanks for checking this. merged so it's ready for full test run

PR #852, opened 2017-05-22
Add .gitignore file to jbpm-persistence module

livthomas:
During the refactoring of the _jbpm-persistence_ module, the **.gitignore** file has been removed from this module's root directory and Git sees all files and directories created by IDEs now. This pull-request addresses this problem.

### DIFF
--- a/jbpm-persistence/.gitignore
+++ b/jbpm-persistence/.gitignore
@@ -1,0 +1,12 @@
+/target
+/local
+/bin
+
+# Eclipse, Netbeans and IntelliJ files
+/.*
+!.gitignore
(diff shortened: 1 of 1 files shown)


tsurdilo:
@livthomas would you mind creating a jira for this pr so we can associate it after? thanks

livthomas:
@tsurdilo, is it really necessary even for such "cosmetic" changes? I know that we want to keep everything important tracked but in my opinion it is a waste of time to create Jiras even for such small things.

tsurdilo:
@livthomas OK I think for this change it is fine. Thanks.

PR #860, opened 2017-05-30
[6.5.x] RHBPMS-4659 - persisting org.slf4j.Logger fails - temp fix for test

tsurdilo:
change in https://github.com/kiegroup/droolsjbpm-build-bootstrap/commit/c1df75a4d209ae8c705a2848f9679658f8051c7d#diff-600376dffeb79835ede4a0b285078036R77
weirdly affected this test where a slf4j logger was intersted into ksession and was persisted. 

Temp fix for this test while someone can figure out why this change in bootstrap affected it.


psiroky:
We should IMO not merge this. We need to upgrade the `slf4j-api` and that will fix the test as well.

PR #862, opened 2017-05-31
JBPM-6013: Extend case stage completion condition tests

livthomas:
I have extended case stage completion condition tests as described in [JBPM-6013](https://issues.jboss.org/browse/JBPM-6013).

Most of the tests are failing because of [JBPM-6029](https://issues.jboss.org/browse/JBPM-6029). It seems the engine checks completion condition only when some task is completed but variables used in this condition may be changed in the meantime and the stage should be completed immediately when this happens.

The other thing I am not sure about is using case file variable without `caseFile_` prefix in the completion condition. I was told it should be possible to use case file variables directly if there is no process variable with the same name. However, it seems it does not work. See the ignored test.

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
@@ -1,0 +1,395 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.casemgmt.impl;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.jbpm.casemgmt.api.dynamic.TaskSpecification;
+import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
+import org.jbpm.casemgmt.api.model.instance.CaseInstance;
+import org.jbpm.casemgmt.impl.model.instance.CaseInstanceImpl;
+import org.jbpm.casemgmt.impl.util.AbstractCaseServicesBaseTest;
+import org.junit.Test;
+import org.kie.api.task.model.TaskSummary;
+import org.kie.internal.query.QueryFilter;
+
+import static org.assertj.core.api.Assertions.*;
+
+public class StageCompletionConditionTest extends AbstractCaseServicesBaseTest {
+
+    private static final String STAGE_WITH_TASK_AUTOCOMPLETE = "StageWithTaskAutocomplete";
+    private static final String STAGE_WITH_TASK_COUNT = "StageWithTaskCount";
+    private static final String STAGE_WITH_TASK_NO_AUTOSTART = "StageWithTaskNoAutoStart";
+    private static final String STAGE_WITH_TASK_PROCESS_VARIABLE = "StageWithTaskProcessVariable";
+    private static final String STAGE_WITH_TASK_CASE_FILE_VARIABLE = "StageWithTaskCaseFileVariable";
+    private static final String STAGE_WITH_TASK_CASE_FILE_VARIABLE_NO_PREFIX = "StageWithTaskCaseFileVariableNoPrefix";
+    private static final String STAGE_WITH_TASK_CASE_DATA_BOOLEAN = "StageWithTaskCaseDataBoolean";
+    private static final String STAGE_WITH_TASK_CASE_DATA_INTEGER = "StageWithTaskCaseDataInteger";
+    private static final String STAGE_WITH_TASK_CASE_DATA_STRING = "StageWithTaskCaseDataString";
+    private static final String STAGE_WITH_TASK_CASE_FILE_AND_PROCESS_VARIABLE = "StageWithTaskCaseFileAndProcessVariable";
+    private static final String USER_TASK_PROCESS = "UserTask";
+
+    private static final String INSIDE_TASK = "InsideTask";
+    private static final String DYNAMIC_TASK = "DynamicTask";
+    private static final String HELLO_TASK = "Hello";
+
+    @Override
+    protected List<String> getProcessDefinitionFiles() {
+        List<String> processes = new ArrayList<>();
+        processes.add("cases/stage/completion/StageWithTaskAutocomplete.bpmn2");
+        processes.add("cases/stage/completion/StageWithTaskCount.bpmn2");
+        processes.add("cases/stage/completion/StageWithTaskNoAutoStart.bpmn2");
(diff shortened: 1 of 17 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
```diff
@@ -0,0 +1,359 @@
...
+        assertThat(insideTask.getName()).isEqualTo(INSIDE_TASK);
+        userTaskService.completeAutoProgress(insideTask.getId(), USER, null);
+        assertCaseInstanceActive(caseId);
+
+        caseService.addDataToCaseFile(caseId, "continue", true);
+        assertCaseInstanceNotActive(caseId);
+    }
+
+    @Test
+    @Ignore
```
rsynek: It would be good to tell the reason for ignoring a test - @Ignore annotation has a parameter for that.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
```diff
@@ -0,0 +1,362 @@
...
+    private static final String INSIDE_TASK = "InsideTask";
+    private static final String DYNAMIC_TASK = "DynamicTask";
+    private static final String HELLO_TASK = "Hello";
+
+    private static final String USER = "john";
+
+    private KModuleDeploymentUnit deploymentUnit;
+
+    @Before
+    public void prepare() throws Exception {
```
rsynek: The method prepares several different things that can be refactored into standalone methods in a parent class and reused in other test cases (if those methods are properly parameterized). Now is has two level of abstraction, as it orchestrates all steps of preparation and deals with details of every step.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
```diff
@@ -0,0 +1,362 @@
...
+import org.kie.api.task.model.TaskSummary;
+import org.kie.internal.query.QueryContext;
+import org.kie.internal.query.QueryFilter;
+import org.kie.internal.runtime.conf.DeploymentDescriptor;
+import org.kie.internal.runtime.conf.NamedObjectModel;
+import org.kie.internal.runtime.conf.RuntimeStrategy;
+import org.kie.scanner.MavenRepository;
+
+import static org.assertj.core.api.Assertions.*;
+import static org.kie.scanner.MavenRepository.*;
```
rsynek: Another star import. In contrast to that, Assertions.* on the line above is fine, as it's an exception in checkstyle configuration (and in general, start imports are usually used for assertions in tests).

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
```diff
@@ -0,0 +1,362 @@
...
+        userTaskService.completeAutoProgress(insideTask.getId(), USER, null);
+        assertCaseInstanceActive(caseId);
+
+        caseService.addDataToCaseFile(caseId, "continue", true);
+        assertCaseInstanceNotActive(caseId);
+    }
+
+    @Test
+    @Ignore("This way of getting case file variable is probably not supported")
+    public void testCaseFileVariableWithoutPrefix() {
```
rsynek: I just wonder what's the expectation? The variable is addresses as a process variable (not a case variable) in the condition of corresponding case.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
livthomas: I have used KIE Group code style and formatted it automatically by the IDE. If such imports are not allowed, it should be changed in the code style.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
livthomas: As I described in the last paragraph of the description of this pull-request, based on the discussion with @MarianMacik it should work. I want to discuss it with developers in this pull-request and if it is not allowed to use case variables in completion conditions this way, I will just remove the test.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
livthomas: I have just copied it from other test classes and already moved some logic to the parent. If I find some time, I may do further refactoring.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
rsynek: @livthomas thanks, I didn't notice that explanation. Makes sense now.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageCompletionConditionTest.java
rsynek: @livthomas that's strange - using https://github.com/kiegroup/droolsjbpm-build-bootstrap/blob/master/ide-configuration/intellij-configuration/code-style/intellij-code-style_droolsjbpm-java-conventions.xml - Idea reformats my code to single import even if I used start import.

mswiderski:
@livthomas if you want to take advantage of reactive nature of completion condition then you need to use drools as language of the condition like here https://github.com/kiegroup/jbpm/blob/master/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/org/jbpm/casemgmt/demo/insurance/CarInsuranceClaimCase.bpmn2#L181 

Using mvel only won't work as there is nothing that evaluates that data without triggering of the process instance

livthomas:
@mswiderski, thanks, I have changed the tests to use Drools language.

livthomas:
@mswiderski, what about the ignored test? Should not it be possible to use case file variable directly without `caseFile_` prefix in stage completion condition?

mswiderski:
that construct is not supported mainly as it uses mvel dialect as well, so it has capabilities to refer to case file data but only to process variables

MarianMacik:
@mswiderski so this construct (caseFile_) only works when mvel is used or is not supported at all? I remember a couple of months ago that it was possible to do that - it was mentioned in your document about case management as well.

mswiderski:
@MarianMacik in general you need to be consistent - so either refer to it as variable or case file data. To make this condition work as it is now, it is enough to declare it as process variable. Once it's there engine will make first check to see if such value is even expected to be in the process, if so then it will evaluate both variables and case file. Or make this variable a case file and then refer to it as case file in the condition.

what I mentioned is that you can refer to case file by it's short name but it needs to be declared

MarianMacik:
Ok, thanks. But still there is a limitation that it only works with mvel, right?

mswiderski:
I wouldn't call it a limitation, every language has its syntax that should be followed to make good use of it e.g. drools and how we access case data etc.

livthomas:
@mswiderski, thanks for clarification. I have changed the case definition used in previously ignored test and it seems to work.

I think this is now ready to be merged.

mswiderski:
Jenkins retest this

PR #866, opened 2017-06-05
[JBPM-6066]  Allow usage of configuration entries in creation of MarshallingStrategies

ngs-mtech:
We are using complex deployment descriptors full of configuration entries, but we are not able to use them in creation of object marshalling strategies. With this small change we can define a configuration entry named project_persistence_unit and use it in the calling to the creation of our custom marshalling strategy.


tsurdilo:
ok to test

mswiderski:
@ngs-mtech could you please create jira and describe the problem/improvment need?

Looks like you have bunch of commits that should already be on master...

And as usual test case would be good to add

ngs-mtech:
I have created the jira to describe the problem: JBPM-6066

As Github says, the branch has no conflicts with the master so these comits may aren't commited yet on the master... :(

Let me think a way to unit test this, may I do a dummy marshalling strategy.

mswiderski:
Could you add a test case for this change?

And please remove the document related commit as it is not part of this fix

ngs-mtech:
Done

shami8099:
hello im interested in learning jbpm can any 1 help me in this

mswiderski:
lack of response for quite some time ... closing

PR #867, opened 2017-06-05
[DROOLS-1593] anticipate flush of propagation queue when evaluating a…

mariofusco:
… rule from a process


mariofusco:
ok to test

PR #872, opened 2017-06-07
[DROOLS-1608] move processes related classes from drools to jbpm

mariofusco:


mariofusco:
depends on https://github.com/kiegroup/drools/pull/1307

mariofusco:
ok to test

mariofusco:
@mswiderski There's one failing test now, but I believe that it is a false negative totally unrelated with this change. What concerns me is that I'm afraid that this change will break backward compatibility since those class names are hardcoded in the process definition files. Is that correct? Can we do something about this?

mswiderski:
@mariofusco yes, that is breaking bakward compatibility though only for rule flows I believe. What we could do is to handle it here [1] in the parser when found org.drools.core package replace it with org.jbpm, wdyt?

[1] https://github.com/kiegroup/jbpm/blob/master/jbpm-flow-builder/src/main/java/org/jbpm/compiler/xml/processes/TypeHandler.java#L58

mariofusco:
@mswiderski I believe that the solution you're suggesting is ok, but unfortunately this is neither the only nor the main point. I discussed this with @mdproctor and @krisv and they both suggested that moving these classes inside jbpm is conceptually wrong because even if at the moment they're not used by drools they could be used (for instance in dtables?) in future. I honestly haven't well understood how, both of course I trust their judgement on this. However I also would like to move a bit forward that better separation between drools and jbpm that we always hoped in the past but never had time to finalize.

Given the current situation I think we should move these and other (even potentially) common classes to kie-internal instead (or to another submodule shared by drools and jbpm). Anyway this is not an urgent issue, so my suggestion is to leave these PRs open for now and rediscuss this with more calm when you'll be back from PTO.

mswiderski:
jenkins retest this

mswiderski:
@mariofusco there seems to be some conflicts, would you mind to rebase?

and what about the issues with backward compatibility of the rule flows where the types are referenced by org.drools?

PR #873, opened 2017-06-07
[JBPM-5975] Enumerations can't be used as predefined values

mariofusco:


tsurdilo:
@mariofusco good to merge when green. thanks!

jbpm-flow-builder/src/test/java/org/jbpm/process/workitem/WorkDefinitionImplTest.java
```diff
@@ -122,7 +122,7 @@ public void testParameterValuesWithEnumsOnly() throws Exception {
             if (entry.getKey().equals("param1")) {
                 EnumDataType paramEnum = (EnumDataType) entry.getValue();
                 assertEquals("org.jbpm.process.workitem.enums.AnimalsEnum", paramEnum.getClassName());
-                Map<String, Object> paramValuesMap = paramEnum.getValueMap();
+                Map<String, Object> paramValuesMap = paramEnum.getValueMap(null);
```
psiroky: I understand this is just quick fix to make the build green again, but passing `null` as an argument into into the method seems weird.

tsurdilo: Agree. I think the drools change for this should have added a getValueMap method that allows to pass CL instead of replacing existing no param one.

mariofusco: @psiroky the method takes a ClassLoader as argument which however is optional. The case of a null arg is correctly managed inside it.

psiroky: Would it then make sense to add the `paramEnum.getValueMap()`again and encapsulate the pain of the `null` in that method?

jsoltes:
@tsurdilo my pull-request builds are failing because of this pull-request (see https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-common-pullrequests/1284/console). Can you please fix/revert it?

mariofusco:
@jsoltes This pull request should actually fix the problem, not cause it. Please double check.

mariofusco:
@psiroky My first idea was also adding back the method with no-arg, but I'm discussing with @mswiderski a bigger change that also involve that class, so I decided to not touch it for now.

tsurdilo:
@jsoltes yeah as @mariofusco said this actually fixes issue with build.

PR #878, opened 2017-06-09
[DROOLS-1582] split kie-ci

mariofusco:


mariofusco:
depends on https://github.com/AppFormer/uberfire/pull/740

jbpm-services/jbpm-kie-services/pom.xml
```diff
@@ -93,7 +93,11 @@
         </exclusion>
       </exclusions>
     </dependency>  
-  
+    <dependency>
+      <groupId>org.uberfire</groupId>
```
krisv: We are introducing an uberfire dependency in the core?  This seems strange to me, UF / AppFormer is all / only about our webtooling?

mswiderski:
@porcelli why was this merged witout being approved? Especially that it brought in appformer dependency into jbpm...

porcelli:
@mswiderski this has been discussed on irc all morning

mswiderski:
@porcelli fine, but there is no conclusion given here even though there is open question on the PR itself. I'd appreciate if such information is given so all (not only on irc) are aware of

porcelli:
@mswiderski i'll try to summarize it here:
- appformer is a merge of UF, Dashbuilder, Guvnor and some parts of kie-wb-common
- appformer is part of KIE organization, but has an independent release cycle
- appformer and kie projects needed to share a minimal set of utils for maven
- kie-* already depends of UF for the web tooling
- jbpm itself already depends of appformer modules (ie. https://github.com/kiegroup/jbpm/blob/b653ccb2fbebe236523828eb94bcdc9e30763389/jbpm-services/jbpm-kie-services/pom.xml#L118)

An alternative (or future solution, open to discussions) is to have a different repository that would have all the shared infra (maven utils and maybe some pieces of dashbuilder), but in order to unblock a series of tasks and people, were decided to go with the maven utils as part of appformer (without any web related depedency) for a matter of simplification of the build procedure for now (probably after the first batch of repositories merge into AppFormer has been done, we'll probably have a better understand of this potential new shared repository).

mswiderski:
@porcelli thanks for details.

Good to know what is part of appformer. Let's keep it as is and see if that can be improved in anyway later on. Will ping you once I am back to get to know all the details. Again, thanks for taking the time to describe it!

PR #883, opened 2017-06-15
JBPM 6073: getComments method not returning desired page

annabaker:
Core CaseService was not paginating case comment results via getCaseComments method.  New method added that accepts page and pageSize as parameters + unit test.

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseService.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseService.java
@@ -139,7 +139,7 @@ public interface CaseService {
      * @throws CaseActiveException thrown when case is still active
      */
     void reopenCase(String caseId, String deploymentId, String caseDefinitionId, Map<String, Object> data) throws CaseNotFoundException;
-    
+
     /*
      * dynamic case operations section
      */
@@ -247,7 +247,7 @@ public interface CaseService {
      * @throws CaseNotFoundException thrown in case case was not found with given id
      */
     void triggerAdHocFragment(Long processInstanceId, String fragmentName, Object data) throws CaseNotFoundException;
-    
+
     /*
      * Case file section
      */
@@ -284,7 +284,7 @@ public interface CaseService {
      * @throws CaseNotFoundException thrown in case case was not found with given id
      */
     void removeDataFromCaseFile(String caseId, List<String> variableNames) throws CaseNotFoundException;
-    
+
     /*
      * Case role section
      */
@@ -315,7 +315,7 @@ public interface CaseService {
      * @throws CaseNotFoundException thrown in case case was not found with given id
      */
     Collection<CaseRoleInstance> getCaseRoleAssignments(String caseId) throws CaseNotFoundException;
-    
+
     /*
      * Case comments section
      */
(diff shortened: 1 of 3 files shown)


tsurdilo:
ok to test

jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseService.java
```diff
@@ -338,6 +338,16 @@
...
 
     /**
+     * Returns all case comments sorted with given sortBy
+     * @param caseId unique case id in the format PREFIX-GENERATED_ID as described on startCase method
+     * @param page defines the specific page index to query
+     * @param pageSize defines how many items per page
+     * @return a specific page of comments added to given case
+     * @throws CaseNotFoundException thrown in case case was not found with given id
+     */
+    Collection<CommentInstance> getCaseComments(String caseId, int page, int pageSize, QueryContext queryContext) throws CaseNotFoundException;
```
mswiderski: QueryContext already carries pagination information so there is no need for new method

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
```diff
@@ -438,45 +428,62 @@ public void removeFromCaseRole(String caseId, String role, OrganizationalEntity
...
+
+        int allCommentsSize = commentsToPaginate.size();
+        int pageIndex = (allCommentsSize + pageSize - 1) / pageSize;
+
+        if (allCommentsSize < pageSize) {
+            return comments;
+        } else if (pageIndex == page + 1) {
+            return commentsToPaginate.subList(page * pageSize, allCommentsSize);
+        } else {
+            return commentsToPaginate.subList(page * pageSize, (page * pageSize) + pageSize);
```
mswiderski: The main reason why we didn't add pagination in first place was it is an in memory (list modification) operations so it has limited value as it needs to be loaded from the store anyway. So I thought it would be better to return all and let the client to pagination as it could then apply sorting etc. 
but if we see value in this then I am fine too, just do this in single method

annabaker: @cristianonicolai any input?  since we are dealing with comments, the data set we need to pull is probably going to be relatively small (or maybe i am making an incorrect assumption).  in that case, would it make more sense to handle pagination on the client side?

cristianonicolai: @mswiderski @annabaker I would still be in favor of adding the pagination here, to try to keep consistency in how the UI interact with most services, limiting the amount of data we transfer and load. That I think can also be future proof in terms of core services changing how comments are stored/loaded for instance.

mswiderski: fine with me to do pagination here

annabaker: fair enough.  finishing this up now.

PR #884, opened 2017-06-16
JBPM-6094 - No exception thrown when adding not existing subprocess

tsurdilo:


tsurdilo:
associated pr: https://github.com/kiegroup/droolsjbpm-integration/pull/1012

tsurdilo:
@mswiderski @livthomas can you guys pls review? Thanks.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java
```diff
@@ -55,195 +55,262 @@
...
-            logger.error("Could not find process {}", processId);
+            logger.error("Could not find process {}",
+                         processId);
             logger.error("Aborting process");
-        	processInstance.setState(ProcessInstance.STATE_ABORTED);
-        	
-        	return -1;
+            processInstance.setState(ProcessInstance.STATE_ABORTED);
+
+            return -1;
```
mswiderski: I would prefer to throw an exeption here instead of returning -1.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java
```diff
@@ -55,195 +55,262 @@
...
+        Process process = ksession.getKieBase().getProcess(processId);
         if (process == null) {
-            logger.error("Could not find process {}", processId);
+            logger.error("Could not find process {}",
+                         processId);
             logger.error("Aborting process");
-        	processInstance.setState(ProcessInstance.STATE_ABORTED);
-        	
-        	return -1;
+            processInstance.setState(ProcessInstance.STATE_ABORTED);
```
mswiderski: @krisv I think we should not abort the main process instance ( which might be the case nstance) when process id was not found, wdyt?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDynamicProcessCommand.java
```diff
@@ -38,34 +39,51 @@
...
+                                                       processId,
+                                                       parameters);
+
+        long subProcessInstanceId = DynamicUtils.addDynamicSubProcess(processInstance,
+                                                                      ksession,
+                                                                      processId,
+                                                                      parameters);
+
+        if (subProcessInstanceId < 0) {
+            throw new ProcessDefinitionNotFoundException("No process definition found with id: " + processId);
```
mswiderski: We shoul get the exception (most likely if different type as DynamicUtils is in jbpm-flow so ProcessDefinitionNotFoundException is not available there) here and then wrap it with ProcessDefinitionNotFoundException and rethow it

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDynamicProcessToStageCommand.java
```diff
@@ -74,12 +80,25 @@ public Long execute(Context context) {
...
+                                                       processId,
+                                                       parameters);
+
+        long subProcessInstanceId = DynamicUtils.addDynamicSubProcess(dynamicContext,
+                                                                      ksession,
+                                                                      processId,
+                                                                      parameters);
+
+        if (subProcessInstanceId < 0) {
+            throw new ProcessDefinitionNotFoundException("No process definition found with id: " + processId);
```
mswiderski: Same here -catch and rethow as ProcessDefinitionNotFoundException

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java
tsurdilo: @mswiderski when debugging this i noticed that  setting the process state to aborted is i think even wrong. imo user error should not abort it in this case. wdyt?

tsurdilo:
@mswiderski amended with the throwing exception requested changed. will wait for @krisv response on the process abort question to amend that as well if neded

tsurdilo:
Jenkins retest please

tsurdilo:
Jenkins retest please

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java
mswiderski: Fully agreed, especially when we deal with case mgmt

tsurdilo:
@mswiderski i went ahead an removed the case instance abort and updated tests. will follow up with @krisv if needed but im pretty sure he would agree with removing the abort. Think this is ready now.

PR #885, opened 2017-06-19
JBPM 6073: getComments method not returning desired page

annabaker:
-Modified core getComments method to use page/pageSize info from query context for pagination purposes

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
@@ -876,6 +876,65 @@ public class CaseServiceImplTest extends AbstractCaseServicesBaseTest {
             }
         }
     }
+    
+    @Test
+    public void testCaseWithCommentsPagination() {
+        Map<String, OrganizationalEntity> roleAssignments = new HashMap<>();
+        roleAssignments.put("owner", new UserImpl("john"));
+
+        Map<String, Object> data = new HashMap<>();
+        CaseFileInstance caseFile = caseService.newCaseFileInstance(deploymentUnit.getIdentifier(), USER_TASK_STAGE_AUTO_START_CASE_P_ID, data, roleAssignments);
+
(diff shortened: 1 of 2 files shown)


tsurdilo:
as always there should be a test associated with changes. thanks!

annabaker:
@tsurdilo right, sorry.  done.

tsurdilo:
ok to test

tsurdilo:
@annabaker QueryContext iirc has a default count of 10 -- with your changes I don't think there is any more way to get all of the case comments. With the default users will get 10 at a time which may be good for the UI, but how do i now retrieve all comments like before? Thanks.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -876,6 +876,65 @@ public void testCaseWithComments() {
             }
         }
     }
+    
+    @Test
+    public void testCaseWithCommentsPagination() {
```
tsurdilo: need at all test case where no explicit offset and size are passed to queryContext (new QueryContext()) to make sure in that case all of case comments can be returned (not only 10). Otherwise imo this is a regression.

annabaker: @tsurdilo saw your last comment and good point.  i have something working locally now.. what i did was refactor the method signature to pass in a boolean parameter for pagination.  if this is set to false, it will return all comments.  but not sure if that's the best solution, or if it would be best to remove the default count on the QueryContext.. @mswiderski @cristianonicolai

tsurdilo: @annabaker after talking to @mswiderski about this decided that it's fine - these methods take QueryContext as param so possible paging is expected. Sorry for the alarming :)

tsurdilo:
Jenkins retest please

mswiderski:
@annabaker I'd recommend to remove this extra method with boolean as it has limited value. What we need is to only add the pagination into the getComments method that relies on the QueryContext data.

annabaker:
@tsurdilo @mswiderski removed the boolean and removed test for retrieving all comments

PR #888, opened 2017-06-23
JBPM-6144 - Create a wid specific MVELEvaluator

tsurdilo:
JBPM-6144 - Create a wid specific MVELEvaluator

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/util/WidMVELEvaluator.java
+++ b/jbpm-flow/src/main/java/org/jbpm/util/WidMVELEvaluator.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.util;
+
+import java.util.HashMap;
+
+import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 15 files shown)


tsurdilo:
in wid configs inports to core data types no longer needed. 
imports for custom data types still needed.
wids with legacy imports of data types can still be parsed correctly

mswiderski:
@tsurdilo I think we need to add support for KiePolicy to make sure we don't have security issues when security manager is enabled, similar to this one https://github.com/kiegroup/drools/blob/master/drools-core/src/main/java/org/drools/core/util/MVELSafeHelper.java#L30

Except this it looks good

tsurdilo:
@mswiderski good point - added with test . thanks.

tsurdilo:
@mswiderski note that WidMVELEvaluatorSafeTest has to be ignored as setting java.security.policy breaks surefire. Same issue with org.drools.core.util.MVELSafeHelperTest but it can be executed manually and it runs ok.

tsurdilo:
Jenkins retest please

jbpm-flow/src/main/java/org/jbpm/util/WidMVELEvaluator.java
```diff
@@ -0,0 +1,69 @@
...
+                @Override
+                public Object run() {
+                    return MVEL.executeExpression(compiler.compile(),
+                                                  new HashMap());
+                }
+            }, KiePolicyHelper.getAccessContext());
+        }
+
+        return MVEL.executeExpression(compiler.compile(),
+                                      new HashMap());
```
mswiderski: Shouldn't this be in else statement as right now I believe it will be executed twice in case security manager is enabled

tsurdilo: @mswiderski when debugging tests for WidMVELEvaluator.java it is not being executed twice. the return on line 49 completes the eval method call if KiePolicyHelper.isPolicyEnabled() returns true. I can add the else if you still want no problem. thanks

tsurdilo:
@mswiderski added. thanks

jbpm-flow/src/main/java/org/jbpm/util/WidMVELEvaluator.java
mswiderski: Ah now I see, my bad sorry for false comment :(

PR #891, opened 2017-06-26
JBPM-5880 - 'async' as a default option for all tasks - test stabiliz…

mswiderski:
…ation

@tsurdilo here is an attempt to make these tests more stable (or completely stable) ... as they cause false alarms on PR builds

### DIFF
--- a/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/AsyncContinuationSupportTest.java
+++ b/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/AsyncContinuationSupportTest.java
@@ -33,7 +33,10 @@ import javax.persistence.EntityManagerFactory;
 import javax.persistence.Persistence;
 
 import org.jbpm.bpmn2.handler.ServiceTaskHandler;
+import org.jbpm.executor.AsynchronousJobEvent;
+import org.jbpm.executor.AsynchronousJobListener;
 import org.jbpm.executor.ExecutorServiceFactory;
+import org.jbpm.executor.impl.ExecutorServiceImpl;
 import org.jbpm.process.core.async.AsyncSignalEventCommand;
 import org.jbpm.process.instance.impl.demo.SystemOutWorkItemHandler;
 import org.jbpm.runtime.manager.impl.DefaultRegisterableItemsFactory;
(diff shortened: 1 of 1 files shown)


tsurdilo:
@mswiderski very nice!!!

MarianMacik:
Interesting, what was the problem with Process Event Listener?

mswiderski:
@MarianMacik  the problem is that process engine and its transaction will complete earlier than the job that triggered to reach the end - as they are two separate transactions. So it could be that the latch is released before executor saved the status of the last command making it fail on assert.

Again, this is just an attempt so we'll see if it will help...

mswiderski:
ok, still failing ... will take another look

mswiderski:
jenkins retest this

tsurdilo:
Jenkins retest please

MarianMacik:
So I checked the output a little bit and the last AsyncSignalEventCommand with id=3 is still in status RUNNING. And after that you can see that there is:
``` 
ERROR The object is already closed [90007-173]
ERROR HHH000337: Unable to mark for rollback on PersistenceException: 
org.hibernate.service.UnknownServiceException: Unknown service requested
```

but maybe it is because the executor is destroyed while the command is still running. I am only guessing though. So should the test wait a little bit more to make sure that all jobs are in DB?

mswiderski:
@MarianMacik exactly, that was the purpose of the refined commit to collect more information. This now proves it is the issue with command not being yet completed before we assert on it.

One option would be to query first for running jobs and wait a bit if there are found otherwise go straight to asserts. Wdyt?

MarianMacik:
Yes, that would be quite clean. We can include even a loop which will e.g. wait 100 ms, but max 10 iterations and every time it will check if there are any jobs which are still running - I think it is better then permanent Thread sleep for 1 sec. So I agree with your solution.

With regards to my last section in previous comment - why is this even happening? I would have thought that Executor thread will complete process by executing async signal which will invoke Process listener which will register afterCompletion countDown(), so after Executor job is finished and committed, everything should be in DB and the test can continue.

mswiderski:
@MarianMacik correct, that was what I expected as well. Though keep in mind that executor will commit tx for process and then do some work on the job itself - like call other callbacks (if any) and then update the job and commit to db. Somehow this last step is not done or done too late... using the async job listeners should prevent it but it then led to timeout of the test which would indicate the job does not complete somehow...

MarianMacik:
@mswiderski so you are saying that there is another possibility to use AsyncJobListener (as you did with logging) but then it timed out, that shouldn't be happenning right? That is even stranger. But we will see after a little wait if the last job gets finished, I hope it will. Thanks for clarification ;)

mswiderski:
@MarianMacik @tsurdilo first build is green with the wait for all jobs to finish check. let's see few more times...

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
jenkins retest this

mswiderski:
ok, 4 builds in a roll without failures with AsyncMode tests, so I think we are good

mswiderski:
removed the debug listener so we'll have another build

mswiderski:
Ok 5 builds in a roll with all green, I think it's ready to be merged

PR #898, opened 2017-07-03
JBTM

tomjenkinson:
@mswiderski please do let me know the protocol to get this merged.


mswiderski:
ok to test

mswiderski:
@tomjenkinson thanks for the PR.

in general it needs to be reviewed and build green, that's all :)

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/util/PoolingDataSource.java
```diff
@@ -0,0 +1,148 @@
...
+            xads = (XADataSource)Class.forName(className).newInstance();
+            String url = driverProperties.getProperty("URL");
+            xads.getClass().getMethod("setURL", new Class[] {String.class}).invoke(xads, new Object[] {url});
+
+            connection = getConnection(driverProperties.getProperty("user"), driverProperties.getProperty("password"));
+            connection.getMetaData();
+//            Connection connection = xaConnection.getConnection();
+//            connection.prepareStatement("DROP ALL OBJECTS");
+//            connection.commit();
+//            xaConnection.close();
```
MarianMacik: Please remove commented code if it's not needed. Thanks!

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/DeadlinesBaseTest.java
```diff
@@ -409,24 +409,24 @@ public void testDelayedEmailNotificationOnDeadlineTaskCompletedMultipleTasks() t
...
           reader = new InputStreamReader(getClass().getResourceAsStream(MvelFilePath.DeadlineWithNotification));
           InternalTask task2 = (InternalTask) TaskFactory.evalTask(reader, vars);          
-          ((InternalTaskData) task2.getTaskData()).setSkipable(true);          
+          ((InternalTaskData) task2.getTaskData()).setSkipable(true);
           task2.setPeopleAssignments(assignments);
-          
+
           taskService.addTask(task, new HashMap<String, Object>());
-          taskService.addTask(task2, new HashMap<String, Object>());
+          taskService.addTask(task2, new HashMap<String, Object>());  // TODO - the TSR is accessed by drools during afterCompletion
```
MarianMacik: Will this be done in another PR?

tomjenkinson: Sorry, I can remove that - it was addressed in my changes to the  Synchronization. I will remove the wording.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/util/PoolingDataSource.java
tomjenkinson: Thanks - I have removed this now

mswiderski:
jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
looks like the failures are related to remove of PoolingDataSource/bitronix in jbpm-test module

manstis:
What about all the REST failing tests on [kie-wb-distributions](https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/kie-wb-distributions/) (since 4th July) can those failures also possibly be related!?

mswiderski:
@manstis this change is only to replace bitronix in jbpm with narayana so it has no impact on workbench at all. Moreover in v7 we don't have any persistence need in workbench either thus no need for anything related to this PR

tsurdilo:
@tomjenkinson to run the integration tests you do with
mvn clean install -Pwildfly10

to run a specific integration test you can do

mvn clean install -Pwildfly10 -Dit.test=myTestName

thanks

manstis:
@tomjenkinson The authentication issue has been reported to support by @mbiarnes

MarianMacik:
@tomjenkinson @mswiderski please see #902 for explanation regarding jbpm-remote-ejb-test issues.

Regarding the jbpm-in-container-test module, BTM is only used with Tomcat since it does not support JTA out of the box. Therefore I would expect changes only in Tomcat profile. Is there any reason the datasource name was changed with every container?

Thanks

jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/pom.xml
```diff
@@ -101,6 +96,12 @@
       <scope>test</scope>
     </dependency>
 
+    <!--<dependency>-->
+      <!--<groupId>org.hibernate</groupId>-->
+      <!--<artifactId>hibernate-core</artifactId>-->
+      <!--<scope>provided</scope>-->
+      <!--<scope>provided</scope>-->
+    <!--</dependency>-->
```
MarianMacik: Why was this dependency commented?

tomjenkinson: oversight - I will revert

jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml
```diff
@@ -6,10 +6,16 @@
...
+    <Transaction factory="org.jboss.narayana.tomcat.jta.UserTransactionFactory" />
 
-    <Transaction factory="bitronix.tm.BitronixUserTransactionObjectFactory" />
+    <!-- Database resources -->
+    <Resource name="h2DataSource" uniqueName="h2Datasource" auth="Container" type="org.h2.jdbcx.JdbcDataSource" username="sa"
+              user="sa" password="sa" url="jdbc:h2:mem:test-db;MVCC=TRUE;DB_CLOSE_DELAY=-1" description="H2 Data Source" loginTimeout="0"
+              factory="org.h2.jdbcx.JdbcDataSourceFactory"/>
+    <Resource name="jdbc/jbpm" uniqueName="transactionalDataSource" auth="Container" type="javax.sql.DataSource"
+              username="sa" password="sa" driverClassName="com.arjuna.ats.jdbc.TransactionalDriver"
+              url="jdbc:arjuna:java:comp/env/h2DataSource" description="Transactional Driver Data Source"/>
```
MarianMacik: Why are there 2 datasources?

I suspect this will only work with H2 since I don't see any filtering as previously where there were org.jbpm.datasource.* Maven properties. We need to test with any DBs, not only with H2.

mswiderski:
@MarianMacik we need to do similar as we did for KIE Server - externalise the actual data source for tomcat via context.xml. See point 5 in this instructionhttps://raw.githubusercontent.com/kiegroup/droolsjbpm-integration/master/kie-server-parent/kie-server-wars/kie-server/src/main/webc-resources/README.md

that way we can have context.xml per db that most likely would be a filtered resource so we could add various properties in there.

MarianMacik:
@mswiderski well yes I think that if we move it to filtered-resources and add those maven properties to it it should work. Is this what you meant?

mswiderski:
@MarianMacik yes, that is what I tried to say :)

MarianMacik:
Well I am seeing only jbpm-remote-ejb-test module failing, but you didn't do any changes to it. Or did I miss anything?

jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml
tomjenkinson: The com.arjuna.ats.jdbc.TransactionalDriver datasource wraps the h2DataSource - it should be able to wrap any DS type.

MarianMacik:
Yes, exactly, I am talking about those remote ejb tests which haven't passed on CI, but they should. It is strange since there were no modifications.

Regarding the tomcat and web.xml. Right now it is copied every time regardless of the used container, which is OK. It is done so [here](https://github.com/tomjenkinson/jbpm/blob/91c0aab3801943198638e3dd36416c2f672661c4/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/archive/LocalTransactions.java#L90-L90).

BTW tests are not run using -P parameter. There is a [readme](https://github.com/tomjenkinson/jbpm/tree/91c0aab3801943198638e3dd36416c2f672661c4/jbpm-container-test/jbpm-in-container-test) for these tests.
The `container.profile` property is here because it activates shrinkwrap dependencies in `shrinkwrap-war-profiles`. Just run `mvn clean install -Dcontainer.profile=tomcat8` from `jbpm-in-container-test` directory which acts as a root for all submodules.

MarianMacik:
Sorry, I was thinking about tomcat context.xml file. Well I guess you will have to do it in the code by checking port probably. But when I run `mvn clean install -Dcontainer.profile=tomcat8` I am getting
``` Caused by: java.lang.ClassNotFoundException: bitronix.tm.resource.ResourceObjectFactory ```
Which is there because it is configured in https://github.com/tomjenkinson/jbpm/blob/master/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/pom.xml#L698 so what is the equivalent for Narayana/JBTM? I think we should start here.

**Comment by mswiderski** [comment]
@tomjenkinson I think you can avoid to modify web.xml if you annotate the listener class with http://docs.oracle.com/javaee/7/api/javax/servlet/annotation/WebListener.html that way it will be directly loaded by tomcat once it's on class path

**Comment by MarianMacik** [comment]
@tomjenkinson I see it on the command line. [Here](https://pastebin.com/wrt3N91v) it is.

**Comment by mswiderski** [comment]
maybe cargo config we use for kie server can help here 
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/pom.xml#L367
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/pom.xml#L321

**Comment by MarianMacik** [comment]
@mswiderski the config you posted is for H2 only since kie-server is not prepared for production DBs in community. We have separate configuration which we use for our tests, but jbpm-in-container tests config is more or less the same as our internal one. So we will see. At least we have another source of configuration.

Anyway, I would prefer the profiles way which you posted in your config here:
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-tests/kie-server-integ-tests-jbpm/pom.xml#L321

And why there is no such listener in kie-server tests?

**Comment by MarianMacik** [comment]
Great! Btw ignore my previous post since it was with context.xml file, but we are now talking about web.xml :D

**Comment by MarianMacik** [comment]
@tomjenkinson I have found that you deleted datasource from cargo plugin - so I don't think it will work now.

**Comment by mswiderski** [comment]
@tomjenkinson will check it out tomorrow why it does not find it

**Comment by MarianMacik** [comment]
@tomjenkinson I have cloned your repo and tried to run ejb remote test _without_ your change to ```persistence.xml``` using ```mvn clean install -Dcontainer.profile=wildfly10``` and it all passed. I really cannot see any connection between `java.io.EOFException: Read past end of file` and this Hibernate property. Do we really need to turn on that property since it is a bad practice? Thanks.

File: jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml (hunk 4 above)
Comment (MarianMacik) [inline jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml]: OK so can this be made a filtered resource so we can use e.g. PostgreSQL with it like before?

Comment (tomjenkinson) [inline jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml]: I can certainly try, I am not sure how the filtered resources work right now

**Comment by MarianMacik** [comment]
@tomjenkinson But jbpm-remote-ejb is not run with bitronix or narayana on classpath. It is run only with WildFly and it is passing for me **_without_** the change in persistence.xml. Can you try that too?

**Comment by mswiderski** [comment]
@tomjenkinson looking at the compilation issue at the moment

File: jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml (hunk 4 above)
Comment (MarianMacik) [inline jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml]: Please look on two files above btm-resources.properties and btm-config.properties, they were filtered resources.

File: jbpm-runtime-manager/src/test/resources/kjar/jbpm-module (1).iml [hunk 5]
```diff
@@ -0,0 +1,12 @@
...
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_5">
+    <output url="file://$MODULE_DIR$/target/classes" />
+    <output-test url="file://$MODULE_DIR$/target/test-classes" />
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/target" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
MarianMacik: Also, please remove this file from commit since it is only for IDE.

mswiderski:
fix for compilation error is here https://github.com/kiegroup/jbpm/pull/908

jbpm-services/jbpm-services-cdi/src/test/resources/kjar-jpa/persistence-test (2).iml
```diff
@@ -0,0 +1,12 @@
...
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_5">
+    <output url="file://$MODULE_DIR$/target/classes" />
+    <output-test url="file://$MODULE_DIR$/target/test-classes" />
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/target" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
MarianMacik: Same as before - IDE file.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-client/src/test/resources/kjar/custom-data-project.iml
```diff
@@ -0,0 +1,12 @@
...
+  <component name="NewModuleRootManager" LANGUAGE_LEVEL="JDK_1_5">
+    <output url="file://$MODULE_DIR$/target/classes" />
+    <output-test url="file://$MODULE_DIR$/target/test-classes" />
+    <content url="file://$MODULE_DIR$">
+      <excludeFolder url="file://$MODULE_DIR$/target" />
+    </content>
+    <orderEntry type="inheritedJdk" />
+    <orderEntry type="sourceFolder" forTests="false" />
+  </component>
+</module>
```
MarianMacik: Same as before - IDE file.

MarianMacik:
@tomjenkinson OK, I think I may have found the root cause: There is a change in https://github.com/kiegroup/jbpm/pull/898/files#diff-8acfbfff2711aaf51b103a385a64750e  -> TaskTransactionInterceptor. Can you please revert those changes back and run it on our KIE Jenkins so we will see if remote-ejb tests will still fail or will pass? I don't know why changes were made there but it may have changed the way the engine behaves now since this.service.tpm.endCommandScopedEntityManager(); is not called anymore in afterCompletion() method. @mswiderski wdyt? Thanks.

mswiderski:
@tomjenkinson  if that is the case why it does work on WildFly/EAP at the moment?

mswiderski:
moreover when looking at the code it should also fail for other parts as exactly same code is used for processes - org.drools.persistence.jpa.AbstractPersistenceContextManager.endCommandScopedEntityManager() is the method that is actually called here. but then a process without user tasks is also calling this method https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/PersistableRunner.java#L494

MarianMacik:
@mswiderski @tomjenkinson I have confirmed it by running jbpm-remote-ejb-test locally. Once with master version of jbpm-human-task-jpa - tests have passed and once with jbpm-human-task-jpa from this PR - tests have failed. So it is not Narayana/BTM change, rather it is TaskTransactionInterceptor change which causes jbpm-remote-ejb tests to fail.

BTW I have found this change too: https://github.com/kiegroup/jbpm/pull/898/files#diff-268b3f2ef2189c820758c35ba5983f82 Not sure if it is a fix indeed - what would happen if the scheduler can't be shutted down for whatever reason? Won't it permanently stuck in a loop? Or how long do we want to wait?

mswiderski:
@tomjenkinson correct, the entity manager is not cleared in that case but this line is invoked https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-jpa/src/main/java/org/drools/persistence/jpa/AbstractPersistenceContextManager.java#L120

Since that code already handles case when there is no transaction so TSR will not be available (to be more precise its resources) then it returns null but will still clean up the ksession environment. So I would say we should keep the call to the endCommandScopedEntityManager in there as it should not cause any problems. Did you get any errors when running this without your modification of this commit : https://github.com/kiegroup/jbpm/pull/898/commits/981393739b82d982f07b5b3ef799608a5f126178

jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml
tomjenkinson: I have looked at this and it is still not clear to me how the filtered-resources work sorry :(

I tracked references of the file btm-resources.properties to bitronix.resources.properties in the pom.xml, I can see that is referenced then in btm-config.properties via the property bitronix.tm.resource.configuration. btm-config is referenced back in the pom by the properties bitronix.config.properties which sets a system property bitronix.tm.configuration.

I therefore assume that BTM itself is processing those files so I just not sure what is needed here.

Let me explain the config I have put in the tomcat-context.xml and then you can let me know what you think is best.

Narayana wraps XA datasources within the com.arjuna.ats.jdbc.TransactionalDriver. So in the config I have here I need a standard H2 datasource (h2DataSource), then I need to reference that wrapped by the com.arjuna.ats.jdbc.TransactionalDriver. Arjuna registers for jdbc:arjuna URLS and the rest of the URL is then used to determine the URL of the datasource we need to wrap.

Is this tomcat-context.xml filtered at all I wonder? Perhaps I can use ${org.jbpm.datasource.connection.url} for the H2 URL and ${org.jbpm.datasource.username} and so on directly in that XML file and they might be replaced at build time?

tomjenkinson: I think I can put those properties in this file and they seem to be filtered

mswiderski:
@tomjenkinson about to test this locally, let me spend some time on the afterCompletion thingy as it might be simply an issue with test itself.

jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml
tomjenkinson: Yeah. That worked - thanks for the feedback.

MarianMacik: Sorry, for the late response. Yes, resources are filtered in pom.xml file. Thanks!

mswiderski:
nice we have a green build :)

@tomjenkinson I sent PR to your repo with small adjustments https://github.com/tomjenkinson/jbpm/pull/1 

so is there anything else to be tackled? Any TODOs or missing bits?

jbpm-audit/pom.xml
```diff
@@ -67,6 +67,10 @@
       <groupId>com.thoughtworks.xstream</groupId>
       <artifactId>xstream</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.jboss.narayana.jta</groupId>
+      <artifactId>narayana-jta</artifactId>
```
mswiderski: shouldn't this be in test scope? same as the removed btw was? Otherwise it will be included in all other assemblies like for EAP or WAS/Weblogic

jbpm-container-test/jbpm-remote-ejb-test/jbpm-remote-ejb-test-app/src/main/resources/META-INF/persistence.xml
```diff
@@ -68,6 +68,7 @@ http://java.sun.com/xml/ns/persistence/orm http://java.sun.com/xml/ns/persistenc
 			<!-- BZ 841786: AS7/EAP 6/Hib 4 uses new (sequence) generators which seem 
 				to cause problems -->
 			<property name="hibernate.id.new_generator_mappings" value="false" />
+			<property name="hibernate.enable_lazy_load_no_trans" value="true" />
```
MarianMacik: Well I am still struggling with this change. :) Won't it work without this property after Maciej's changes? I am reading everywhere that this setting is not a good practice.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/DeadlinesBaseTest.java
```diff
@@ -389,7 +389,7 @@ public void testDelayedReassignmentOnDeadline() throws Exception {
         assertTrue(ids.contains("Jabba Hutt"));
     }
 
-      @Test(timeout=10000)
+      @Test(timeout=1000000)
```
mswiderski: my bad as I increased that while debugging, can be go back to default 10000

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/concurrent/SessionTest.java
```diff
@@ -688,7 +688,76 @@ public void run() {
...
+//						at org.hibernate.jpa.spi.AbstractEntityManagerImpl.flush(AbstractEntityManagerImpl.java:1303)
+//						at org.drools.persistence.TriggerUpdateTransactionSynchronization.beforeCompletion(TriggerUpdateTransactionSynchronization.java:76)
+//						at org.drools.persistence.jta.JtaTransactionSynchronizationAdapter.beforeCompletion(JtaTransactionSynchronizationAdapter.java:54)
+//						at com.arjuna.ats.internal.jta.resources.arjunacore.SynchronizationImple.beforeCompletion(SynchronizationImple.java:76)
+//						at com.arjuna.ats.arjuna.coordinator.TwoPhaseCoordinator.beforeCompletion(TwoPhaseCoordinator.java:368)
+//						at com.arjuna.ats.arjuna.coordinator.TwoPhaseCoordinator.end(TwoPhaseCoordinator.java:91)
+//						at com.arjuna.ats.arjuna.AtomicAction.commit(AtomicAction.java:162)
+//						at com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionImple.commitAndDisassociate(TransactionImple.java:1289)
+//	... 17 more
+//						Caused by: org.hibernate.StaleStateException: Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1
```
mswiderski: can we maybe reduce the stack trace here and leave only the root cause info?

jbpm-container-test/jbpm-remote-ejb-test/jbpm-remote-ejb-test-app/src/main/resources/META-INF/persistence.xml
MarianMacik: Isn't it even a bug in remote EJBs? Why do we have to "hack" this externally with the Hibernate property?

tomjenkinson: It should do - you are right. I will revert it and let CI verify.

pom.xml
```diff
@@ -135,6 +140,21 @@
...
+      <dependency>
+        <groupId>org.jboss.narayana.jta</groupId>
+        <artifactId>narayana-jta</artifactId>
+        <version>5.6.3.Final-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>org.jboss.narayana.jta</groupId>
+        <artifactId>tomcat-jta</artifactId>
+        <version>5.6.3.Final-SNAPSHOT</version>
+      </dependency>
```
psiroky: We really should not be depending on 3rd party SNAPSHOTs.

(We should also move these deps into kie-parent, but that can be done later).

MarianMacik: +1

jbpm-container-test/jbpm-remote-ejb-test/jbpm-remote-ejb-test-app/src/main/resources/META-INF/persistence.xml
MarianMacik: Thanks ;)

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/factory/InMemorySessionFactoryLeakTest.java
```diff
@@ -79,14 +83,18 @@ public InMemorySessionFactoryLeakTest(String strategy) {
...
         Properties properties= new Properties();
         properties.setProperty("mary", "HR");
         properties.setProperty("john", "HR");
         userGroupCallback = new JBossUserGroupCallbackImpl(properties);
         
         createRuntimeManager();
+
+        new InitialContext().rebind("java:comp/UserTransaction", com.arjuna.ats.jta.UserTransaction.userTransaction());
+        new InitialContext().rebind("java:comp/TransactionManager", TransactionManager.transactionManager());
+        new InitialContext().rebind("java:comp/TransactionSynchronizationRegistry", new com.arjuna.ats.internal.jta.transaction.arjunacore.TransactionSynchronizationRegistryImple());
```
mswiderski: these should not be needed, I am taking a look at this test now

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/DeadlinesBaseTest.java
tomjenkinson: OK, I revert that part

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/concurrent/SessionTest.java
tomjenkinson: Sure, I have trimmed this now

pom.xml
tomjenkinson: +1 - I can do a release of Narayana once we all agree that is all that is outstanding

jbpm-audit/pom.xml
tomjenkinson: +1 - I am not sure how I missed that

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/factory/InMemorySessionFactoryLeakTest.java
tomjenkinson: great - it was a new one added recently and seemed to need the transaction manager available for some reason (or maybe it was TSR)

mswiderski:
@tomjenkinson sent new PR for some additional updates, I think these are the last :)

jbpm-examples/pom.xml
```diff
@@ -94,6 +94,10 @@
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-test</artifactId>
     </dependency>
+      <dependency>
+          <groupId>org.jbpm</groupId>
+          <artifactId>jbpm-test-util</artifactId>
+      </dependency>
```
MarianMacik: Is it really needed here? I don't see any Pooling Data Source in this module. But maybe it is here because of something else...

MarianMacik: CloseSafeMemoryContextFactory isn't used either.

MarianMacik: Ok, it is needed since it is used in org/jbpm/examples/checklist/impl/CheckListUI.java but I was unable to find it since org.jbpm.test.util.PoolingDataSource is return type of JBPMHelper.setupDataSource() and the return value is ignored so it isn't used anywhere. But maven enforcer catches this as a direct usage. So this is OK.

jbpm-human-task/jbpm-human-task-workitems/pom.xml
```diff
@@ -95,6 +100,11 @@
 	  <artifactId>jbpm-human-task-jpa</artifactId>
 	  <scope>test</scope>
 	</dependency>
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-persistence-jpa</artifactId>
+      <scope>test</scope>
+    </dependency>
```
MarianMacik: @tomjenkinson Why was this dependency added? Thanks.

tomjenkinson: some are added for the transient enforcer check - I will remove it and run a build and see if it's still required

jbpm-human-task/jbpm-human-task-jpa/pom.xml
```diff
@@ -75,6 +75,11 @@
     </dependency>
 
     <!-- TEST -->
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-test-util</artifactId>
+      <scope>test</scope>
+    </dependency>
```
MarianMacik: This shouldn't be needed here since there are no tests.

jbpm-human-task/jbpm-human-task-workitems/pom.xml
tomjenkinson: seems OK without it

jbpm-human-task/jbpm-human-task-jpa/pom.xml
tomjenkinson: This was picked up due to a search and replace: https://github.com/kiegroup/jbpm/tree/master/jbpm-human-task/jbpm-human-task-jpa/src/test/resources

If you are sure it's not used I will remove the src/test/resources folder and the pom updates

tomjenkinson: It's here: https://github.com/kiegroup/jbpm/pull/898/commits/e1683e403aff2063ab5026e25039b0e2d4486ee7

MarianMacik: Thanks, @mswiderski can you confirm it is OK?

mswiderski: that seems to be correct, there are no tests at all in that module so having dependencies and configuration for transaction does not make sense.

pom.xml
tomjenkinson: IMO this is the only thing we need now - once you can confirm I will do the Narayana release and roll this to non-SNAPSHOT version

mswiderski:

pom.xml
mswiderski: correct, I would just wait for the patch Karel (@sutaakar) was about to send for MariaDB. Once that is all in we should be good to release narayana and merge this.

@MarianMacik any other comments regarding tests on different containers/dbs or we are good there too?

mswiderski:
I would say best to squash everything into single commit and use this jira (both number and title) as commit message - https://issues.jboss.org/browse/JBPM-6171

mswiderski:
@tomjenkinson you've done the most of the work so all credits go to you :)

MarianMacik:
OK, I think it is good to merge now. The only change with container tests was with tomcat8 and it works OK. After merge we will reconfigure our Cargo automation with @sutaakar so there might be some little changes in case we want to have 1:1 Cargo config between internal and community automation.

Just a side question - @tomjenkinson why is there jbossts-properties.xml file in some modules and in others it is not?

pom.xml
MarianMacik: @mswiderski See https://github.com/kiegroup/jbpm/pull/898#issuecomment-315025696

mswiderski:
@tomjenkinson this looks ready to be merged - so I think it's time to release narayana and update the version in main pom.xml. Once that is done we'll trigger last full build and merge.

mswiderski:

mswiderski:

mswiderski:

MarianMacik:
Thanks, @tomjenkinson. Good job!

PR #908, opened 2017-07-11
DROOLS-1629 - Allow multiple ObjectMarshallingStrategy of the same class

mswiderski:


hasys:

hasys:

mswiderski:
I think it can be merged without full downstream as it only changes tests so it has no way to affect any other repo

hasys:
@mswiderski it is more about the check that it will fix Full build :)

PR #910, opened 2017-07-12
RHBPMS-4811 - Improve jbpm services to allow its usage in spring with…

mswiderski:
… resource local


mswiderski:
Jenkins retest this

PR #912, opened 2017-07-17
JBPM-6171 - Replace bitronix with Narayana for jBPM - additional chan…

mswiderski:
…ges to allow it to be used for kie server and spring tests

depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/471


mswiderski:
@MarianMacik do you mind taking a look at this?

PR #916, opened 2017-07-19
[6.5.x] RHBPMS-4815 - Escape commas in LDAPUserGroupCallbackImpl filters

mswiderski:


mswiderski:
@rsynek it was mainly due to cherry pick that turned out to not be so clean due to many differences in LDAPUserGroupCallbackImpl - master vs 6.5.x

PR #919, opened 2017-07-21
jbpm-in-container-test: Add support for production DBs with Narayana on Tomcat

MarianMacik:
Hi, @mswiderski and @tomjenkinson,

here is the PR which adds support for production DBs with Narayana on Tomcat. There is some refactoring too because I noticed that some methods are used only by TransactionsTest class, so I moved them there.

Can you please review?

Thanks,

Marian


jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/org/jbpm/test/container/archive/localtransactions/tomcat-context.xml
```diff
@@ -13,8 +13,9 @@
 
     <!-- Database resources -->
     <Resource name="rawDataSource" uniqueName="rawDataSource" auth="Container" type="${org.jbpm.datasource.driver.class}" username="${org.jbpm.datasource.username}"
-              user="${org.jbpm.datasource.username}" password="${org.jbpm.datasource.password}" url="${org.jbpm.datasource.connection.url}" description="Raw Data Source" loginTimeout="0"
-              factory="${org.jbpm.datasource.driver.factory}"/>
+              user="${org.jbpm.datasource.username}" password="${org.jbpm.datasource.password}" url="${org.jbpm.datasource.connection.url}" URL="${org.jbpm.datasource.connection.url}" description="Raw Data Source" loginTimeout="0"
+              factory="${org.jbpm.datasource.driver.factory}" portNumber="${org.jbpm.datasource.port}" serverName="${org.jbpm.datasource.hostname}" databaseName="${org.jbpm.datasource.dbName}" testOnBorrow="false"
+              schema="${org.jbpm.persistence.schema}" driverType="4" />
```
tomjenkinson: I don't 100% follow all these settings - I guess it is because of the other DB types. I wonder how setDriverType works on a Resource that doesn't have that option - perhaps Tomcat tests for each attribute before setting it or passes them as properties? I guess your testing showed that to be the case.

MarianMacik: Hi,

driverType=4 is because of IBM DB2 10.5 database. From my understanding it switches the mode in which the DB driver operates. And yes, I think that Tomcat passes them as properties to the driver, e.g. Sybase 16.0 wasn't working until there was a portNumber property set. So it was quite a ride to make all this DBs working. But fortunately, we don't have to user separate file for each DB, it seems that every driver just reads properties it needs. Is this explanation sufficient?

tomjenkinson: Yes - thanks!

MarianMacik:
I think we can merge it now. Thanks for reviewing.

PR #939, opened 2017-08-09
RHBPMS-4857 - Completing task with empty map overrides the current ou…

mswiderski:
…tput variables

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
@@ -32,6 +32,7 @@ import java.util.Map;
 import org.assertj.core.api.Fail;
 import org.jbpm.services.task.events.DefaultTaskEventListener;
 import org.jbpm.services.task.exception.PermissionDeniedException;
+import org.jbpm.services.task.impl.TaskContentRegistry;
 import org.jbpm.services.task.impl.factories.TaskFactory;
 import org.jbpm.services.task.impl.model.xml.JaxbContent;
 import org.jbpm.services.task.utils.ContentMarshallerHelper;
@@ -47,6 +48,7 @@ import org.kie.api.task.model.Status;
 import org.kie.api.task.model.Task;
 import org.kie.api.task.model.TaskSummary;
 import org.kie.api.task.model.User;
+import org.kie.internal.task.api.ContentMarshallerContext;
 import org.kie.internal.task.api.EventService;
 import org.kie.internal.task.api.TaskModelProvider;
 import org.kie.internal.task.api.model.AccessType;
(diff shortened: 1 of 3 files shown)


mswiderski:
Jenkins retest this

tsurdilo:
Jenkins retest please

tsurdilo:
@mswiderski very nice

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
```diff
@@ -2388,7 +2390,9 @@ public void beforeTaskStartedEvent(TaskEvent event) {
 
             @Override
             public void beforeTaskCompletedEvent(TaskEvent event) {
-                assertNull(event.getTask().getTaskData().getTaskInputVariables());                
+                assertNotNull(event.getTask().getTaskData().getTaskInputVariables());
+                assertEquals(1, event.getTask().getTaskData().getTaskInputVariables().size());
+                assertTrue(event.getTask().getTaskData().getTaskInputVariables().containsKey("input"));
```
MarianMacik: Why is it now not null? Because of context.loadTaskVariables(task); in CompleteTaskCommand? Isn't it a change in behaviour so a doc update is needed? Same for afterTaskCompletedEvent.

MarianMacik:
2 failing tests test the same behaviour but inside a process, so please update them and update the comment https://github.com/kiegroup/jbpm/blob/master/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/HumanTaskVariablesAccessTest.java#L131 as well. Thanks!

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
mswiderski: correct, due to complete task command is loading variables they will already be available in the beforeTaskCompleted event. I don't think this requires docs update unless it's already there saying there won't be any data. Anyway, this is not guaranteed as one listener can load it meaning all other will have the data in place.

mswiderski:
@MarianMacik thanks, tests updated

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/LifeCycleBaseTest.java
MarianMacik: OK, I think you have a point there. But I think I don't understand the last sentence. So this listener is not guaranteed to be triggered so there can be situations which data are not loaded in?

mswiderski: what I meant is that having empty task input/output data was not guaranteed as there might be multiple listeners registered, and once one of them loads the data it will be then non empty for all subsequent listeners

MarianMacik: OK, clear now. Thanks!

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #940, opened 2017-08-10
JBPM-5708 MapDB integration proposal - being able to partially reuse …

mswiderski:
…services


mswiderski:
this PR does not need full down stream as the only changes in there are method/field visibility

manstis:
@mswiderski TBH I thought the downstream build stuff had been disabled?! @psiroky?

mswiderski:
@manstis it does not seem so, still being triggered after successful base build

PR #951, opened 2017-08-21
JBPM-6235 - Investigate NoSQL integration with drools/jbpm

mswiderski:


mswiderski:
@tsurdilo here is the base impl to enable to build custom EventEmitter implementations for different NoSQL stores.

mswiderski:
and here is the base implementation that uses ElasticSearch as storage for improved searhability across several kie servers https://github.com/kiegroup/droolsjbpm-contributed-experiments/pull/7

tsurdilo:
this is very useful @mswiderski !

mswiderski:
@tsurdilo let's keep it open for a while to work on different use cases for this integration and see if it covers all or we need further improvements.

jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/processinstance/objects/TestEventEmitter.java
```diff
@@ -0,0 +1,73 @@
...
+import org.jbpm.persistence.api.integration.base.BaseEventCollection;
+
+
+public class TestEventEmitter implements EventEmitter {
+
+
+    private static Set<InstanceView<?>> events = new LinkedHashSet<>();
+    
+    @Override
+    public void deliver(Collection<InstanceView<?>> data) {
```
domhanak: @mswiderski Could you specify a use case for this method? When do I want to implement it?

mswiderski: Take a look at Jira where I added two links that should give you some answers

domhanak: Thanks, it helped a lot. I'll plan some jira's for integration tests.

mswiderski:
@krisv this is now up to date and configured to not load task variables unless there is an actual event emitter available so this does not have any impact on performance any more. Tests where modified as they do load "test" emitter and thus had to be altered.

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -126,6 +130,9 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
                 }  
                 TransactionManagerHelper.addToUpdatableSet(txm, processInstanceInfo);
                 processInstanceInfo.updateLastReadDate();
+                
+
+                EventManagerProvider.getInstance().get().update(new ProcessInstanceView(processInstance));
```
MarianMacik: Just a question on how update of ProcessInstanceInfo works. I see that it is added to updatable set and then somehow I guess a transform method is called and it is persisted, right? Does this happen automatically with any ProcessInstance change after every commit? I was quite surprised that actually a getProcessInstance method updates a ProcessInstance in DB. And PersistenceEventManager will be notified about update only when getProcessInstance is called which may be much later I guess. Imagine that the ProcessInstance is updated by the engine, e.g. Human Task is encountered. Now we wait for a Human Task to be completed, let's say it is completed after 5 minutes - I guess only then is getProcessInstance called again so for the PersistenceEventManager  the process is updated only after that, right? But I am only guessing how this works, of course :)

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskContentServiceImpl.java
```diff
@@ -93,7 +94,10 @@ public long addOutputContent(long taskId, Map<String, Object> params) {
         }
         taskEventSupport.fireBeforeTaskOutputVariablesChanged(task, context, initialContent);
         ((InternalTaskData)task.getTaskData()).setTaskOutputVariables(params);
+                
+        ((InternalTaskData)task.getTaskData()).setTaskOutputVariables(params);        
         taskEventSupport.fireAfterTaskOutputVariablesChanged(task, context, params);
+        persistenceContext.updateTask(task);
```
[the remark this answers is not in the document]
mswiderski: this is not really needed for JPA as it's anyway "recorded" in persistence context but since the integration with other data stores is done through persistence layer it needs to be there for consistency reason.

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/base/BaseEventCollection.java
```diff
@@ -0,0 +1,56 @@
...
+    private Set<InstanceView<?>> events = new LinkedHashSet<>();
+    
+    @Override
+    public void update(InstanceView<?> item) {
+        this.events.add(item);
+    }
+    
+    @Override
+    public void remove(InstanceView<?> item) {
+        this.events.add(item);
```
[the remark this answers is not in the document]
mswiderski: fixed

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/base/TransactionalPersistenceEventManager.java
```diff
@@ -0,0 +1,151 @@
...
+    
+    public boolean isActive() {
+        return emitter != null;
+    }
+    
+    protected EventCollection getCollection() {
+        EventCollection collection = (EventCollection) this.tm.getResource(EVENT_COLLECTION);
+        if (collection == null) {
+            collection = emitter.newCollection();
+            this.tm.putResource(EVENT_COLLECTION, collection);
```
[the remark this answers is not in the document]
mswiderski: yes, this is the whole point of this EventManager - to keep stuff in transaction and emit them at the transaction completion if and only if it was successfully completed

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
mswiderski: update to process instance is done only once upon transaction completion (using TransactionSychronization.beforeCompletion). This is when every process instance that was loaded (not in read only mode) will be persisted back to data base - and this is retrieved from updatable set. So getProcessInstance is not updating anything, it's just marking it to be persisted when tx completes.

MarianMacik: OK, understand now. So EventManager is just updated with a new ProcessInstanceView of a processInstance and since it is only a pointer, any change made to the processInstance after this method is called will be still recorded, right?

mswiderski: yes, ProcessInstanceView has method copyFromSource and that is when it actually transforms the source (process instance) into process instance view. And this is called when collection is about to be delivered to emitter https://github.com/kiegroup/jbpm/pull/951/files#diff-ec43d650e5daf7c5415d81f2783a3f06R115

mswiderski:
Jenkins retest this

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/base/BaseEventCollection.java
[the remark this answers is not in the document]
mswiderski: re-fixed it :) it was actually as it should. This collection records all operations though it will avoid  duplicates. The reason for it is if the process instance was started and finished with single transaction it will be removed - but in context of event manager if we would truly remove it then event manager wouldn't know that such instance was created and thus wouldn't record it at all. So the responsibility with what to do with such InstanceView is on the actual emitters some might store it some might discard instances that are already done.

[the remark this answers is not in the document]
MarianMacik: Understand, it is up to users to alter this as they wish. But at the first sight it was really weird :)

mswiderski:
Jenkins retest this

jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/persistence/JPATaskPersistenceContext.java
```diff
@@ -128,20 +129,27 @@ public Task persistTask(Task task) {
         	this.em.flush();
             return this.em.find(TaskImpl.class, task.getId(), LockModeType.PESSIMISTIC_FORCE_INCREMENT );
```
rbarriuso, line 130: Now I see this: Shouldn't `create()` be called if `pessimisticLocking` is enabled too?

MarianMacik: It seems to be you are right. `create()` should be called before the if.

PR #952, opened 2017-08-21
Add a check for non-NULL key

izodev:
In a situation when we complete a task by the kie-client-rest Api, and there is a bidirectional relations between two entities. We have to use @XmlID, @XmlAttribute and @XmlIDREF , but we don't need really to persist both variables. So as a work-around we can send one of them with NULL key in the task's output map and then my update will ignore it and persist only the desired variable. Otherwise we will get an exception of org.hibernate.MappingException: Unknown entity during the unmarshalling.


mswiderski:
could you please open jira for this issue and provide a test case for this change so it can be accepted.

PR #953, opened 2017-08-22
JBPM-6225 - Explicit Accept Header on RestWorkItemHandler

tsurdilo:


tsurdilo:
also removes inner TestWorkItemManager to use the already existing one.

PR #959, opened 2017-08-25
JBPM-6309: Investigate and fix jBPM community DB tests with Narayana

domhanak:
Surrounds "setURL" method invocation on db driver with try-catch and tries to use "setUrl" method in case no "setURL" method is present when initializing datasource. (Some drivers have setUrl some setURL)
Catch InvocationTargetException and log it.

Adds try-catch blocks around DB operation calls. Some operations throw exceptions when DB configuration is bad and these exception are hiding the configuration problem.
Sets _logLevel_ to _warn_ for narayana for some modules, since some relevant information wasn't being logged on _error_ level.

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/correlation/CorrelationPersistenceTest.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/correlation/CorrelationPersistenceTest.java
@@ -36,6 +36,7 @@ import javax.transaction.UserTransaction;
 
 import org.jbpm.test.util.AbstractBaseTest;
 import org.junit.After;
+import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 import org.junit.runner.RunWith;
@@ -68,9 +69,8 @@ public class CorrelationPersistenceTest extends AbstractBaseTest {
         EntityManagerFactory emf = (EntityManagerFactory) context.get(EnvironmentName.ENTITY_MANAGER_FACTORY);
         UserTransaction ut = InitialContext.doLookup("java:comp/UserTransaction");
         ut.begin();
-        EntityManager em = emf.createEntityManager();
-
         try {
+            EntityManager em = emf.createEntityManager();
             em.persist(factory.newCorrelationKey("test123"));
 
             List<String> props = new ArrayList<String>();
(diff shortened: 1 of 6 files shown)


domhanak:
@MarianMacik @mswiderski  Please take a look. I'll squash and create pull-request for 7.3.x when approved.
@baldimir Should I add ?pinGlobalTxToPhysicalConnection=true into mariadb profile?
https://github.com/kiegroup/jbpm/blob/c60cbeafac4e96d1879496f87d0a576dc719ff53/pom.xml#L225

MarianMacik:
Looks good to me and if ?pinGlobalTxToPhysicalConnection=true property makes mariadb work with Narayana (as you have just found out on our Jenkins) then there is no doubt I would add it to a mariadb profile.

domhanak:
@mswiderski I noticed that there are several definition of <maven.jdbc.url> in _jbpm/<module>/pom.xml_ that override properties from _jbpm/pom.xm_ when not specified as command line arguments. 
https://github.com/kiegroup/jbpm/blob/f2dee0ad116c847a2ed436e80dde43d2ceea9be5/jbpm-bpmn2/pom.xml#L22
Is there any specific purpose for them? They make underlying tests fails when you run them locally with mariadb profile as they override the correct jdbc.url. Can they be removed?

mswiderski:
I believe these were supposed to be defaults so all other pom.xmlshould override them

domhanak:
@tsurdilo could you please review and accept-merge/reject

jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/correlation/CorrelationPersistenceTest.java
```diff
@@ -70,13 +70,18 @@ public void before() throws Exception {
...
-        em.persist(factory.newCorrelationKey(props));
-        
+            List<String> props = new ArrayList<String>();
+            props.add("test123");
+            props.add("123test");
+
+            em.persist(factory.newCorrelationKey(props));
+        } catch (Exception ex) {
+            ex.printStackTrace();
+            ut.rollback();
```
krisv: In case there is an exception, wouldn't this let the test pass?  shouldn't we add a fail here?

jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/correlation/CorrelationPersistenceTest.java
```diff
@@ -87,8 +92,13 @@ public void after() {
...
             ut.begin();
             EntityManager em = emf.createEntityManager();
-            em.createQuery("delete from CorrelationPropertyInfo").executeUpdate();
-            em.createQuery("delete from CorrelationKeyInfo").executeUpdate();
+            try {
+                em.createQuery("delete from CorrelationPropertyInfo").executeUpdate();
+                em.createQuery("delete from CorrelationKeyInfo").executeUpdate();
+            } catch (Exception ex) {
+                ex.printStackTrace();
+                ut.rollback();
```
krisv: Same here

jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/correlation/CorrelationPersistenceTest.java
domhanak: It should not, because there is  ut.commit() right after the try-catch block, which will fail as there is no active transaction due to ut.rollback() call in catch block. But, I agree that Assert.fail(); is better option here.

rsynek: @domhanak could you please also rather use logging instead of ex.printStackTrace() ?

baldimir: I think the whole transaction should be in try-catch block. See here: https://gist.github.com/baldimir/334084038b994ceb25c5b09d2cedb1ec

When anything fails after transaction began, it must be rolled back.

domhanak:
@baldimir @krisv @rsynek rebased and incorporated your feedback.

PR #964, opened 2017-08-31
JBPM-6309: Additional fixes for DB tests with narayana

domhanak:
There is an additional problem that appears when running with db2 and
sybase - they have no setURL/setUrl methods, so a guard is added to skip
this piece of code. These DB's driver asks for url in connection so as long as it
is in driverProperties it should be set.

Also adds url property into driverProperties for all
databases.


domhanak:
@tomjenkinson  I still have a problem that DB2 doesn't set driverType=4 and is still using 2, even after it gets it in properties. To me it look like it is being set to 2, when newInstance() method of driver is invoked. Any ideas on how to correctly set it? Documentation of db2 suggest setting <xa-datasource-property ... > but do not have anything like that present yet.

domhanak:
@tomjenkinson it is there, but properties do not apply into driver - I need to invoke drivers setDriverType in order for it to get set
`xads.getClass().getMethod("setDriverType", new Class[]{int.class}).invoke(xads, new Object[]{4});`

I'll get more data an write on the mailing list.

domhanak:
@tomjenkinson Yes I managed to add that logic locally and now all seems to be set in driver and tests begin to setup, however they fail when creating session. What is strange is that I need to invoke all methods for DB2 = "setDriverType", "setPortNumber", "setDatabaseName" in order to get the connection and just passing the property file does not seem to work?
One more thing, what I don't like is that eventually, when we move to java 9 this might be not working as we do not know if vendors are going to support this.
if vendors will allow reflection on their **drivers.**

PR #968, opened 2017-09-04
RHBPMS-4844 - 'getProcessInstancesByVariableNameAndValue' does not wo…

mswiderski:
…rk with multiple process variables

this includes just a test case as fix was already done prior v7


domhanak:
@mswiderski just a curious question, you mentioned "fix was already done prior v7" Does that mean the fix is on 7.3.x branch aswell? If that is the case shouldn't we add this there too?

mswiderski:
yes, fix is already on 7.3.x too, I don't think we need to add anything on 7.3.x as it was already built for release.

domhanak:
Sure, I agree but since this is adding a test-case it would be nice to see it pass with 7.3.x too and I believe we still want bugfixes there right?

mswiderski:
@domhanak I think only those that should go to LA anyway, I will prepare back port, don't think it will hurt :)

mswiderski:
here it is https://github.com/kiegroup/jbpm/pull/972

PR #970, opened 2017-09-05
JBPM-6309: Fix jBPM community DB tests with Narayana

domhanak:
This adds additional method that handles setting up properties
for Sybase and DB2. For DB2 there are 2 specific properties
and Sybase requieres username and password setup before
actuall getConnection() call.
Removes redundant array creations and enhances logging.

@baldimir @MarianMacik  This should be the last change required for tour matrix jobs to work.

@mswiderski could you please review?


mswiderski:
@domhanak in general looks ok, just one question - port numbers - should they be hardcoded?

domhanak:
@mswiderski No that is a mistake, let me correct it

PR #975, opened 2017-09-06
[7.3.x] JBPM-6309: Fix incorrect type in method invocation

domhanak:
cherry-pick of #974


domhanak:
this is a fix for DB tests

PR #976, opened 2017-09-07
Remove dead code in BusinessCalendarImpl

muff1nman:
Since the #init method doesn't allow a null `businessCalendarConfiguration`, any checks against `null` will never pass.


mswiderski:
Ok to test

mswiderski:
+1

PR #979, opened 2017-09-11
[6.5.x] RHBPMS-4778 - Ability to specify priority for Asynchronous Tasks (#450)

mswiderski:


jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorImpl.java
```diff
@@ -320,6 +320,20 @@ public Long scheduleRequest(String commandId, Date date, CommandContext ctx) {
         } else {
             requestInfo.setRetries(retries);
         }
+        int priority = 5;
```
rsynek: Is "5" a default priority? If so, it should be defined in a named constant. The same applies to lower and upper priority boundaries several lines below. Let's avoid magic constants.

mswiderski: @rsynek I merged it as is to be inline with what is on 7 already. Do you  mind opening jira for 7 to accommodate these comments?

rsynek: @mswiderski well, here is it https://issues.jboss.org/browse/JBPM-6430. On the other hand, these are so small changes that would be better to address them directly next time.

PR #984, opened 2017-09-18
JBPM-6417 - Kie Server rest Signals API does not resolve mvel express…

tsurdilo:
…ions


tsurdilo:
@mswiderski i changed back the method visibility update. For the other small updates I left them because think they are useful to have. If you rather have those as part of a separate Jira or not at all let me know and ill change that as well. Thanks!

mswiderski:
@tsurdilo fine with me, though these changes might make it harder to back port to 6.5.x but that's minor thing.

PR #990, opened 2017-09-19
JBPM-6425 - Workitem Archetype - generated .wid file should use name …

tsurdilo:
…based on user project


tsurdilo:
@mswiderski just adds annotation processor argument so that the generated .wid name can be based on the user project (not hard-coded)

PR #991, opened 2017-09-19
[7.3.x] JBPM-6425 - Workitem Archetype - generated .wid file should use name …

tsurdilo:
…based on user project

### DIFF
--- a/jbpm-workitems/jbpm-workitems-core/src/main/java/org/jbpm/process/workitem/core/util/WidProcessor.java
+++ b/jbpm-workitems/jbpm-workitems-core/src/main/java/org/jbpm/process/workitem/core/util/WidProcessor.java
@@ -77,7 +77,6 @@ public class WidProcessor extends AbstractProcessor {
             "}; separator=\",\"$\n" +
             "]";
 
-
     public boolean process(final Set<? extends TypeElement> annotations,
                            final RoundEnvironment roundEnv) {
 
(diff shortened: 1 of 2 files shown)


jbpm-workitems/jbpm-workitems-core/src/main/java/org/jbpm/process/workitem/core/util/WidProcessor.java
```diff
@@ -119,9 +118,17 @@ public boolean postProcessWorkItemDefinition() {
             widTemplate.add("widInfo",
                             wrappedResults);
 
+            String widName = "WorkDefinitions.wid";
+            if (processingEnv.getOptions().containsKey("widName")) {
+                widName = processingEnv.getOptions().get("widName");
+            } else {
+                processingEnv.getMessager().printMessage(Diagnostic.Kind.NOTE,
+                                                         "Unable to find option \"widName\", using default.");
+            }
```
triceo: Assuming that `getOptions()` returns a `Map`, the JDK can do this for you:

    String widName = processingEnv.getOptions().computeIfAbsent("widName", key -> {
        processingEnv.getMessager().printMessage(Diagnostic.Kind.NOTE, "Unable to find option \"widName\", using default.");
        return "WorkDefinitions.wid";
    });

Does the same thing, uses less code.

tsurdilo: @triceo sorry cant apply this because it returns instance of UnmodifiableMap and with that computeIfAbsent throws an UnsupportedOperationException. Will keep it in mind for next time i use this - thanks!!!

PR #992, opened 2017-09-20
[7.3.x] RHBPMS-4950 - UpdateTimerCommand does not work for timers in embedded…

mswiderski:
… sub process

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/TimerUpdateTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/TimerUpdateTest.java
@@ -25,6 +25,7 @@ import org.jbpm.test.listener.CountDownProcessEventListener;
 import org.junit.Before;
 import org.junit.Test;
 import org.kie.api.event.process.DefaultProcessEventListener;
+import org.kie.api.event.process.ProcessNodeLeftEvent;
 import org.kie.api.event.process.ProcessStartedEvent;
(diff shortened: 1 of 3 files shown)


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/TimerUpdateTest.java
```diff
@@ -192,6 +197,45 @@ public void beforeProcessStarted(ProcessStartedEvent event) {
...
+        kieSession.addEventListener(new DefaultProcessEventListener() {
+            @Override
+            public void beforeProcessStarted(ProcessStartedEvent event) {
+                list.add(event.getProcessInstance().getId());
+            }
+            @Override
+            public void afterNodeLeft(ProcessNodeLeftEvent event) {
+                if (TIMER_SUBPROCESS_NAME.equals(event.getNodeInstance().getNodeName())) {
+                    System.setProperty(TIMER_FIRED_TEXT, "");
+                    System.setProperty(TIMER_FIRED_TIME_PROP, System.currentTimeMillis()+"");
```
MarianMacik: Everything looks great but I was thinking a little bit if there is something missing or not in "" string so I would rather prefer String.valueOf() or Long.toString() so there is no confusion. Otherwise than that it looks OK.

mswiderski: fixed

PR #995, opened 2017-09-21
JBPM-6429 - Workitem Archetype - allow @Wid parameter inheritance fro…

tsurdilo:
…m handler interfaces


tsurdilo:
This allows users to "inherit" (merge) their @Wid information from the interface if their handler. For example if you have

```
@Wid(....)
public interface MyHandlerInterface {...}
```

and 

```
@Wid 
public class MyTestHandler implements MyHandlerInterface {...}
```

the resulting .wid file of assembly will be a merge of the MyHandlerInterface and MyHandlerInterface @Wid info. 
Rule is that the concrete class info will always overwrite the interface info for the wid.

PR #997, opened 2017-09-22
JBPM-6431 moving syn method to block further down

ianwatsonrh:


mswiderski:
ok to test

mswiderski:
@ianwatsonrh thanks

PR #1000, opened 2017-09-25
JBPM-6434 - ProcessInstance is not correctly persisted when loaded wi…

tkobayas:
…th readOnly = true early in the same transaction

- Unit test

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
@@ -142,7 +142,7 @@ public class JPAProcessInstanceManager
             return null;
         }
         processInstance = (org.jbpm.process.instance.ProcessInstance)
-        	processInstanceInfo.getProcessInstance(kruntime, this.kruntime.getEnvironment());
+        	processInstanceInfo.getProcessInstance(kruntime, this.kruntime.getEnvironment(), readOnly);
         if (!readOnly) {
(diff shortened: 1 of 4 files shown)


mswiderski:
Jenkins retest this

mswiderski:
@tkobayas congratulations at submitting pull request number 1000 to jBPM repository!!!

you have won a free fix for the submitted bug, enjoy!
🍰  🎈 

P.S.
I couldn't replace this nice PR so I updated it ;)

tkobayas:
yeaaaaaah  :laughing:  :laughing:  :laughing: :birthday: :birthday: :birthday:

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -160,7 +160,11 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
                 kruntime.getProcessInstance(parentProcessInstanceId);
             }
             processInstance.setKnowledgeRuntime( kruntime );
+            
             ((ProcessInstanceImpl) processInstance).reconnect();
+            if (readOnly) {
+                internalRemoveProcessInstance(processInstance);
```
MarianMacik: Just a short question - why is the process instance removed if it is read only and if I see correctly it is not added to the processInstance map anywhere in the code?

mswiderski: @MarianMacik because it should not be there, it should always be loaded from the store as it should not receive any updates as they won't be persisted.

PR #1001, opened 2017-09-27
Avoid generation of NPE while launching a Process without Parent

ngs-mtech:
The proposed code avoids the generation of NPE every time you launch a process with no parent.

The inboxing performed by the JVM to create a Long with a NULL value and the later outboxing to assign a long from a NULL Long value generates a NPE every time you launch a process without parent.

This is a performance fix, please apply it.

Regards

### DIFF
--- a/jbpm-audit/src/main/java/org/jbpm/process/audit/event/DefaultAuditEventBuilderImpl.java
+++ b/jbpm-audit/src/main/java/org/jbpm/process/audit/event/DefaultAuditEventBuilderImpl.java
@@ -54,18 +54,10 @@ public class DefaultAuditEventBuilderImpl implements AuditEventBuilder {
         if (correlationKey != null) {
         	log.setCorrelationKey(correlationKey.toExternalForm());
         }
-        try {
-            Object value = pi.getMetaData().get("ParentProcessInstanceId");
-	    if( value != null ){
-	        long parentProcessInstanceId = (Long) value;
-                log.setParentProcessInstanceId( parentProcessInstanceId );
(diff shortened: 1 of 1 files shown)


jbpm-audit/src/main/java/org/jbpm/process/audit/event/DefaultAuditEventBuilderImpl.java
```diff
@@ -55,8 +55,13 @@ public AuditEvent buildEvent(ProcessStartedEvent pse) {
...
         try {
-            long parentProcessInstanceId = (Long) pi.getMetaData().get("ParentProcessInstanceId");
-            log.setParentProcessInstanceId(parentProcessInstanceId);
+            Object value = pi.getMetaData().get("ParentProcessInstanceId");
+	    if( value != null ){
+	        long parentProcessInstanceId = (Long) value;
+                log.setParentProcessInstanceId( parentProcessInstanceId );
+	     }else{
+		log.setParentProcessInstanceId(-1L);
+	     }
```
mswiderski: wouldn't this be way easier to use a single line as follows instead of all these checks...

```
long parentProcessInstanceId = (Long) pi.getMetaData().getOrDefault("ParentProcessInstanceId", -1L);
```

ngs-mtech: Sounds perfect to me. I will apply it. Let me check!.

mswiderski:
Ok to test

PR #1006, opened 2017-10-03
JBPM-6459 - Update taskInstance attributes and variables inside the s…

mswiderski:
…ame REST API call


mswiderski:
Jenkins retest this

PR #1007, opened 2017-10-04
[DON'T MERGE] google calendar integration workitem

tsurdilo:
@mswiderski this will go into the jbpm-work-tems repo when its created. Just wanted to do pr here maybe if you have time take a look and let me know if you see any issues. 
I have turned off maven-enforcer-plugin for the gcal because this will not be merged, but it brings up a question if we want to use as strict enforcement rules as we do in other repos? Especially what I think will come up is transitive dependencies checks and version declaration checks. Google documentation on their api does not take in account transitive depends and just give you a single depends that users that create new workitems will surely use. In these cases users will have to do extra work to contribute back which might be a turnoff. WDYT?

### DIFF
--- a/jbpm-workitems/jbpm-workitems-google-calendar/.gitignore
+++ b/jbpm-workitems/jbpm-workitems-google-calendar/.gitignore
@@ -1,0 +1,12 @@
+/target
+/local
+/bin
+
+# Eclipse, Netbeans and IntelliJ files
+/.*
+!.gitignore
+/nbproject
+/*.ipr
+/*.iws
+/*.iml
+
--- a/jbpm-workitems/jbpm-workitems-google-calendar/pom.xml
+++ b/jbpm-workitems/jbpm-workitems-google-calendar/pom.xml
@@ -1,0 +1,85 @@
+<?xml version="1.0"?>
+<project
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd"
+    xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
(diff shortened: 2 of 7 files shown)


mswiderski:
@tsurdilo first thing that came to my mind was to split this single handler into several that will be dedicated for particular action - like handler for adding event and another one for adding calendar. That way the parameters will be much simpler and easier to use, most likely same case for results. Imagine business user to pick add event node from the palette instead of calendar and then need to figure out what params are needed for add event action. wdyt?

fine with me when it comes with transitive deps, @psiroky do you see any downsides of it?

psiroky:
@mswiderski any additional 3rd party dep means more maintenance for us (regular upgrading, looking for CVE issues, etc). We need to be sure that we are prepared to accept that. Especially if we start bundling this in our distros.

mswiderski:
@psiroky there is no intention to bundle that in any distributions. we plan to have these in service repository catalog at community. Long run there might be product decision to take some (that make sense from product point of view) and have product service repository so at that time we could have it then aligned with all policies. Does this make sense or is it completely insane? (be honest ;))

psiroky:
Makes sense.I am not saying we should not do this. Just trying to stress out that it's not just about adding the dependency and forgetting about it. It should be regularly updated and even if available just in community, we should make sure we are not depending on artifacts with knows vulnerabilities.

mswiderski:
@psiroky +1, thanks for your input

tsurdilo:
@mswiderski +1 on splitting up - will do. Also +1 for @psiroky . Thanks guys.

mswiderski:
@tsurdilo shall it be still open or can we close and move this to new repo - jbpm-work-items?

tsurdilo:
@mswiderski yes it will be in the new repo. finishing pr there. i will close this.

PR #1019, opened 2017-10-06
RHBPMS-4942 - LazyInitializationException when getting process variab…

tsurdilo:
…les which extends VariableEntity

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceWithEntitiesTest.java
+++ b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceWithEntitiesTest.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.kie.services.test;
+
+import java.io.File;
+import java.io.FileOutputStream;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.drools.compiler.kie.builder.impl.InternalKieModule;
+import org.drools.core.util.IoUtils;
+import org.drools.persistence.jpa.marshaller.MappedVariable;
+import org.drools.persistence.jpa.marshaller.VariableEntity;
+import org.hibernate.LazyInitializationException;
+import org.jbpm.kie.services.impl.KModuleDeploymentUnit;
+import org.jbpm.kie.test.util.AbstractKieServicesBaseTest;
+import org.jbpm.runtime.manager.impl.deploy.DeploymentDescriptorImpl;
+import org.jbpm.services.api.model.DeploymentUnit;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.kie.api.KieServices;
+import org.kie.api.builder.ReleaseId;
+import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.internal.runtime.conf.DeploymentDescriptor;
+import org.kie.internal.runtime.conf.ObjectModel;
+import org.kie.scanner.KieMavenRepository;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import static org.junit.Assert.*;
+import static org.kie.scanner.KieMavenRepository.getKieMavenRepository;
+
+public class ProcessServiceWithEntitiesTest extends AbstractKieServicesBaseTest {
+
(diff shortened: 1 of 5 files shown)


tsurdilo:
depends on drools pr https://github.com/kiegroup/drools/pull/1519

tsurdilo:
Jenkins retest please

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceWithEntitiesTest.java
```diff
@@ -0,0 +1,239 @@
...
+                "    <persistence-unit name=\"org.jbpm.test:test-module:1.0.0-SNAPSHOT\" transaction-type=\"JTA\">\n" +
+                "        <provider>org.hibernate.ejb.HibernatePersistence</provider>\n" +
+                "        <jta-data-source>java:jdbc/testDS1</jta-data-source>\n" +
+                "        <class>example.CaseDetail</class>\n" +
+                "        <class>org.drools.persistence.jpa.marshaller.MappedVariable</class>\n" +
+                "        <class>org.drools.persistence.jpa.marshaller.VariableEntity</class>\n" +
+                "        <exclude-unlisted-classes>true</exclude-unlisted-classes>\n" +
+                "        <properties>\n" +
+                "            <property name=\"hibernate.dialect\" value=\"org.hibernate.dialect.PostgreSQLDialect\"/>\n" +
```
MarianMacik: @tsurdilo I have one question. Why is this persistence.xml in String and not in a seperate resource file, moreover, is it mentioned to be run only with PostgreSQL?

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceWithEntitiesTest.java
```diff
@@ -0,0 +1,239 @@
...
+                "            <property name=\"hibernate.hbm2ddl.auto\" value=\"update\"/>\n" +
+                "            <property name=\"hibernate.show_sql\" value=\"false\"/>\n" +
+                "            <property name=\"hibernate.id.new_generator_mappings\" value=\"false\"/>\n" +
+                "            <property name=\"hibernate.transaction.jta.platform\" value=\"org.hibernate.service.jta.platform.internal.JBossAppServerJtaPlatform\"/>\n" +
+                "        </properties>\n" +
+                "    </persistence-unit>\n" +
+                "</persistence>";
+    }
+
+    private static String getCaseDetailEntitySource() {
```
MarianMacik: Can't this be in a seperate file for better control of the code so it is less error-prone?

tsurdilo:
@MarianMacik updated according to your comments - thanks! Note that the entity class used in test has to be on the kjar classloader (not the test classloader) so it has to be loaded in its source format.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceWithEntitiesTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        DeploymentDescriptorImpl customDescriptor = new DeploymentDescriptorImpl("org.jbpm.domain");
+        customDescriptor.setAuditMode(AuditMode.JPA);
+        customDescriptor.setEnvironmentEntries(null);
+
+        List<String> remotableClasses = java.util.Arrays.asList("example.CaseDetail", "org.drools.persistence.jpa.marshaller.MappedVariable");
+        customDescriptor.setClasses(remotableClasses);
+
+        List<ObjectModel> marshallingStrategies = new ArrayList<ObjectModel>();
+        marshallingStrategies.add(new ObjectModel("mvel", "new org.drools.persistence.jpa.marshaller.JPAPlaceholderResolverStrategy(\"org.jbpm.domain\", classLoader)"));
```
mswiderski: shouldn't this use persistence unit named org.jbpm.test:test-module:1.0.0-SNAPSHOT instead of org.jbpm.domain?

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceWithEntitiesTest.java
```diff
@@ -0,0 +1,165 @@
...
+        logger.debug("Preparing kjar");
+        KieServices ks = KieServices.Factory.get();
+        ReleaseId releaseId = ks.newReleaseId(GROUP_ID, ARTIFACT_ID, VERSION);
+        List<String> processes = new ArrayList<String>();
+        processes.add("repo/processes/general/entityprocessvar-process.bpmn2");
+
+
+        DeploymentDescriptorImpl customDescriptor = new DeploymentDescriptorImpl("org.jbpm.domain");
+        customDescriptor.setAuditMode(AuditMode.JPA);
+        customDescriptor.setEnvironmentEntries(null);
```
mswiderski: most likely not needed (audit mode is by default JPA and end entries don't have to be nulled)

tsurdilo:
@mswiderski updated - thanks!!

MarianMacik:
Looks good to me now. Just one question - why does entity class need to be in a .txt format? Can't we use normal .java file? By doing this it will be still loaded in a source format, but IDE will show it like a Java class.

tsurdilo:
@MarianMacik it has to be part of the kjar classloader - same as if user created a .java (still text) file in the workbench (source).

MarianMacik:
I understand but if it was .java - wouldn't it be still readable via IOUtils? And IDE would recognize it as a .java file. I admit that it will be on a normal test classloader in addition to kjar classloader, but is it an issue? Thanks.

PR #1022, opened 2017-10-09
JBPM-6463 - Clustered EJB timers start duplicated process instances w…

mswiderski:
…hen used inside startEvent

start event timers should always be checked thus can be marked as new timer as it will be then skipped by ejb timer service to check the already scheduled with same name.


mswiderski:
thanks @psiroky will prepare 7.4.x back port so it gets included in LA

PR #1023, opened 2017-10-09
JBPM-5651 - Add excluded owner task id filter into potential owner pr…

cristianonicolai:
…eprocessor

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java
+++ b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java
@@ -77,6 +77,7 @@ import static org.kie.scanner.KieMavenRepository.getKieMavenRepository;
 public class QueryServiceImplTest extends AbstractKieServicesBaseTest {
 
     private static final Logger logger = LoggerFactory.getLogger(KModuleDeploymentServiceTest.class);
+    private static final String PO_TASK_QUERY = "select ti.activationTime, ti.actualOwner, ti.createdBy, ti.createdOn, ti.deploymentId, " + "ti.description, ti.dueDate, ti.name, ti.parentId, ti.priority, ti.processId, ti.processInstanceId, " + "ti.processSessionId, ti.status, ti.taskId, ti.workItemId, oe.id, eo.entity_id " + "from AuditTaskImpl ti " + "left join PeopleAssignments_PotOwners po on ti.taskId = po.task_id " + "left join OrganizationalEntity oe on po.entity_id = oe.id " + " left join PeopleAssignments_ExclOwners eo on ti.taskId = eo.task_id ";
 
     private List<DeploymentUnit> units = new ArrayList<DeploymentUnit>();
     protected String correctUser = "testUser";
@@ -92,6 +93,9 @@ public class QueryServiceImplTest extends AbstractKieServicesBaseTest {
 
     @Before
     public void prepare() {
+        System.setProperty("org.jbpm.ht.callback", "custom");
+        System.setProperty("org.jbpm.ht.custom.callback", "org.jbpm.kie.services.test.objects.TestUserGroupCallbackImpl");
+
         this.dataSourceJNDIname = getDataSourceJNDI();
         configureServices();
         logger.debug("Preparing kjar");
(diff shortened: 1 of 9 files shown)


jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/query/QueryResultMapper.java
```diff
@@ -97,7 +97,8 @@
     public static final String COLUMN_TASKID = "TASKID";
     public static final String COLUMN_WORKITEMID = "WORKITEMID";
     public static final String COLUMN_ORGANIZATIONAL_ENTITY = "ID";
-    
+    public static final String COLUMN_EXCLUDED_OWNER = "ENTITY_ID";
```
mswiderski: this sound a bit too generic - while the name of the constant is clear the column name is not. Is this because of the underlying db column name and to avoid using aliases? or could we make it more clear...

cristianonicolai: @mswiderski that's correct, I just kept the column name as the database column to avoid using an alias.

cristianonicolai:
@mswiderski @bsremac added a test process where a task is assigned to a group and a specific user from that group is excluded via _ExcludedOwnerId_.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java
```diff
@@ -443,6 +451,53 @@ public void testGetTaskInstancesAsPotOwners() {
...
+        assertNotNull(registeredQuery);
+        assertEquals(query.getName(), registeredQuery.getName());
+        assertEquals(query.getSource(), registeredQuery.getSource());
+        assertEquals(query.getExpression(), registeredQuery.getExpression());
+        assertEquals(query.getTarget(), registeredQuery.getTarget());
+
+        processInstanceId = processService.startProcess(deploymentUnit.getIdentifier(), "org.jbpm.ExcludedOwner", new HashMap<String, Object>());
+        assertNotNull(processInstanceId);
+
+        identityProvider.setName("maciej");
```
bsremac: Maybe just for the sake of test readability, you could use named constants behind actual user Id's, along the lines of:
```String potentialOwner = "maciej"```
```String excludedOwner = "kris"```

File: jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java [hunk 3]
```diff
@@ -443,6 +451,53 @@ public void testGetTaskInstancesAsPotOwners() {
...
+
+        identityProvider.setName("kris");
+        identityProvider.setRoles(Arrays.asList("admins"));
+
+        taskInstanceLogs = queryService.query(query.getName(), UserTaskInstanceQueryMapper.get(), new QueryContext());
+        assertNotNull(taskInstanceLogs);
+        assertEquals(0, taskInstanceLogs.size());
+
+        processService.abortProcessInstance(processInstanceId);
+        processInstanceId = null;
```
bsremac: You don't have to clean up the reference at the test end, JUnit instantiates a new class for each ```@Test``` method.
If you want to keep consistency with the rest of the tests in the class, I guess you can leave it as it is but,just so you know the statement is redundant.

cristianonicolai: @bsremac it seems that _processInstanceId_ is used in the _cleanup_ method to abort the process instance in case is not null already.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java
cristianonicolai: done

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java
bsremac: I guess in this case you could relax the test from any cleanup steps after the last assertion and let the cleanup method do it's job :)

PR #1030, opened 2017-10-12
JBPM-6487 Jbpm tests stabilisation

dcihak:
Fixes IntermediateEventTest, TaskCommentTest and TaskExpirationDateTest

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
@@ -29,6 +29,7 @@ import java.util.Set;
 import javax.persistence.EntityManagerFactory;
 import javax.persistence.Persistence;
 
+import org.assertj.core.api.Assertions;
 import org.drools.core.command.SingleSessionCommandService;
 import org.drools.core.command.impl.CommandBasedStatefulKnowledgeSession;
 import org.drools.core.command.impl.ExecutableCommand;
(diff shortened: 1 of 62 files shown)


dcihak:
[DO NOT MERGE]

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java
```diff
@@ -1206,7 +1206,7 @@ public void testTimerBoundaryEventInterruptingOnTaskCancelTimer() throws Excepti
         ksession = restoreSession(ksession, true);
         ksession.getWorkItemManager().registerWorkItemHandler("Human Task", handler);
         timers = getTimerManager(ksession).getTimers();
-        assertEquals(0, timers.size());
+        assertTrue(timers == null || timers.isEmpty());
```
rsynek: I would rather avoid using assertTrue (if you are not asserting a value of boolean variable, of course). Reason is that when assertTrue fails, it does not tell you why - was is null or empty? Please take a look at assertJ library, is provides decent asserts for (not only) collections.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/TaskCommentTest.java
```diff
@@ -43,9 +30,19 @@
 import org.kie.internal.task.api.model.InternalComment;
 import org.kie.internal.task.api.model.InternalOrganizationalEntity;
 
+import javax.persistence.EntityManagerFactory;
```
rsynek: Is this change of imports based on KIE group code style? If not, I would rather avoid unnecessary formatting changes like these.

dcihak:
Imports formatting and empty collection assert fixed. If OK, READY TO BE MERGED.

PR #1034, opened 2017-10-18
Initial implementation of JBPM-6502. Support collection of documents.

DuncanDoyle:
Documents is a new type that wraps a List<Document>. The new MarshallingStrategy re-used the existing DocumentMarshallingStrategy for individual document persistence.

### DIFF
--- a/jbpm-document/src/main/java/org/jbpm/document/Documents.java
+++ b/jbpm-document/src/main/java/org/jbpm/document/Documents.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.document;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * A collection of {@link Document Documents}.
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlRootElement(name = "documents-object")
+public class Documents implements Serializable {
+
+	private static final long serialVersionUID = 6962662228758156488L;
+	
+	private List<Document> documents = new ArrayList<>();
+
+	public Documents() {
+	}
+
+	public Documents(List<Document> documents) {
+		this.documents = new ArrayList<Document>(documents);
+	}
+	
+	public void addDocument(Document document) {
+		documents.add(document);
+	}
+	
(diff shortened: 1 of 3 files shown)


jbpm-document/src/test/java/org/jbpm/document/marshalling/DocumentsMarshallingStrategyTest.java
```diff
@@ -0,0 +1,113 @@
...
+		byte[] marshalledDocument = docMarshallingStrategy.marshal(null, null, document);
+		Document unmarshalledDocument = (Document) docMarshallingStrategy.unmarshal(null, null, marshalledDocument, this.getClass().getClassLoader());
+	
+		assertEquals(document.getName(), unmarshalledDocument.getName());
+		assertEquals(document.getLink(), unmarshalledDocument.getLink());
+		
+	}
+	
+	
+	private Document getDocument() {
```
jomarko: I think this could be parameterized by "docName" and then reused in `getDocuments()`

jbpm-document/src/main/java/org/jbpm/document/marshalling/DocumentsMarshallingStrategy.java
```diff
@@ -0,0 +1,103 @@
...
+		
+		return buff.toByteArray();
+	}
+
+	public Object unmarshal(Context ctx, ObjectInputStream objectInputStream, byte[] object, ClassLoader classLoader)
+			throws IOException, ClassNotFoundException {
+		
+		DroolsObjectInputStream is = new DroolsObjectInputStream(new ByteArrayInputStream(object), classLoader);
+		// first we read out the object class and size of the list we stored.
+		int size = is.readInt();
```
mswiderski: do we really read object type here? refers to comment...

jbpm-document/src/main/java/org/jbpm/document/marshalling/DocumentsMarshallingStrategy.java
```diff
@@ -0,0 +1,103 @@
...
+		ByteArrayOutputStream buff = new ByteArrayOutputStream();
+		ObjectOutputStream oos = new ObjectOutputStream(buff);
+		//Write the number of documents in the list.
+		oos.writeInt(documents.getDocuments().size());
+		for (Document nextDocument: documents.getDocuments()) {
+			//Use the DocumentMarshallingStrategy to marshall individual documents.
+			byte[] nextMarshalledDocument = docMarshallingStrategy.marshal(ctx, objectOutputStream, nextDocument);
+			oos.writeInt(nextMarshalledDocument.length);
+			oos.write(nextMarshalledDocument);
+			//Need to call reset on the stream in order for the Document bytes tom be written correctly.
```
mswiderski: tiny typo "tom"

jbpm-document/src/main/java/org/jbpm/document/Documents.java
```diff
@@ -0,0 +1,55 @@
...
+import javax.xml.bind.annotation.XmlAccessType;
+import javax.xml.bind.annotation.XmlAccessorType;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * A collection of {@link Document Documents}.
+ */
+@XmlAccessorType(XmlAccessType.FIELD)
+@XmlRootElement(name = "documents-object")
+public class Documents implements Serializable {
```
mswiderski: so in the end you didn't go for list of documents? Was it because of the type erasure? Or you don't see it as valuable option? Thinking on how this would work with forms? Maybe this approach will be actually better...

DuncanDoyle: Yes, type erasure in Java does not allow to properly determine the generic type of a List (or Collection in general). So, we would need to check at runtime the class type of elements of the collection (and even that does not give you a guarantee). Second, with an empty list you would probably need to dive into reflection.

This wrapper seems to be a cleaner solution.

PR #1036, opened 2017-10-19
JBPM-6506 - Update workitems archetype to also generate the workitem …

tsurdilo:
…index.html


tsurdilo:
Generates index.html file from workitem handler @Wid annotations which can be used when uploaded to workitem repository. index file looks like this:

screen shot 2017-10-19 at 1 51 58 pm

PR #1038, opened 2017-10-20
Stabilize jbpm-container tests

MarianMacik:
Same as #1037, but for master.

### DIFF
--- a/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/pom.xml
+++ b/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/pom.xml
@@ -146,6 +146,10 @@
       <artifactId>drools-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency> <!-- To make use of TransactionManager in CountDownProcessEventListener -->
+      <groupId>org.drools</groupId>
+      <artifactId>drools-persistence-api</artifactId>
+    </dependency>
     <dependency>
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-human-task-core</artifactId>
--- a/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/main/java/org/jbpm/test/container/listeners/CountDownProcessEventListener.java
+++ b/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/main/java/org/jbpm/test/container/listeners/CountDownProcessEventListener.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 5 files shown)


tsurdilo:
@MarianMacik my only worry here is that jbpm already includes 8 CountDownProcessEventListener classes and don't see the need to add yet another one? WDYT?

MarianMacik:
Yes, I am aware of that. The one thing is - this CountDown listener is more configurable (you just override what you want) and the other thing - if I used the one provided in jbpm-test-coverage then I would have to include jbpm-test-coverage module in my classpath + in the generated shrinkwrap archive so I can use it in Arquillian. I agree that we can unify this in the future.

MarianMacik:
@tsurdilo One thing that has crossed my mind - I can make my implementation called DefaultCountDownProcessEventListener which will be inherited by CountDownProcessEventListener - I can call it NodeLeftCountDownProcessEventListener since it reacts on before/afterNodeLeft events. This way it can be used in jbpm-test-coverage (NodeLeftCountDownProcessEventListener) and jbpm-container (DefaultCountDownProcessEventListener) as well. WDYT? However, better thing would be to move it to test-util or similar module - and do it once and for whole jbpm repo - unfortunately, I don't have time to do this refactoring now :(

mswiderski:
@MarianMacik what about moving one of them into jbpm-test-util that all modules depend on anyway...

MarianMacik:
Well, the issues is that one overrides before/afterNodeLeft methods, the other one overrides beforeNodeLeft and afterNodeTriggered methods and so on. So it can be done in the way that there will my Default one which will not override any methods - so it can by customized e.g. as an anonymous class and then there will be other e.g. NodeLeftCountDownProcessEventListener for nodeLeft events (configurable with property if it should check before or after event) and a similar one with nodeTriggered events.

The customizable one will be for use cases which I needed with jbpm-container tests - I needed to combine AfterProcessCompleted with AfterNodeTriggered events. I could do this with two listeners too - one with ProcessCompleted events and the other one would be the NodeTriggered listener which I have already mentioned, but one combined customized listener is enough I think for this use case.

There is also an option with only a customizable listener but there are tests which need the same listener in every test, but with different node name for example - therefore using a constructor is a better solution I guess in this situations. I am for the first option with some already predefined listeners + customizable listener which does not override anything by default. WDYT?

mswiderski:
I would say the one in jbpm-test-util should have the most common one eith implemented methods. Other modules when needed should extend or override

tsurdilo:
Yes you can define depends to org.jbpm:jbpm-test-util in your pom

MarianMacik:
@mswiderski Most of those listeners use nodeLeft and nodeTriggered event methods so I think we should add those as well to the jbpm-test-util since otherwise if I leave just the most common one with no-op methods from DefaultProcessEventListener in jbpm-test-util there would still be a lot of duplication - e.g. 4 listener classes with before and after node left event methods - because 4 modules will override them in exactly the same way.

mswiderski:
@MarianMacik +1 that's exactly what I meant - to add the most commonly used methods from one of the CountDownProcessEventListeners. Do you mind taking care of it or would you prefer to have it in another PR?

MarianMacik:
Since this stabilization and listeners refactoring are two different things, I would prefer to have it separately. So I think you can merge this and I will prepare another PR in which I will continue with my proof of concept - it is nearly done since now I only have to change references in particular modules to those new listeners - mostly it will be NodeLeft/NodeTriggered CountDownProcessEventListener and I think only in jbpm-container I will use the default one with my own overrides of methods. Should be pretty quick.

mswiderski:
@MarianMacik makes sense, thanks

PR #1039, opened 2017-10-23
JBPM-6510 - Refactor tests to use only one CountDownProcessEventListener

MarianMacik:
Hi, @mswiderski,

here is the PR for the change we discussed in #1038.

Now there are 4 ProcessEventListeners:

- DefaultCountDownProcessEventListener - completely customizable
- NodeCountDownProcessEventListener - parent of other two which are for node events
- NodeLeftCountDownProcessEventListener - for NodeLeft events, configurable via reactOnBeforeNodeLeft property, the default is after event
- NodeTriggeredCountDownProcessEventListener - for NodeTriggered events, configurable via reactOnBeforeNodeTriggered, the default is after event


mswiderski:
thanks a lot @MarianMacik

PR #1041, opened 2017-10-25
JBPM-6509 - Support for sub cases both from within case and process

mswiderski:


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
```diff
@@ -71,4 +75,14 @@ protected void triggerRules(KieSession ksession) {
         }
         ksession.fireAllRules();
     }
+    
+    protected CaseFileInstance getCaseFile(KieSession ksession, String caseId) {
+        Collection<? extends Object> caseFiles = ksession.getObjects(new ClassObjectFilter(CaseFileInstance.class));
+        if (caseFiles.size() == 0) {
+            return null;
+        }
+        CaseFileInstance caseFile = (CaseFileInstance) caseFiles.iterator().next();
```
rsynek: Can be more than a single case file returned by `ksession.getObjects`?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/NotifyParentCaseEventListener.java
```diff
@@ -0,0 +1,123 @@
...
+            logger.debug("Parent instance id {}, work item id {}, has been successfully notified about case {} completion", caseFileInstance.getParentInstanceId(), caseFileInstance.getParentWorkItemId(), event.getCaseId());
+        }
+    }
+    
+    protected CaseFileInstance getCaseFile(KieSession ksession) {
+        Collection<? extends Object> caseFiles = ksession.getObjects(new ClassObjectFilter(CaseFileInstance.class));
+        if (caseFiles.size() == 0) {
+            return null;
+        }
+        CaseFileInstance caseFile = (CaseFileInstance) caseFiles.iterator().next();
```
rsynek: Again, if the aim is to get a single item from a collection, because the collection should always have at most 1 item, I would rather throw an exception in case there are more items.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/StartCaseWorkItemHandler.java
```diff
@@ -0,0 +1,252 @@
...
+        if (caseFiles.size() == 0) {
+            logger.debug("No case file available in working memory returning null as case file");
+            return null;
+        }
+        CaseFileInstance caseFile = (CaseFileInstance) caseFiles.iterator().next(); 
+        
+        return caseFile;
+    }
+    
+    protected void parseParameters(WorkItem workItem, Map<String, Object> caseFileData, Map<String, OrganizationalEntity> roleAssignments, Map<String, List<String>> accessRestrictions) {
```
rsynek: Too complex method; every if branch can be delegated into a private method.

tsurdilo:
@mswiderski very nice! thanks

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -396,4 +400,30 @@ protected boolean checkProcessInstance(Activation activation) {
         }
         return true;
     }
+    
+    protected void mapDynamicOutputData(Map<String, Object> results) {
+        if (results != null && !results.isEmpty()) {
+            VariableScope variableScope = (VariableScope) ((ContextContainer) getProcessInstance().getProcess()).getDefaultContext( VariableScope.VARIABLE_SCOPE );
+            VariableScopeInstance variableScopeInstance = (VariableScopeInstance)(VariableScopeInstance)getProcessInstance().getContextInstance(VariableScope.VARIABLE_SCOPE);;
```
rsynek: `VariableScopeInstance` casting is done twice. There are also 2 semicolons in the end of the line.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
mswiderski: yes, when wrong strategy is used

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/StartCaseWorkItemHandler.java
mswiderski: that would mean iterating over parameters many times so I believe it's better to have more complex method instead of interacting over map 4 times

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
mswiderski: will remove duplicates

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
rsynek: In that case - just as a defensive approach, I would rather throw appropriate exception for that.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/StartCaseWorkItemHandler.java
rsynek: I don't think repeated iterating over map is necessary, I was thinking basically about something in line with:
https://gist.github.com/rsynek/ef94e33f63927854592f99a8b4721a9b

mswiderski: Yet another method makes more difficult when debugging but if you strongly believe it will make it better I’ll do it

rsynek: @mswiderski these are suggestions, if you don't believe it will make the code better, no point in changing it. 

My reasoning is that the method combines two level of abstraction - deciding what actions to take based on parameter prefixes and details of these actions. Following principles described in Clean Code (Robert C. Martin), method should deal with only 1 level of abstraction. I think quite the opposite is true about debugging such a method - you add one level to stack trace, truth; on the other hand, the name of the method (if chosen well) serves as a hint of what the method does - making the caller method more obvious and readable.

Whether you change it or not, your call.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
mswiderski: fixed

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/StartCaseWorkItemHandler.java
mswiderski: I choose not to split it into separate methods

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/CaseCommand.java
mswiderski: we don't do defensive programming ... plus to do this it would have to have yet another if that has very limited value in most of the cases

PR #1043, opened 2017-10-25
Jbpm tests stabilisation

dcihak:
AsyncAuditLogProducerTest test fixed: nodes and process instances logging

### DIFF
--- a/jbpm-audit/src/test/java/org/jbpm/process/audit/jms/AsyncAuditLogProducerTest.java
+++ b/jbpm-audit/src/test/java/org/jbpm/process/audit/jms/AsyncAuditLogProducerTest.java
@@ -76,7 +76,7 @@ public class AsyncAuditLogProducerTest extends AbstractBaseTest {
     private HashMap<String, Object> context;
     private ConnectionFactory factory;
     private Queue queue;
-    
+
     private EmbeddedJMS jmsServer;    
     
     @Before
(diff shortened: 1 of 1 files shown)


dcihak:
READY TO BE MERGED

tsurdilo:
My only comment here is that I'm not big fan of mixing assertj and junit in a test. Either do one or the other all the way. Maybe we could remove junit and replace its assertions with assertj.

rsynek:
@tsurdilo @dcihak +1, let's move to assertj whenever we touch some test (doing a global big bang is probably too much).

jbpm-audit/src/test/java/org/jbpm/process/audit/jms/AsyncAuditLogProducerTest.java
```diff
@@ -484,7 +482,7 @@ public void onMessage(Message message) {
             };
             consumer.setMessageListener(rec);
             // since we use message listener allow it to complete the async processing
-            Thread.sleep(2000);
+            Thread.sleep(waitTime);
```
rsynek: Although I like parameterization of Thread.sleep, I can see the waiting raised from 2 secs to 5. Is this the only way to make the tests stable? It prolongs their execution.

MarianMacik:
@mswiderski Is there any chance to check if there are still any JMS messages left to be consumed? So we can e.g. constantly check the number and don't have to use Thread.sleep(). If there is no good and quick solution for it (HornetQMessageConsumer does not have it), we can leave it as is.

By looking at the JMS documentation, the only provider-independent way to check a number of messages in a Queue is to use QueuBrowser and simply every time count number of messages in it, which is not convenient.

dcihak:
Using of junit methods replaced by assertj

dcihak:
Thread.sleep() was replaced by latch countdown to wait until all the messages in the queue are processed.

PR #1058, opened 2017-11-02
Fix case comment event emitting

rsynek:
Fixes wrong case comment events emitting and adds a reproducer.


rsynek:
Failures do not seem to be connected with this change.

Jenkins retest this.

PR #1061, opened 2017-11-06
adding debug log which is useful for trouble-shooting

tkobayas:
This is a small good-to-have debug log for timer related trouble-shooting. Please merge, thanks!


mswiderski:
Jenkins retest this

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
```diff
@@ -169,7 +169,8 @@ public void internalSchedule(TimerJobInstance timerJobInstance) {
 
         // Define a Trigger that will fire "now"
         org.quartz.Trigger triggerq = new SimpleTrigger(quartzJobHandle.getJobName()+"_trigger", quartzJobHandle.getJobGroup(), nextFireTime);
-            
+        logger.debug("triggerq.name = " + triggerq.getName() + ",  triggerq.startTime = " + triggerq.getStartTime()); // nextFireTime is mapped to startTime
```
mswiderski: @tkobayas could you please use "triggerq.name = {}" syntax instead of string concat?

tkobayas:
@mswiderski thanks! I have updated.

PR #1062, opened 2017-11-06
JBPM-6590 - Isolation of Quartz scheduler / triggers per kjar

mswiderski:
this PR provides extension to default SQL delegates that take into considerations deployments that are present in the system. By that queries done by quartz will filter triggers by the deployments that are actually able to execute the job.

To make this possible, trigger (and job) group name is now set to deployment id (if available).

### DIFF
--- a/jbpm-test-coverage/src/test/resources/quartz-db-short-misfire.properties
+++ b/jbpm-test-coverage/src/test/resources/quartz-db-short-misfire.properties
@@ -25,7 +25,7 @@ org.quartz.threadPool.threadPriority = 5
 org.quartz.jobStore.misfireThreshold = 3000
 
 org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreCMT
-org.quartz.jobStore.driverDelegateClass=org.quartz.impl.jdbcjobstore.StdJDBCDelegate
+org.quartz.jobStore.driverDelegateClass=org.jbpm.process.core.timer.impl.quartz.DeploymentsAwareStdJDBCDelegate
 org.quartz.jobStore.useProperties=false
 org.quartz.jobStore.dataSource=myDS
 org.quartz.jobStore.nonManagedTXDataSource=notManagedDS
--- a/jbpm-test-coverage/src/test/resources/quartz-db.properties
+++ b/jbpm-test-coverage/src/test/resources/quartz-db.properties
@@ -25,7 +25,7 @@ org.quartz.threadPool.threadPriority = 5
 org.quartz.jobStore.misfireThreshold = 60000
 
 org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreCMT
-org.quartz.jobStore.driverDelegateClass=org.quartz.impl.jdbcjobstore.StdJDBCDelegate
+org.quartz.jobStore.driverDelegateClass=org.jbpm.process.core.timer.impl.quartz.DeploymentsAwareStdJDBCDelegate
 org.quartz.jobStore.useProperties=false
 org.quartz.jobStore.dataSource=myDS
 org.quartz.jobStore.nonManagedTXDataSource=notManagedDS
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/timer/NamedJobContext.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/timer/NamedJobContext.java
@@ -22,4 +22,8 @@ public interface NamedJobContext extends JobContext {
     String getJobName();
     
     Long getProcessInstanceId();
+    
+    default String getDeploymentId() {
+        return null;
+    }
 }
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzSchedulerService.java
@@ -15,6 +15,14 @@
  */
 package org.jbpm.process.core.timer.impl;
 
+import java.io.NotSerializableException;
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Date;
+import java.util.concurrent.Callable;
+import java.util.concurrent.atomic.AtomicInteger;
+import java.util.concurrent.atomic.AtomicLong;
+
 import org.drools.core.time.InternalSchedulerService;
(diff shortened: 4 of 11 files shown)


mswiderski:
Jenkins retest this

jbpm-test-coverage/src/test/resources/quartz-db-short-misfire.properties
```diff
@@ -25,7 +25,7 @@ org.quartz.threadPool.threadPriority = 5
 org.quartz.jobStore.misfireThreshold = 3000
 
 org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreCMT
-org.quartz.jobStore.driverDelegateClass=org.quartz.impl.jdbcjobstore.StdJDBCDelegate
+org.quartz.jobStore.driverDelegateClass=org.jbpm.process.core.timer.impl.quartz.DeploymnetsAwareStdJDBCDelegate
```
MarianMacik: I think tests will not work since there is a typo in Deploym**nets**AwareStdJDBCDelegate.

jbpm-test-coverage/src/test/resources/quartz-db.properties
```diff
@@ -25,7 +25,7 @@ org.quartz.threadPool.threadPriority = 5
 org.quartz.jobStore.misfireThreshold = 60000
 
 org.quartz.jobStore.class=org.quartz.impl.jdbcjobstore.JobStoreCMT
-org.quartz.jobStore.driverDelegateClass=org.quartz.impl.jdbcjobstore.StdJDBCDelegate
+org.quartz.jobStore.driverDelegateClass=org.jbpm.process.core.timer.impl.quartz.DeploymnetsAwareStdJDBCDelegate
```
MarianMacik: Same typo.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/quartz/DeploymnetsAwareStdJDBCDelegate.java
```diff
@@ -0,0 +1,148 @@
...
+            return nextTriggers;
+        } finally {
+            closeResultSet(rs);
+            closeStatement(ps);
+        }      
+    }
+    
+    @Override
+    public int countMisfiredTriggersInStates(
+                                             Connection conn, String state1, String state2, long ts) throws SQLException {
```
MarianMacik: This is some weird formatting style :) Was it done automatically by KIE code style? If so, then we have a bug in it.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/quartz/QuertzUtils.java
```diff
@@ -0,0 +1,92 @@
...
+
+package org.jbpm.process.core.timer.impl.quartz;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.kie.internal.runtime.manager.RuntimeManagerRegistry;
+import org.quartz.impl.jdbcjobstore.StdJDBCConstants;
+
+public class QuertzUtils implements StdJDBCConstants {
```
MarianMacik: Just 2 things - I was searching for QuartzUtils but then I realized there is a typo in Qu**e**rtzUtils. The second thing is that since this is a util class, can we make methods and constants static? Then we don't have to initialize it inside Quartz delegates.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/quartz/QuertzUtils.java
```diff
@@ -0,0 +1,92 @@
...
+            + TABLE_PREFIX_SUBST + TABLE_TRIGGERS + " WHERE "
+            + COL_NEXT_FIRE_TIME + " < ? " 
+            + "AND ((" + COL_TRIGGER_STATE + " = ?) OR (" + COL_TRIGGER_STATE + " = ?)) ";
+            
+    String MISFIRED_ORDER_BY = "ORDER BY " + COL_NEXT_FIRE_TIME + " ASC";
+    
+    public List<String> getDeployments() {
+        List<String> deploymentIds = new ArrayList<>(RuntimeManagerRegistry.get().getRegisteredIdentifiers());
+        // add jbpm as trigger group for backward compatibility
+        deploymentIds.add("jbpm");
```
MarianMacik: Just to have this clarified - is this for situations when we have old quartz jobs in DB which were created with "jbpm" group and we have to execute them? And all newly scheduled jobs will have group which equals their deployment ID?

jbpm-test-coverage/src/test/resources/quartz-db-short-misfire.properties
MarianMacik: It seems the tests has passed. Do you have any idea why?

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/quartz/DeploymnetsAwareStdJDBCDelegate.java
mswiderski: will double check

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/quartz/QuertzUtils.java
mswiderski: the point here is that it implements the StdJDBCConstants as the extensions refer to other fields for the interface. I made it a class to keep the door open as it might need more work down the road.
Will fix the type on class name

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/quartz/QuertzUtils.java
mswiderski: correct

mswiderski:
tests have passed as types are copied everywhere - name of the Delegate classes have typo as well. Will fix them

Thanks @MarianMacik

mswiderski:
fixes applied

mswiderski:
Jenkins retest this

psiroky:
Build finished. 3581 tests run, 132 skipped, 1 failed.

mswiderski:
Jenkins retest this

PR #1063, opened 2017-11-07
JBPM-6597: Prevent NPE when document storage dir doesn't exist

jhrcek:
Test to reproduce the issue and proposed fix


jhrcek:
@mswiderski please check.

PR #1064, opened 2017-11-07
JBPM-6595 - Case Roles: User/group assignments fail when using Case File

mswiderski:


tsurdilo:
Jenkins retest please

PR #1069, opened 2017-11-14
JBPM-6618 - Enable rule execution limit in business rule task

mswiderski:
@etirelli could you please have a look?

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
@@ -81,9 +81,12 @@ import java.math.BigDecimal;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
 
 @RunWith(Parameterized.class)
 public class ActivityTest extends JbpmBpmn2TestCase {
@@ -1963,5 +1966,32 @@ public class ActivityTest extends JbpmBpmn2TestCase {
         log = logService.findProcessInstance(processInstance.getId());
         assertNotNull(log);
         assertEquals("Parent process should be completed and not aborted", ProcessInstance.STATE_COMPLETED, log.getStatus().intValue());
+    }
+    
+    @Test
+    public void testBusinessRuleTaskFireLimit() throws Exception {
+        KieBase kbase = createKnowledgeBaseWithoutDumper("BPMN2-BusinessRuleTaskLoop.bpmn2",
+                "BPMN2-BusinessRuleTaskInfiniteLoop.drl");
+        ksession = createKnowledgeSession(kbase);
+        ksession.insert(new Person());
+        assertThatExceptionOfType(RuntimeException.class).isThrownBy(() -> { 
+            ksession.startProcess("BPMN2-BusinessRuleTask");
+        })
+        .withMessageContaining("Fire rule limit reached 10000");        
+    }
+    
+    @Test
+    public void testBusinessRuleTaskFireLimitAsParameter() throws Exception {
+        KieBase kbase = createKnowledgeBaseWithoutDumper("BPMN2-BusinessRuleTaskWithDataInputLoop.bpmn2",
+                "BPMN2-BusinessRuleTaskInfiniteLoop.drl");
+        ksession = createKnowledgeSession(kbase);
+        ksession.insert(new Person());
+        
+        Map<String, Object> parameters = Collections.singletonMap("limit", 5);
+        
+        assertThatExceptionOfType(RuntimeException.class).isThrownBy(() -> { 
+            ksession.startProcess("BPMN2-BusinessRuleTask", parameters);
+        })
+        .withMessageContaining("Fire rule limit reached 5");        
     }
 }
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
@@ -75,6 +75,9 @@ public class RuleSetNodeInstance extends StateBasedNodeInstance implements Event
     private static final Logger logger = LoggerFactory.getLogger(RuleSetNodeInstance.class);
     
     private static final String ACT_AS_WAIT_STATE_PROPERTY = "org.jbpm.rule.task.waitstate";
+    private static final String FIRE_RULE_LIMIT_PROPERTY = "org.jbpm.rule.task.firelimit";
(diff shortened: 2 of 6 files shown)


jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -119,7 +121,7 @@ public void internalTrigger(final NodeInstance from, String type) {
                 }
 
                 if (decision != null && !decision.isEmpty()) {
-                    dmnResult = runtime.evaluateDecisionByName(dmnModel, decision, context);
+                    dmnResult = runtime.evaluateByName(dmnModel, context, decision);
```
mswiderski: @tarilabs is this right refactor to not use deprecated method?

tarilabs: indeed as per Javadocs. Thanks!

mswiderski: cool, thanks

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -75,6 +75,8 @@
     private static final Logger logger = LoggerFactory.getLogger(RuleSetNodeInstance.class);
     
     private static final String ACT_AS_WAIT_STATE_PROPERTY = "org.jbpm.rule.task.waitstate";
+    private static final String FIRE_RULE_LIMIT_PARAMETER = "FireRuleLimit";
+    private static final int DEFAULT_FIRE_RULE_LIMIT = Integer.parseInt(System.getProperty("org.jbpm.rule.task.firelimit", "100"));
```
etirelli: @mswiderski I think 100 is too low for a default limit. If the user does not define a limit, I think something like 10000 would be a better limit as it will not break existing cases, but still prevent infinite loops.

etirelli:
Everything else looks good.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
mswiderski: here you go sir, 10000 it is :)

MarianMacik:
jenkins retest this

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -75,6 +75,8 @@
     private static final Logger logger = LoggerFactory.getLogger(RuleSetNodeInstance.class);
     
     private static final String ACT_AS_WAIT_STATE_PROPERTY = "org.jbpm.rule.task.waitstate";
+    private static final String FIRE_RULE_LIMIT_PARAMETER = "FireRuleLimit";
+    private static final int DEFAULT_FIRE_RULE_LIMIT = Integer.parseInt(System.getProperty("org.jbpm.rule.task.firelimit", "10000"));
```
triceo: Are we sure that we want to enable this for everyone, always? My opinion is that this should default to disabled, for the sake of backwards compatibility - if it wasn't a problem for them until now, let's not have them think about this. Especially given the arbitrary nature of the threshold.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
```diff
@@ -151,10 +158,18 @@ public void internalTrigger(final NodeInstance from, String type) {
...
+                    if (inputs.containsKey(FIRE_RULE_LIMIT_PARAMETER)) {
+                        fireLimit = Integer.parseInt(inputs.get(FIRE_RULE_LIMIT_PARAMETER).toString());
+                    }
                     ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
                             .activateRuleFlowGroup(getRuleFlowGroup(), getProcessInstance().getId(), getUniqueId());
 
-                    ((KieSession) getProcessInstance().getKnowledgeRuntime()).fireAllRules();
+                    int fired = ((KieSession) getProcessInstance().getKnowledgeRuntime()).fireAllRules(fireLimit);
+                    if (fired == fireLimit) {
+                        throw new RuntimeException("Fire rule limit reached " + fireLimit);
```
triceo: Any chance we could include the property in the exception message, or some docs pointer? If I saw this exception, I would immediately start wondering what this limit is and where to set it.

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
```diff
@@ -1964,4 +1966,32 @@ public void testCallActivitySkipAbortParent() throws Exception {
...
+    
+    @Test
+    public void testBusinessRuleTaskFireLimitAsParameter() throws Exception {
+        KieBase kbase = createKnowledgeBaseWithoutDumper("BPMN2-BusinessRuleTaskWithDataInputLoop.bpmn2",
+                "BPMN2-BusinessRuleTaskInfiniteLoop.drl");
+        ksession = createKnowledgeSession(kbase);
+        ksession.insert(new Person());
+        
+        Map<String, Object> parameters = new HashMap<>();
+        parameters.put("limit", 5);
```
triceo: You might consider using `Collections.singletonMap()` for slightly nicer code.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
mswiderski: I actually think it is better to enable it by default mainly due to number of people ran into looping rules and making their servers unresponsive (and thus had to kill them) just ti figure out they made mistake in the rule that simply gets into infinitive loop. SO better would be to get few failures due to limit instead of having issues with high cpu utilisation and trying to figure what's wrong with my server.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
mswiderski: +1 here to provide more meaningful message

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
mswiderski: matter of preference :)

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/RuleSetNodeInstance.java
triceo: Yeah, you're right. I forgot that this does actually kill the server if it happens.

mswiderski:
@triceo changes applied, thanks

PR #1072, opened 2017-11-15
fixed jbpm distribution after removing/moving work items

mswiderski:


mswiderski:
@tsurdilo could you please have a look?

PR #1085, opened 2017-11-28
JBPM-6637 - JDK9: fix failing jbpm tests

mswiderski:
depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/625


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/SubprocessesTest.java
```diff
@@ -59,7 +59,7 @@
     private static final String HELLO_WORLD_PROCESS = "org/jbpm/test/functional/common/HelloWorldProcess1.bpmn";
     private static final String HELLO_WORLD_PROCESS_ID = "org.jbpm.test.functional.common.HelloWorldProcess1";
 
-    private static final boolean JAVA8 = System.getProperty("java.version").contains("1.8");
+    private static final boolean JAVA8 = System.getProperty("java.version").contains("1.8") || System.getProperty("java.version").startsWith("9.");
```
MarianMacik: I would rename it to JAVA8_OR_9 then.

jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/settings/JpaSettingsTest.java
```diff
@@ -25,6 +26,7 @@
 import static org.powermock.api.mockito.PowerMockito.*;
 import static org.junit.Assert.*;
 
+@Ignore
```
MarianMacik: Does it mean that these tests will be skipped permanently?

mswiderski: I would say so, not much value in this test to be honest

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/SubprocessesTest.java
mswiderski: it's initial version to see if changes done for jdk9 also work on jdk 8 :) once they do we can do additional cleanup etc

mswiderski:
@psiroky can we test this branch somehow on JDK9?

mswiderski:
for information - currently I disabled remote ejb tests (container test) as jboss marshalling does not seem to work with JDK9 due to access to sun.reflect.* classes and I was not able to figure out how to make it work.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/SubprocessesTest.java
mswiderski: renamed

psiroky:
@mswiderski yes. I think we can do the same as with the narayna job while ago.

MarianMacik:
as we noticed with @psiroky, jboss-marshalling is quite old because of jboss-remote-naming which is of version 2.0.5. So I tried to replace it with widlfly-naming and overrode <version.org.wildfly> to 11.0.0.Final but then I am getting
```
javax.naming.NoInitialContextException: Need to specify class name in environment or system property, or as an applet parameter, or in an application resource file:  java.naming.factory.initial
```
in JBossEJBClientServiceFactory so not sure if it is prepared for the newest WildFly.

mswiderski:
@psiroky can we use the job that was initially used to produce the report about jdk9 for jBPM? Once we know it builds fine we can directly merge it as it seems it does not break java 8 builds

jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/settings/JpaSettingsTest.java
psiroky: IMO if the plan is to never enable the tests again, they should be simply deleted.

psiroky:
I would prefer not to mess with the original job. I've cloned it into https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jbpm-jdk9-mswiderski/ and tweaked to use the `msiderwski:jdk9` branch. The build is currently failing due to the recent changes on master, please rebase.

mswiderski:
@psiroky thanks, rebased, both this one and build bootstrap PR

psiroky:
@mswiderski hmm, the job actually does not take the bootstrap PR into account :( I guess the only way would be merge the bootstrap PR before this one.

mswiderski:
@psiroky I guess it's safe to merge it, wdyt?

psiroky:
@mswiderski I am not 100% sure. I've executed full downstream build, so let's see.

mswiderski:
@psiroky looks like we have green build on JDK 9 according to this one:
https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jbpm-jdk9-mswiderski/2/testReport/

I just did the clean up - removed tests that are of no value and one that is flaky and causes random failures (was suppose to remove it long time ago)

once it finishes I'll run on more build of that jdk9 job on this branch.

@psiroky @MarianMacik any other comments or shall we merge it?

PR #1086, opened 2017-11-30
JBPM-6487 JBPM tests stabilisation

dcihak:
* tasks deadline dates are set just before the task start to avoid premature notification sending
* junit methods replaced by assertj


dcihak:
[READY TO BE MERGED]

PR #1087, opened 2017-12-01
add json generation to wid processor and update html template look/feel

tsurdilo:


tsurdilo:
@mswiderski this adds json template which will be generated alongside the .wid and the .html in the wid processor. Also updated the html look/feel (before when there were multiple workitem handlers in a single workitem module it was really hard to read the page)

PR #1089, opened 2017-12-01
JBPM-6665 - Add 'before' TaskInputVariableChanged and TaskOutputVaria…

mswiderski:
…bleChanged events to TaskLifeCycleEventListener

depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/281

@krisv @mcivantos-tribalyte  here is a PR that introduces task event to be updated when data change, before and after data change and user id to always be available, either from the task command or from security context if none was found on command. That way task events have always some user id attached.

PR #1091, opened 2017-12-05
JBPM - 6641 - Return the input&output data of tasks when using Advanced Queries

mcivantos-tribalyte:
Added updates in data/metadata of task in the TASKEVENT table.

### DIFF
--- a/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
+++ b/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
@@ -763,7 +763,7 @@ public class JPATaskLifeCycleEventListener extends PersistableEventListener impl
             return;
         }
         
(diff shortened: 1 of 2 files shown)


tsurdilo:
ok to test

mswiderski:
@mcivantos-tribalyte this is already covered in this PR https://github.com/kiegroup/jbpm/pull/1089/files#diff-96f9a741dfc90d35e4aa0f49da0af8a8R712

moreover we should not add updated event on task inputs as they are usually added when task is created so you would end up in two events for single operation which in my opinion is wrong.

So overall, this PR duplicates #1089

mcivantos-tribalyte:
@mswiderski You are right. I would only suggest adding on your PR the update of the lastModificationDate in AuditTaskImpl after this https://github.com/kiegroup/jbpm/pull/1089/files#diff-96f9a741dfc90d35e4aa0f49da0af8a8R712
Or I could change this PR, when your PR is merged in order to updated the Audit task info.

mswiderski:
@mcivantos-tribalyte good point, added that to that PR

mswiderski:
closing as it is covered by #1089

PR #1099, opened 2017-12-15
[6.5.x] RHBPMS-5085 - ConcurrentModificationException in hibernate with SINGL…

mswiderski:
…ETON deployments under load (#1093)


mswiderski:
Jenkins retest this

PR #1104, opened 2017-12-22
JBPM-6699 - fix for wid path and file

tsurdilo:


tsurdilo:
@mswiderski my previous commit broke some tests in jbpm-designer which relied on different path settings. this fixes the issue.

PR #1106, opened 2017-12-27
Add missing .gitignore files in several modules

dmvolod:
These are missing .gitignore files in several modules inside jbpm project found during import whole project inside Eclipse. Please add them. Thank you.


tsurdilo:
ok to test

PR #1107, opened 2018-01-02
JBPM-6720 Add timeout to test and replace wait notify with a latch

baldimir:
@mswiderski or anyone else, could you please review? 

The test still fails randomly. I am not sure if there is a bug in JBPM or not. The process listener is invoked 4 times occasionally (looks like just 4 out of 5 processes finish sometimes).

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/async/AsyncIntermediateCatchSignalTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/async/AsyncIntermediateCatchSignalTest.java
@@ -25,6 +25,7 @@ import org.jbpm.test.JbpmTestCase;
 import org.jbpm.test.wih.FirstErrorWorkItemHandler;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.kie.api.event.process.DefaultProcessEventListener;
 import org.kie.api.event.process.ProcessCompletedEvent;
@@ -91,6 +92,7 @@ public class AsyncIntermediateCatchSignalTest extends JbpmTestCase {
         latch.await();
     }
 
+    @Ignore("JBPM-6720 Possible jBPM bug, test fails randomly. Ignored till resolved.")
     @Test(timeout = 20000)
     public void testCorrectProcessStateAfterExceptionSignalCommandMulti() throws InterruptedException {
         latch = new CountDownLatch(5);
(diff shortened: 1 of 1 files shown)


tsurdilo:
@baldimir i am still getting " test timed out after 20000 milliseconds" when running this test (with your changes) locally.

tsurdilo:
Jenkins retest please

baldimir:
@tsurdilo yes it still fails randomly, please see PR description and the JIRA. It might be a jBPM bug. However if there is no quick solution we should consider merging this or ignoring the test, because without these changes it just hangs indefinitely sometimes.

tsurdilo:
@baldimir I will look into the cause of the issue, but for now here is small change to your pr here that will allow this test to not time out and run ok all the time  (without ignoring the test) - https://gist.github.com/tsurdilo/ac61831fb6511d4c1a3e94173d722e1b

tsurdilo:
@baldimir if you still want to use afterProcessCompleted event for countdown you can use this (just dont use the FirstErrorWorkItemHandler which I think in this test should not be used honestly anyways) and define your own passthrough one:

https://gist.github.com/tsurdilo/6ee51515fcee1538276517185a0a85fc

baldimir:
@tsurdilo thanks, will take a look at your suggestions.

baldimir:
@tsurdilo at the end I just @Ignored the test till resolved. I think this is the best for now. Is it ok? Will take a look meanwhile at the problem.

jhrcek:
@tsurdilo If ok like this, please merge this soon. Full downstr.build is freezing very ofthen because of this test, blocking merging important PRs.

baldimir:
Found the commit that causes this test to fail: https://github.com/baldimir/jbpm/commit/d25305ff856b722243c7b474433d549789cd8073

MarianMacik:
We have a nice suite of listeners for that: https://github.com/kiegroup/jbpm/tree/master/jbpm-test-util/src/main/java/org/jbpm/test/listener

Just extend the Default one. The issue with a plain afterProcessCompleted listener is that it is invoked before data is committed to DB, so it might happen that a test is faster than DB and it continues even though data is not in DB yet. To do it properly you have to do it via tx sync in afterCompletion method: https://github.com/kiegroup/jbpm/blob/master/jbpm-test-util/src/main/java/org/jbpm/test/listener/DefaultCountDownProcessEventListener.java#L66

So I think this is the way to go. Just override the afterProcessCompleted method and do countdown() there. Then register it on ksession.

Hope this helps.

PR #1108, opened 2018-01-09
JBPM-6720 - Test AsyncIntermediateCatchSignalTest.testCorrectProcessS…

mswiderski:
…tateAfterExceptionSignalCommandMulti fails randomly

@baldimir do you mind to take a look?

In general the problem was with work item handler that assumed that calls to the executeWorkItem method will go in sync for given process instance while it is not really the case as executor can pick any job and thus it led to execution of same process instance twice in a roll and thus not completing the other instance that was tried next as it set the first back to true.

### DIFF
--- a/jbpm-test-coverage/src/main/java/org/jbpm/test/wih/FirstErrorWorkItemHandler.java
+++ b/jbpm-test-coverage/src/main/java/org/jbpm/test/wih/FirstErrorWorkItemHandler.java
@@ -16,7 +16,9 @@
 
 package org.jbpm.test.wih;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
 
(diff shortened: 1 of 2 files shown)


baldimir:
@mswiderski if green +1, looks good. Thanks for explanation and fix.

mswiderski:
@baldimir you mean green for 10 times in a roll (at least in regards to this test class) ;)

mswiderski:
first green, 9 to go...

Jenkins retest this

baldimir:
@mswiderski I think it is enough if it passes here next time and 10 times locally on your machine :)

mswiderski:
let's try it one more time ... 

jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1109, opened 2018-01-09
JBPM-6728 - Missing process definitions when creating duplicated cont…

mswiderski:
…ainer with same processes

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/ProcessDescriptor.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/ProcessDescriptor.java
@@ -195,5 +195,30 @@ public class ProcessDescriptor implements Serializable {
         referencedClasses.clear();
         referencedRules.clear();
     }
+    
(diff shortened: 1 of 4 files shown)


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/bpmn2/ProcessDescriptor.java
```diff
@@ -108,7 +108,7 @@ public void resolveReusableSubProcessNames( Collection<Process> deploymentProces
 
 
     public ProcessAssetDesc getProcess() {
-        return process;
+        return process.copy();
```
tsurdilo: Is copy always required? Maybe make it default to copy and configurable to not if its not needed?

mswiderski: yes, it must be copied always to avoid this kind of issues as we can't rely on checks if something is already using it or not.

mswiderski:

mswiderski:

PR #1115, opened 2018-01-16
JBPM-6713 - Fluent Rule Flow API - signal example

tsurdilo:
Adds test to show correct usage for JBPM-6713 and can set dynamic process via RuleFlowProcessFactory now (plus formatting)


tsurdilo:
Jenkins retest please

PR #1118, opened 2018-01-18
JBPM-6732 - Return the input&output data of tasks when using Advanced

mcivantos-tribalyte:
Custom Queries


mswiderski:
ok to test

PR #1126, opened 2018-02-01
JBPM-6001 - Trigger AdHoc Node is not implemented for stage, JBPM-600…

mswiderski:
…8 - Triggering ad-hoc fragments in specific case stage, added tests to cover close case from rules, from within process execution and from async command

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
@@ -39,6 +39,7 @@ import org.jbpm.casemgmt.api.dynamic.TaskSpecification;
 import org.jbpm.casemgmt.api.generator.CaseIdGenerator;
 import org.jbpm.casemgmt.api.model.AdHocFragment;
 import org.jbpm.casemgmt.api.model.CaseDefinition;
+import org.jbpm.casemgmt.api.model.CaseStage;
 import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
 import org.jbpm.casemgmt.api.model.instance.CaseInstance;
 import org.jbpm.casemgmt.api.model.instance.CaseMilestoneInstance;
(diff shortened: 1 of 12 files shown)


mswiderski:
Jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/TriggerAdHocNodeInStageCommand.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public Void execute(Context context) {
+        KieSession ksession = ((RegistryContext) context).lookup( KieSession.class );
+        
+        ProcessInstance processInstance = ksession.getProcessInstance(processInstanceId);
+        
+        DynamicNodeInstance dynamicContext = (DynamicNodeInstance) ((WorkflowProcessInstanceImpl) processInstance).getNodeInstances(true).stream()
+                .filter(ni -> (ni instanceof DynamicNodeInstance) && stageId.equals(ni.getNode().getMetaData().get("UniqueId")))
+                .findFirst()
+                .orElse(null);
```
MarianMacik: Cannot this be changed to orElseThrow?

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/wih/AsyncWorkItemHandler.java
```diff
@@ -29,6 +29,7 @@
 import org.kie.api.runtime.process.WorkItemHandler;
 import org.kie.api.runtime.process.WorkItemManager;
 import org.kie.api.runtime.query.QueryContext;
+import org.kie.internal.runtime.Cacheable;
```
MarianMacik: Just curious, does it provide any benefits to have WIH cacheable? Maybe performance improvement?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseServiceImpl.java
```diff
@@ -390,6 +392,37 @@ public void triggerAdHocFragment(Long processInstanceId, String fragmentName, Ob
         ProcessInstanceDesc pi = runtimeDataService.getProcessInstanceById(processInstanceId);
         internalTriggerAdHocFragment(pi, fragmentName, data);
     }
+    
+    @Override
+    public void triggerAdHocFragment(String caseId, String stageId, String fragmentName, Object data) throws CaseNotFoundException {
+        authorizationManager.checkAuthorization(caseId);
```
MarianMacik: This call is missing in the second method.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/wih/AsyncWorkItemHandler.java
mswiderski: it means that the work item handler does not have to be recreated for each execution - especially important for process instance and request strategies.

mswiderski:
@MarianMacik all comments applied.

mswiderski:
Jenkins retest this

PR #1133, opened 2018-02-09
JBPM-6821: Information not available in NodeInstance event listener

mcivantos-tribalyte:
@mswiderski This PR stores the eventDate in NODEINSTANCELOG the same way you did in processes and tasks.


mswiderski:
ok to test

PR #1134, opened 2018-02-13
JBPM-6828 - REST endpoint containers/{containerId}/processes/instance…

tsurdilo:

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
@@ -17,6 +17,7 @@
 package org.jbpm.kie.services.impl;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
```diff
@@ -505,6 +511,32 @@ public Object execute(org.kie.api.runtime.Context context) {
...
+        for(NodeInstance activeNode : activeNodes) {
+            uniqueIdNodeInstanceMap.put((String) activeNode.getNode().getMetaData().get("UniqueId"), activeNode);
+        }
+
+
+        try {
+            for(Node processNode : processNodesList) {
+                if(processNode instanceof BoundaryEventNode) {
+                    activeBoundaryNodesSignals.add(VariableUtil.resolveVariable((String) processNode.getMetaData().get("SignalName"),
+                                                                                uniqueIdNodeInstanceMap.get(((BoundaryEventNode) processNode).getAttachedToNodeId())));
```
mswiderski: isn't that part uniqueIdNodeInstanceMap.get can return null and then resolveVariable will throw NPE? Maybe it would be better to avoid that... exceptions might be expensive ...

tsurdilo:
@mswiderski thanks - amended

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
```diff
@@ -505,6 +511,32 @@ public Object execute(org.kie.api.runtime.Context context) {
         	disposeRuntimeEngine(manager, engine);
         }
 	}
+
+	private List<String> getActiveBoundaryNodesSignals(ProcessInstance processInstance, Collection<NodeInstance> activeNodes) {
```
MarianMacik: BTW I noticed that this method is private and it is mixed among public methods. Not sure how is your attitude to that, just sharing my observation.

tsurdilo:
@MarianMacik good point. moved it down with other util methods and made it protected. thanks.

PR #1136, opened 2018-02-15
[7.5.x] JBPM-6828 - REST endpoint containers/{containerId}/processes/instance…

tsurdilo:

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
@@ -17,6 +17,7 @@
 package org.jbpm.kie.services.impl;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


tsurdilo:
related jiras RHBA-392, RHBPMS-5120

mswiderski:
7.5 branch is no longer updated so I don't think we need it here

tsurdilo:
@krisv can you please comment on this. you mentioned we might need a pr for 7.5.x for consistency but it seems we don't as @mswiderski said. Please let us know what you think. Thanks.

krisv:
@tsurdilo @mswiderski  7.5.x branch is still active, it's the RHDM 7.0.x product branch.  So yes, this should go in there.  It's 7.6.x that is no longer maintained.

mareknovotny:
@tsurdilo don't forget to push this on master too

mswiderski:
@krisv this does not have any value for RHDM as it is part of jBPM kie server extension which is disabled in RHDM.

@mareknovotny it's already on master

krisv:
@mswiderski agreed, and I discussed this with Edson, we believe it might be better for consistency to cherry-pick all fixes we do to 6.5.x to more recent product branches (7.5.x in this case) and master of course, so our supported streams don't diverge.

PR #1137, opened 2018-02-15
[6.5.x] JBPM-6828 - REST endpoint containers/{containerId}/processes/instance…

tsurdilo:


tsurdilo:
@mswiderski in 6.5.x we do not have resolution of JBPM-5869 and i cannot simply port that because it needs some code changes i believe happened before that as well (and they depends on java 1.8 too). So this is the port which has the same implementation just with knowing that for 6.5.x boundary events when signal is expression will not work (maybe should document?). wdyt?

tsurdilo:
related jiras RHBA-392, RHBPMS-5120

mareknovotny:
Jenkins please retest this

PR #1138, opened 2018-02-16
update to wid processor - externalize resource templates to repository

tsurdilo:


tsurdilo:
@mswiderski this removes the hard-coded templates for generation of .wid .json and index.html from the wid processor and moves it to the modules (archetype, repository) which use it. Allows those modules to control the templates (will allow users to "skin" the service repository completely which I think is very useful)

tsurdilo:
Jenkins retest please

tsurdilo:
@MarianMacik good point - i meant to replace this with a new test method but noticed did not do it. Now added (amended). Thanks!

PR #1122, opened 2018-01-22
[Do not MERGE yet] RHDM-251 Wildfly 11 upgrade

mareknovotny:

PR #1113, opened 2018-01-15
[do not merge] Installer - upgrade wf10 --> wf11

nmirasch:
installer configuration files to upgrade wf10 -> wf11

PR #1139, opened 2018-02-16
RHDM-251 - Wildfly 11 upgrade

mareknovotny:
this is another round of reverted https://github.com/kiegroup/jbpm/pull/1122 and https://github.com/kiegroup/jbpm/pull/1113 from @nmirasch


mareknovotny:
it seems that master has the same set of failed tests like this PR - https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/view/master/job/jbpm/552/

mareknovotny:
What is worse I can't compile it locally now
```

**Comment by mswiderski** [comment]
@mareknovotny rebase your branch of droolsjbpm-knowledge

**Comment by mareknovotny** [comment]
ok, thanks for that advice.

**Comment by mswiderski** [comment]
@mareknovotny looks like PR job is running already with wildfly 11

```
-e -nsu -fae -B -T1C -Pwildfly11 clean install
```

then this branch does not have provide named that way in here https://github.com/mareknovotny/jbpm/tree/9e771e308e39b69704f037a971b00451af655f77/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/resources/arquillian

I believe you should rename the wildfly10x to wildfly11x and then those EJB tests should work.

MarianMacik:
Maciej is right, each container has its own arquillian.xml file.

mswiderski:
and we are green here, thanks @mareknovotny

MarianMacik:
Looks good, though jBPM Executor redesign is already present on master, isn't it? Why does GitHub show it?

mareknovotny:
@MarianMacik that was my  pure rebase, so the order is a little confusing imho, but could be something different

mareknovotny:
@mswiderski @MarianMacik I redid the rebase and commits are now in correct order

PR #1141, opened 2018-02-21
RHBA-400 - Unable migrate process instance with active timer

mswiderski:


mswiderski:
Jenkins retest this

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/migration/TimerMigrationManagerTest.java
```diff
@@ -0,0 +1,412 @@
...
+        log = auditService.findProcessInstance(pi1.getId());
+        auditService.dispose();
+        assertNotNull(log);
+        assertEquals(BOUNDARY_TIMER_ID_V2, log.getProcessId());
+        assertEquals(DEPLOYMENT_ID_V2, log.getExternalId());
+        assertEquals(ProcessInstance.STATE_COMPLETED, log.getStatus().intValue());
+
+    }
+    
+    @Test(timeout=1000000)
```
MarianMacik: I think 10 000 should be enough.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
```diff
@@ -113,9 +131,15 @@ public MigrationReport migrate(Map<String, String> nodeMapping) {
         KieSession tobe = null;
 	    TransactionManager txm = null;
 	    boolean transactionOwner = false;
+	    InternalRuntimeManager currentManager = (InternalRuntimeManager) RuntimeManagerRegistry.get().getManager(migrationSpec.getDeploymentId());     
+        InternalRuntimeManager toBeManager = (InternalRuntimeManager) RuntimeManagerRegistry.get().getManager(migrationSpec.getToDeploymentId());
+        
+        Map<Long, List<TimerInstance>> timerMigrated = null;
```
MarianMacik: I would check the formatting in all files of this PR.

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/migration/TimerMigrationManagerTest.java
jhrcek: Also note that since java 7 it's possible to [use '_' in numeric literals](https://docs.oracle.com/javase/8/docs/technotes/guides/language/underscores-literals.html), which makes them easier to read, like `10_000`. I like that feature and use it for all literals with 5 digits or more :)

mswiderski:
@MarianMacik reformatted entire MigrationManager class and changed the value of timeout

mswiderski:
Jenkins retest this

PR #1145, opened 2018-02-23
JBPM-6771: Information not available in ProcessEventListener

mcivantos-tribalyte:


jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessInstance.java
```diff
@@ -64,5 +65,9 @@
 	String getDeploymentId();
 	
 	void setDeploymentId(String deploymentId);
+	
+	Date getStartDate();
+	
+	void setStartDate(Date startDate);
```
mswiderski: setter is certainly not needed as it should never be possible to set it other than process engine itself.

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -123,7 +123,8 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
                 ProcessInstanceInfo processInstanceInfo = (ProcessInstanceInfo) context.findProcessInstanceInfo( id );
                 if ( processInstanceInfo == null ) {
                     return null;
-                }                
+                }  
+                processInstance.setStartDate(processInstanceInfo.getStartDate());
```
mswiderski: I would rather set this start date in org.jbpm.persistence.processinstance.ProcessInstanceInfo.getProcessInstance(InternalKnowledgeRuntime, Environment, boolean) ... and I see you dod that in the next file so why this is needed?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -751,6 +753,14 @@ public String getCorrelationKey() {
     public void setCorrelationKey(String correlationKey) {
         this.correlationKey = correlationKey;
     }
+    
+    public Date getStartDate() {
+        return startDate;
+    }
+    
+    public void setStartDate(Date startDate) {
+        this.startDate = startDate;
```
mswiderski: here setter is fine though I would most likely rename it to internalSetStartDate so it's not directly seen as set* and in addition allow setting it only when it was not set yet.

mcivantos-tribalyte:
@mswiderski All your changes requested are already done.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -751,6 +753,14 @@ public String getCorrelationKey() {
...
         this.correlationKey = correlationKey;
     }
+    
+    public Date getStartDate() {
+        return startDate;
+    }
+    
+    public void internalSetStartDate(Date startDate) {
+        this.startDate = startDate;
+    }
```
MarianMacik: "in addition allow setting it only when it was not set yet." option is still missing.

mcivantos-tribalyte: Done

mswiderski:
ok to test

PR #1146, opened 2018-02-25
JBPM-6890: Adding the customTaskID as a unique identifier for an active

mcivantos-tribalyte:
Added processInstanceDescription as a businessId


mswiderski:
ok to test

mswiderski:
@mcivantos-tribalyte I had to revert this commit as it was causing compilation errors in droolsjbpm-integration module. Could you please create it once again, and please also add references between PRs if there are any so they won't be merged separately as it can cause problems (like in this case)

mcivantos-tribalyte:
@mswiderski  Ok. I'll send again and link to the other one related to it.

mcivantos-tribalyte:
@mswiderski Here, you have the new one
https://github.com/kiegroup/jbpm/pull/1148

PR #1148, opened 2018-03-04
JBPM-6890: Adding the customTaskID as a unique identifier for an active

mcivantos-tribalyte:
Take into account that this PR is related to https://github.com/kiegroup/droolsjbpm-integration/pull/1377


mswiderski:
ok to test

PR #1150, opened 2018-03-05
JBPM-6580 - RestWorkItemHandler creates a HttpClient for each call

tsurdilo:


tsurdilo:
@mswiderski the difference between WebServiceWorkItemHandler and RESTWorkItemHandler is that WebServiceWorkItemHandler does not explicitly close the client where RESTWorkItemHandler does. So the best way I could think of was to tell RESTWorkItemHandler to not close client with a workitem parameter. If there is some better way please let me know. Thanks.

mswiderski:
@tsurdilo I was thinking maybe we could make it:
- optional so we preserve the same behavior as is - new http client for each call
- user needs to provide the name/key of the client so it can be cached
- in general I'd say one http client per handler should be enough, no need to use map

so it would work if user gives the name/key when registering handler it will then create and cache http client instance otherwise it will work as right now, wdyt?

Please also not that in such case http client will be used my many threads so there might be some settings needed, like number of concurrent threads

tsurdilo:
@mswiderski updated. thanks for help!

tsurdilo:
Jenkins retest please

tsurdilo:
don't think failed tests have to do w/ the pr. will try look into whats going on

mswiderski:
Jenkins retest this

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -763,10 +884,31 @@ protected HttpClient getHttpClient(Integer readTimeout,
...
+
+        return cachedClient;
+    }
+
     protected void close(HttpClient httpClient,
                          Object httpMethod) throws IOException {
         if (HTTP_CLIENT_API_43) {
-            ((CloseableHttpClient) httpClient).close();
+            // dont close if we are using pooled connections
+            // this will be closed in abort
```
mswiderski: misleading comment - it's not closed in abort :)

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -738,6 +851,14 @@ public void abortWorkItem(WorkItem workItem,
 
     protected HttpClient getHttpClient(Integer readTimeout,
                                        Integer connectTimeout) {
+        if (getDoCacheClient() && HTTP_CLIENT_API_43) {
+            if (cachedClient != null) {
+                return cachedClient;
+            } else {
+                getNewPooledHttpClient(readTimeout,
```
mswiderski: shouldn't this return the newly created client? otherwise it will be overridden down the road of this method, isn't it?

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -317,7 +430,7 @@ public void executeWorkItem(WorkItem workItem,
                 close(httpClient,
                       methodObject);
             } catch (Exception e) {
-                // no idea if this throws something, but we still don't care!
+                handleException(e);
```
mswiderski: thank you!

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -133,6 +141,11 @@
         } catch (ClassNotFoundException e) {
             HTTP_CLIENT_API_43 = false;
         }
+
+        // setup pooling connection manager
+        connectionManager = new PoolingHttpClientConnectionManager();
+        connectionManager.setMaxTotal(500);
+        connectionManager.setDefaultMaxPerRoute(50);
```
mswiderski: would make sense to have this configurable though it would mean yet another (or more) constructors...

tsurdilo:
@mswiderski updated. thank you!

tsurdilo:
Jenkins retest please

tsurdilo:
Jenkins retest please

MarianMacik:
@mswiderski Sorry, Maciej, didn't catch your comment, but looks good anyway! Nice tests, @tsurdilo.

PR #1154, opened 2018-03-09
RHBPMS-5141: Reuse XStream instance to improve performance

martinweiler:


hifly81:
tested with 1000 process instances without problems

tsurdilo:
This is going in without a test just like the master commit https://github.com/kiegroup/jbpm/commit/053b766a1a8bd310ab21585bd9e8e816e4914bbb. Is that OK? Just asking cause it may be not feasible to write a test for this??

MarianMacik:
I think we can at least do something similar like [here](https://github.com/kiegroup/jbpm/pull/1150/files#diff-463a9e7cb992241a16a74c029fa9eb1aR80). Here it would be a simple check that XStream was initialized (initXStream method was called exactly once).

PR #1155, opened 2018-03-14
RHBA-519: Cache AsyncAuditLogProducer in DefaultRegisterableItemsFact…

martinweiler:
…ory to improve performance


mswiderski:
@martinweiler would it be possible to add simple test - just to make sure we cover it with some verification?

martinweiler:
@mswiderski sure, I'll try to add a test. Do you happen to have a good pointer for an existing test that I could use as a starting point?

mswiderski:
@martinweiler maybe just add simple test case for DefaultRegisterableItemsFactory that will mock things needed to get this to build a new instance and cache it.

martinweiler:
@mswiderski test case added, please review. thanks!

mswiderski:
Jenkins retest this

jbpm-runtime-manager/src/test/resources/jbpm.audit.jms.properties
```diff
@@ -0,0 +1,2 @@
+
```
mswiderski: is this intentionally left empty? if so maybe just add a comment about it so it won't be removed by accident

mswiderski:
Jenkins retest this

martinweiler:
@mswiderski thanks, let me know in case any other formatting changes are required.

mswiderski:
Jenkins retest this

PR #1160, opened 2018-03-20
JBPM-7043 - Process migration should fail if strategy, JBPM-7047 - case-data not saved in CaseFileDataLog

mswiderski:


mswiderski:
Jenkins retest this

PR #1161, opened 2018-03-20
RHBA-592 - KieSession already used when starting process with PER_PRO…

mswiderski:
…CESS_INSTANCE strategy

it turned out the issues is with not checking the results of found context which is actually the right one. So that seems to be related in dealing with persistence context on tomcat to some extend. Though the validation logic was truly not complete.

@sutaakar could you please have a look and try if possible as we don't run PR jobs on tomcat. I tried it myself but would be good to have another pair of eyes (and hands) on it


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1166, opened 2018-03-23
JBPM-7069 - Manage Tasks only display tasks if user is set/member of …

cristianonicolai:
…Kie Server Human task admin system property


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/BusinessAdminTasksPreprocessor.java
```diff
@@ -49,6 +54,11 @@ public void preprocess(DataSetLookup lookup) {
             return;
         }
 
+        if (ADMIN_USER.equals(identityProvider.getName()) ||
+                identityProvider.getRoles().stream().filter(s -> s.equals(ADMIN_GROUP)).findFirst().isPresent()) {
+            return;
```
mswiderski: @cristianonicolai I am just wondering about this in context of what we discussed in the jira - as this will essentially disable query by user name and group, isn't it? I mean if user is not the global admin nor is part of global Administrators group then query will not have additional filter by his/her user name and groups and thus might lead to either:
- returning tasks without org entity checks
- returning no tasks at all even if given user is business admin but not a global one

is that what we want with this change?

bsremac: @mswiderski My understanding regarding the desired behavior is:

- If the user is either the global admin or is part of global Administrators group, then all tasks should be returned without org entity checks

- If the user is the business admin, but not the global one, then the org entity check should kick in and return only appropriate tasks which the user is allowed to see

mswiderski: @bsremac you're right and now it all makes sense, thanks!

PR #1168, opened 2018-03-27
RHBA-85 - NPE when trying to list variables for an aborted process

mswiderski:


jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceImplTest.java
```diff
@@ -1096,4 +1096,29 @@ public void testStartAndSignalBoundaryWithExpression() {
 		pi = processService.getProcessInstance(processInstanceId);
 		assertNull(pi);
 	}
+	
+    @Test
```
tsurdilo: Maybe better to do @Test(expected = ProcessInstanceNotFoundException.class) instead of the try/catch inside?

mswiderski: @tsurdilo I'd say matter of preference and I tried both and somehow think this is more readable at least for me :)

tsurdilo: @mswiderski ok :)

MarianMacik: @tsurdilo @mswiderski AFAIK this is not available in jUnit 5 (expected = ), so I would either do it this way or better use AssertJ for that, which I understand is not an ideal solution when there is a big class with jUnit Asserts.

PR #1176, opened 2018-04-05
RHBA-248 - LogCleanupCommand follow-up

MarianMacik:
@mswiderski when playing a little bit with LogCleanupCommand I found that if you skip task logs and delete only process logs, it behaves correctly, but when you then try to delete everything including task logs, task logs are not deleted because they don't have active process instances in ProcessInstanceLog table. So my proposal is to delete them too since they are orphans now.

Quick summary of changes:
* Delete everything which either belongs to non-active process instance or doesn't belong to any - it now applies to NodeInstanceLog, VariableInstanceLog, TaskEventImpl and AuditTaskImpl (it worked with Node/VariableInstanceLog before since it is deleted together with ProcessInstanceLog, so there are no orphans left, only task-related things can have orphans)
* Unignored remaining LogCleanupCommand tests - they were ignored for nearly 3 years since they were brought to community from QE internal repo, but they seem to be running well

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/jobexec/LogCleanupCommandTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/jobexec/LogCleanupCommandTest.java
@@ -28,13 +28,11 @@ import org.jbpm.runtime.manager.impl.jpa.EntityManagerFactoryManager;
 import org.jbpm.services.task.audit.service.TaskJPAAuditService;
 import org.jbpm.test.JbpmAsyncJobTestCase;
 import org.jbpm.test.listener.CountDownAsyncJobListener;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.kie.api.runtime.KieSession;
 import org.kie.api.runtime.process.ProcessInstance;
 import org.kie.api.task.TaskService;
 import org.kie.internal.executor.api.CommandContext;
-
 import qa.tools.ikeeper.annotation.BZ;
 
 /**
@@ -111,7 +109,6 @@ public class LogCleanupCommandTest extends JbpmAsyncJobTestCase {
         Assertions.assertThat(getNodeInstanceLogSize(HELLO_WORLD_ID)).isPositive();
     }
 
-    @Ignore
     @Test(timeout=10000)
     public void skipTaskLog() throws Exception {
         KieSession kieSession = null;
@@ -166,7 +163,6 @@ public class LogCleanupCommandTest extends JbpmAsyncJobTestCase {
         // Verify presence of data
         Assertions.assertThat(getProcessLogSize(runProcess)).isPositive();
         Assertions.assertThat(getNodeInstanceLogSize(runProcess)).isPositive();
-//        Assertions.assertThat(getExecutorLogSize()).isZero(); // TBD: Should be zero but is > zero
 
         // Set to NOW if date was not provided
         if (date == null) {
@@ -188,7 +184,6 @@ public class LogCleanupCommandTest extends JbpmAsyncJobTestCase {
         Assertions.assertThat(getNodeInstanceLogSize(HELLO_WORLD_ID)).isPositive();
     }
 
-    @Ignore
     @Test(timeout=10000)
     @BZ("1190881")
     public void deleteAllLogsOlderThanNow() throws Exception {
@@ -200,7 +195,6 @@ public class LogCleanupCommandTest extends JbpmAsyncJobTestCase {
         Assertions.assertThat(getNodeInstanceLogSize(HELLO_WORLD_ID)).isZero();
     }
 
-    @Ignore
     @Test(timeout=10000)
     public void deleteAllLogsOlderThanTomorrow() throws Exception {
         CountDownAsyncJobListener countDownListener = new CountDownAsyncJobListener(1);
@@ -221,7 +215,6 @@ public class LogCleanupCommandTest extends JbpmAsyncJobTestCase {
         Assertions.assertThat(getNodeInstanceLogSize(HELLO_WORLD_ID)).isPositive();
     }
 
-    @Ignore
     @Test(timeout=10000)
     @BZ("1190881")
     public void deleteAllLogsOlderThanPeriod() throws Exception {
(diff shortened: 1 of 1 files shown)


jbpm-audit/src/main/java/org/jbpm/process/audit/query/AbstractAuditDeleteBuilderImpl.java
```diff
@@ -32,10 +32,12 @@
...
 public abstract class AbstractAuditDeleteBuilderImpl<T> extends AbstractDeleteBuilderImpl<T> implements AuditDeleteBuilder<T> {
 
-    protected static String ONLY_COMPLETED_PROCESS_INSTANCES = 
+    protected static String ONLY_COMPLETED_PROCESS_INSTANCES_OR_ORPHANS =
             " l.processInstanceId in (select spl.processInstanceId \n"
             + "FROM ProcessInstanceLog spl \n"
-            + "WHERE spl.status in (2, 3))";
+            + "WHERE spl.status in (2, 3)) \n"
+            + "OR l.processInstanceId not in (select spl.processInstanceId \n "
+            + "FROM ProcessInstanceLog spl)";
```
mswiderski: I am afraid this can be really heavy operation when there are millions of records in the ProcessInstanceLog table... and since this can be rare case when people clean up process instances without cleaning tasks this can be simply an edge case which might not be worth to impact all typical use cases, wdyt?

MarianMacik: Do you think it would impact performance a lot when there is already an index on processInstanceId? Right now DB has to check at first if there is a PID in table with only active PI and secondly if it is not present at all in PI log. If there is an index on processInstanceId, it might be quite quick to determine if there is a hit or not. But maybe I look at it very simply. Anyway, at least I tried, but if you think it is not worth it, you can close it, I have no problem with it ;)

MarianMacik: OK, I did a little research and yes, it is killing performance quite badly - but bear in mind that if there are thousand of running processes, it would kill it too. And what about consistency? Shouldn't AuditTaskImpl processInstanceId reference processInstanceId in log by foreign key? Because if you skip Task log in that command, task would point at non-existent process.

MarianMacik: OK, apparently this was a very good exercise. NOT IN has to check also NULL values, therefore it has to go through whole table (because 5 != NULL means unknown, which means false in where clause :) ). So I changed it for NOT EXISTS which is much much quicker. After executing scripts for ProcessInstanceLog table which build all indexes, I was able to compare these 2 queries:

```

EXPLAIN DELETE
FROM AuditTaskImpl l
WHERE l.processInstanceId IN (SELECT spl.processInstanceId
                              FROM ProcessInstanceLog spl
                              WHERE spl.status IN ( 2, 3))
      OR
      l.processInstanceId NOT IN (SELECT spl.processInstanceId
                                  FROM ProcessInstanceLog spl);
```
produces
```
Delete on audittaskimpl l  (cost=8.61..60070689.06 rows=38144 width=6)
  ->  Seq Scan on audittaskimpl l  (cost=8.61..60070689.06 rows=38144 width=6)
        Filter: ((hashed SubPlan 1) OR (NOT (SubPlan 2)))
        SubPlan 1
          ->  Index Scan using idx_pinstlog_status on processinstancelog spl  (cost=0.29..8.61 rows=1 width=8)
                Index Cond: (status = ANY ('{2,3}'::integer[]))
        SubPlan 2
          ->  Materialize  (cost=0.00..2234.99 rows=50866 width=8)
                ->  Seq Scan on processinstancelog spl_1  (cost=0.00..1781.66 rows=50866 width=8)

```

Now the correct version using NOT EXISTS instead of NOT IN

```
EXPLAIN DELETE
FROM AuditTaskImpl l
WHERE l.processInstanceId IN (SELECT spl.processInstanceId
                              FROM ProcessInstanceLog spl
                              WHERE spl.status IN (2,3))
      OR NOT EXISTS (SELECT spl.processInstanceId
                     FROM ProcessInstanceLog spl
                     WHERE l.processinstanceid = spl.processinstanceid);
```

produces

```
Delete on audittaskimpl l  (cost=8.61..424652.49 rows=38144 width=6)
  ->  Seq Scan on audittaskimpl l  (cost=8.61..424652.49 rows=38144 width=6)
        Filter: ((hashed SubPlan 1) OR (NOT (SubPlan 2)))
        SubPlan 1
          ->  Index Scan using idx_pinstlog_status on processinstancelog spl  (cost=0.29..8.61 rows=1 width=8)
                Index Cond: (status = ANY ('{2,3}'::integer[]))
        SubPlan 2
          ->  Index Only Scan using idx_pinstlog_pinstid on processinstancelog spl_1  (cost=0.29..8.31 rows=1 width=0)
                Index Cond: (processinstanceid = l.processinstanceid)

```

Original SQL query:
```
EXPLAIN DELETE
FROM AuditTaskImpl l
WHERE l.processInstanceId IN (SELECT spl.processInstanceId
                              FROM ProcessInstanceLog spl
                              WHERE spl.status in (2,3))
```

produces

```
Delete on audittaskimpl l  (cost=8.62..2147.72 rows=1 width=12)
  ->  Hash Semi Join  (cost=8.62..2147.72 rows=1 width=12)
        Hash Cond: (l.processinstanceid = spl.processinstanceid)
        ->  Seq Scan on audittaskimpl l  (cost=0.00..2005.59 rows=50859 width=14)
        ->  Hash  (cost=8.61..8.61 rows=1 width=14)
              ->  Index Scan using idx_pinstlog_status on processinstancelog spl  (cost=0.29..8.61 rows=1 width=14)
                    Index Cond: (status = ANY ('{2,3}'::integer[]))
```

So cost is apparently much much lower. But anyway, it was a good exercise. At least I have found ignored tests which work so will alter the PR to just unignore them, wdyt?

MarianMacik: Ok, my last call, I have managed to optimize it a little bit :)
```
EXPLAIN DELETE
FROM AuditTaskImpl l
WHERE NOT EXISTS (SELECT spl.processInstanceId
                              FROM ProcessInstanceLog spl
                              WHERE l.processinstanceid = spl.processinstanceid
                              AND spl.status NOT IN (2,3))
```

it now gives:

```
Delete on audittaskimpl l  (cost=2793.65..5864.11 rows=1 width=12)
  ->  Hash Anti Join  (cost=2793.65..5864.11 rows=1 width=12)
        Hash Cond: (l.processinstanceid = spl.processinstanceid)
        ->  Seq Scan on audittaskimpl l  (cost=0.00..2005.59 rows=50859 width=14)
        ->  Hash  (cost=1908.83..1908.83 rows=50866 width=14)
              ->  Seq Scan on processinstancelog spl  (cost=0.00..1908.83 rows=50866 width=14)
                    Filter: (status <> ALL ('{2,3}'::integer[]))

```

I think the original one might be more costly if there are more completed processes than active ones, mine will be more costly if there are more active processes I guess.

MarianMacik:
I have altered the PR to just unignore tests which seem to be working + I have altered one test to have a correct behaviour.

MarianMacik:
jenkins retest this

PR #1179, opened 2018-04-10
JBPM-5223 - Improve WorkItemHandler parameter definition

tsurdilo:
Added "required" attribute to the WidParameter allowing users to set certain handler input params to be set as required (handler will not execute if required params cannot be found) 

Updated WidProcessor to generate a parameter validation class for each of the handlers. This class checks existence of the required parameters and will throw exception if they cannot be found.

Added ValidatingAbstractLogOrThrowWorkitemHandler which workitem handlers can extend. It provides a validate method that will call the handlers validation class

Small changes to workitem archetype to add generation of the deployment descriptor

### DIFF
--- a/jbpm-workitems/jbpm-workitems-archetype/src/main/resources/archetype-resources/jbpm-workitems-__rootArtifactId__/src/main/resources/kie-ddtemplate.st
+++ b/jbpm-workitems/jbpm-workitems-archetype/src/main/resources/archetype-resources/jbpm-workitems-__rootArtifactId__/src/main/resources/kie-ddtemplate.st
@@ -1,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<deployment-descriptor xsi:schemaLocation="http://www.jboss.org/jbpm deployment-descriptor.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <persistence-unit>org.jbpm.domain</persistence-unit>
+    <audit-persistence-unit>org.jbpm.domain</audit-persistence-unit>
+    <audit-mode>JPA</audit-mode>
+    <persistence-mode>JPA</persistence-mode>
+    <runtime-strategy>SINGLETON</runtime-strategy>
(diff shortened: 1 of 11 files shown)


tsurdilo:
example on how to use required param on the WidParameter annotation and validate the required params can be found here https://github.com/kiegroup/jbpm-work-items/pull/31

tsurdilo:
@mswiderski updated - created util class users can call rather than writing the annotation retrieval code themselves, for example in executeWorkitem:

RequiredParameterValidator.validate(this.getClass(), workItem);

jbpm-workitems/jbpm-workitems-archetype/src/main/resources/archetype-resources/jbpm-workitems-__rootArtifactId__/src/main/resources/kie-ddtemplate.st
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<deployment-descriptor xsi:schemaLocation="http://www.jboss.org/jbpm deployment-descriptor.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+    <persistence-unit>org.jbpm.domain</persistence-unit>
+    <audit-persistence-unit>org.jbpm.domain</audit-persistence-unit>
+    <audit-mode>JPA</audit-mode>
+    <persistence-mode>JPA</persistence-mode>
+    <runtime-strategy>SINGLETON</runtime-strategy>
```
MarianMacik: Does this mean runtime strategy is not configurable? Or it does not matter?

tsurdilo: That's just the default that the workbench uses as well. Users are free to change before copy/paste into their project if they wish, or they can just take the handlers part out of it too.

PR #1180, opened 2018-04-11
JBPM-7062 - Not possible to send a Signal Data using Signal Intermedi…

mswiderski:
…ate Event


tsurdilo:
@mswiderski thank you!

PR #1187, opened 2018-04-16
 JBPM-7164: Add support to TypedWorkItem(Handler)s

evacchi:
Proposal for extending the WorkItem(Handler) models with support for typed POJOs for parameters.

This PR contains an example for a typed version of the Email WorkItem

Depends on https://github.com/kiegroup/drools/pull/1869


evacchi:

tsurdilo:
I don't think we should add this jbpm-workitems-email-typed. Any new workitems should be added to the https://github.com/kiegroup/jbpm-work-items. I think the use of this should be in tests only since email workitem already exists and uses can then chose if they want to use the *typed interfaces or not when they create their new ones. wdyt?

jbpm-workitems/jbpm-workitems-core/src/main/java/org/jbpm/process/workitem/core/AbstractLogOrThrowHandler.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Map;
+
+import org.jbpm.bpmn2.handler.WorkItemHandlerRuntimeException;
+import org.jbpm.workflow.instance.node.WorkItemNodeInstance;
+import org.kie.api.runtime.process.NodeInstance;
+import org.kie.api.runtime.process.NodeInstanceContainer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class AbstractLogOrThrowHandler {
```
tsurdilo: Whats the point of just copying existing class here ?

jbpm-workitems/jbpm-workitems-core/src/main/java/org/jbpm/process/workitem/core/AbstractLogOrThrowTypedWorkItemHandler.java
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.process.workitem.core;
+
+import org.kie.api.runtime.process.TypedWorkItem;
+import org.kie.api.runtime.process.TypedWorkItemHandler;
+
+public abstract class AbstractLogOrThrowTypedWorkItemHandler<W extends TypedWorkItem<?, ?>> extends AbstractLogOrThrowHandler implements TypedWorkItemHandler<W> {
```
tsurdilo: Sorry I have a question here maybe I don't understand something - in the doc for this it says "Seamlessly convert between TypedWorkItem<?,?> to WorkItem to retain backwards compatiblity ad interim." - on the contrary this seems somewhat invasive as workitems have to declare use of typed or not, not sure where the backwards compat is used - please explain.

jbpm-workitems/jbpm-workitems-email-typed/pom.xml
```diff
@@ -0,0 +1,70 @@
+<?xml version="1.0"?>
+<project
```
tsurdilo: I don't think adding a copy of the existing email workitem here is good. it will confuse users more than anything to have two of the exact same thing. (same comment already added :) )

jbpm-workitems/jbpm-workitems-email-typed/src/main/java/org/jbpm/process/workitem/email/typed/Connection.java
```diff
@@ -0,0 +1,129 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.jbpm.process.workitem.email.typed;
+
+import java.lang.reflect.Field;
+
+public class Connection {
```
tsurdilo: same here..not needed imo

jbpm-workitems/jbpm-workitems-email-typed/src/main/java/org/jbpm/process/workitem/email/typed/Connection.java
```diff
@@ -0,0 +1,129 @@
...
+
+    public Connection() {
+    }
+
+    public Connection(String host,
```
tsurdilo: same here..not needed imo

jbpm-workitems/jbpm-workitems-email-typed/src/main/java/org/jbpm/process/workitem/email/typed/Email.java
```diff
@@ -0,0 +1,72 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+
+package org.jbpm.process.workitem.email.typed;
+
+public class Email {
```
tsurdilo: same here..not needed imo

mswiderski:
I agree with @tsurdilo copied email work item project does not make sense here and for sure should not be copied with all pieces that original has. Frankly speaking I'd drop these changes from here and if needed add new one project that uses typed version in jbpm-workitems repository.

MarianMacik:
I agree with @mswiderski and @tsurdilo.

evacchi:
So do not intend this PR as code that *has* to be merged, rather it's an example of an *alternative implementation* of the Email work item to show how the Typed version works, and to start a discussion. The reason why the code is duplicated is to be self-contained (I did also consider to share this in a separate repository)

Anyway, the doc @tsurdilo is quoting

> Seamlessly convert between TypedWorkItem to WorkItem to retain backwards compatiblity ad interim

refers to another PR https://github.com/kiegroup/drools/pull/1869. The TypedWorkItem feature is *of course* an opt-in, user-facing feature, so this *is intended* to impact the way user interact *with* the engine; OTOH, what's seamless is the interaction of the *feature with engine*: to the engine, the use of a typed vs. a traditional map-based WorkItem is *transparent*, as the conversion back and forth WorkItems and TypedWorkItems is done at the very end of the dispatching mechanism, [in the handler](https://github.com/evacchi/drools/blob/JBPM-7164/drools-core/src/main/java/org/drools/core/process/instance/impl/UntypedWorkItemHandlerImpl.java)

Please do share additional thoughts

mswiderski:
> The TypedWorkItem feature is of course an opt-in, user-facing feature, so this is intended to impact the way user interact with the engine

can you explain it? How will users use this feature? Please take into account process engine and not only directly calling work item handler as users do not do it.


> OTOH, what's seamless is the interaction of the feature with engine: to the engine, the use of a typed vs. a traditional map-based WorkItem is transparent

where is this transparent mechanism implemented?

evacchi:
I have decided to move the example out of this repo. I will follow up with a new PR, with a slightly different design for the feature.

PR #1193, opened 2018-04-18
[6.5.x] RHBPMS-5154: Test case for DisabledFollowOnLockOracle10gDialect

martinweiler:


mswiderski:
Jenkins retest this

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/ExecutorQueryServiceTest.java
```diff
@@ -0,0 +1,163 @@
...
+ * -Dmaven.jdbc.db.server=localhost 
+ * -Dmaven.jdbc.db.port=1521 
+ * -Dmaven.jdbc.db.name=ORCLPDB1 
+ * -Dmaven.jdbc.url=jdbc:oracle:thin:@localhost:1521/ORCLPDB1 
+ * -Dmaven.jdbc.schema=jboss
+ */
+public class ExecutorQueryServiceTest {
+
+	protected ExecutorService executorService;
+	private static final int THREADS = 2;
```
MarianMacik: Just one question. Are 2 threads enough? I personally would increase it to at least 4. But just a suggestion :) Wdyt @mswiderski.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/jpa/ExecutorQueryServiceImpl.java
```diff
@@ -236,9 +236,12 @@ public RequestInfo execute(Context context) {
...
 
-                    Number lockedRequest = ctx.nativeQueryAndLockWithParametersInTransaction("NativePendingRequestsForProcessing", params, true, Number.class);
-                    if (lockedRequest != null) {
-                        request = ctx.find(org.jbpm.executor.entities.RequestInfo.class, lockedRequest.longValue());
+                    Object[] lockedRequest = (Object[])ctx.nativeQueryAndLockWithParametersInTransaction("NativePendingRequestsForProcessing", params, true, Object.class);
+                    if (lockedRequest != null && lockedRequest.length > 0) {
+                        try {
+                            request = ctx.find(org.jbpm.executor.entities.RequestInfo.class, ((Number)lockedRequest[0]).longValue());
+                        } catch(NumberFormatException nfe) {
+                        }
```
MarianMacik: Shouldn't we include there at least a warning log message? Empty catch blocks are really not a good practice even if we know that there is a very little chance of throwing that exception.

mareknovotny:
@MarianMacik do you still insist on your comments before merging, right?

MarianMacik:
@mareknovotny Well, at least I would like to hear opinion on that from @martinweiler. Wdyt?

martinweiler:
@MarianMacik - 2 threads are ok to reproduce the issue. For the NFE, I have added some logging.

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/ExecutorQueryServiceTest.java
```diff
@@ -0,0 +1,163 @@
...
+ *
+ * Running the test case with default H2 database succeeds, as only one thread succeeds in obtaining the RequestInfo entity
+ *
+ * To run the test with Oracle, pass in the following system parameters:
+ *
+ * // it is expected that the test case fails with default oracle dialect:
+ * -Dmaven.hibernate.dialect=org.hibernate.dialect.Oracle10gDialect
+ * 
+ * // the test should be successful using the custom dialect
+ * -Dmaven.hibernate.dialect=org.jbpm.persistence.jpa.hibernate.DisabledFollowOnLockOracle10gDialect
```
MarianMacik: Thanks, @martinweiler. Just one last question - the issue was with the DisabledFollowOnLockOracle10gDialect not working properly, but Oracle10gDialect working as intended? Because the JavaDoc says that it should fail with org.hibernate.dialect.Oracle10gDialect. Is this still true?

martinweiler: org.hibernate.dialect.Oracle10gDialect is not working as intended, as it does not lock a job properly. You will see 'HHH000444: Encountered request for locking however dialect reports that database prefers locking be done in a separate select' WARN messages, and the test will fail as the job is executed on both threads.

MarianMacik: Then I see one issue with the test. When we run full certification matrix of jbpm, we use Oracle10gDialect which is the first to be supported afaik. Is this DisabledFollowOnLockOracle10gDialect supported too? Anyway, this means that our certification matrix will fail with the Oracle database. So do we still have a bug with default Oracle10gDialect or this use case is not supported with the default Oracle10gDialect?

martinweiler: I guess then we need to disable this test in the default test suite. See https://bugzilla.redhat.com/show_bug.cgi?id=1234592 as well as https://hibernate.atlassian.net/browse/HHH-1168?focusedCommentId=48846 for background information on why this custom dialect was necessary to begin with.

mareknovotny: @martinweiler Are you going to disable the test in this PR then?

mareknovotny:
Merged as https://github.com/kiegroup/jbpm/commit/d1d20e701099f3e6c5822a552ea01f33f8effbfc and added additional `@Ignore` in https://github.com/kiegroup/jbpm/commit/2816ff70d3b91af572b23203c00396897f06b293

PR #1194, opened 2018-04-18
JBPM-6235 TaskInstanceView Test for NoSQL

MarianMacik:
Hi @mswiderski,

I have decided to add a test which tests also TaskInstanceView.

The test added is just slightly altered [testProcess](https://github.com/MarianMacik/jbpm/blob/bab1b01a8eb7a3d4f9d306edb3c0f5ec6d6b2590/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/ProcessHumanTaskTest.java#L55) test so in theory I can just extend the existing one with TestEventEmitter check, but then it would test more than one thing... plus I would have to either enable session persistence for whole class or just override it in the existing process like I did in this PR.

So wdyt?


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

MarianMacik:
Jenkins retest this

PR #1203, opened 2018-04-24
JBPM-4646 - Covered remaining methods

MarianMacik:
Covered remaining methods:

* getCaseInstanceDataItemsByType
* getCaseInstanceDataItemsByName


MarianMacik:
@mswiderski Added second commit with API corrections.

PR #1211, opened 2018-05-01
JBPM-7209 - Allow multiple users and groups assignments to role when …

cristianonicolai:
…starting a case

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseService.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseService.java
@@ -430,7 +430,7 @@ public interface CaseService {
      * @param rolesAssignment initial role assignment
      * @return returns new instance (not associated with case) of CaseFileInstance populated with given data
      */
-    CaseFileInstance newCaseFileInstance(String deploymentId, String caseDefinition, Map<String, Object> data, Map<String, OrganizationalEntity> rolesAssignment);
+    CaseFileInstance newCaseFileInstance(String deploymentId, String caseDefinition, Map<String, Object> data, Map<String, OrganizationalEntity[]> rolesAssignment);
     
     /**
      * Builds and returns new CaseFileInstance with given data and roles assignments. Not yet associated with any case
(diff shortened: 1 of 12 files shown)


cristianonicolai:
Dependent PRs:
- https://github.com/kiegroup/jbpm-wb/pull/1107
- https://github.com/kiegroup/jbpm/pull/1211
- https://github.com/kiegroup/droolsjbpm-integration/pull/1456

cristianonicolai:
@mswiderski I agree this is probably looking to handle some edge cases, but it is for any use case where having multiple assignments to a single role is needed, which is something is already handle anyway by allowing people to define the role cardinality at design time. More important to me is that we do allow multiple assingment already by manipulating the case file after it started but not when starting a case. We are currently restricting users to only assingn one org entity at the start of the case, which I believe shouldn't be done at data model level but leaving at the validation of the cardinality in the role to define.

mswiderski:
@cristianonicolai the idea with this assignment at case start was to make it simple and at the same time promote better (more concise) designs. Obviously you could make your role assignment to as many users or groups you like but would it be really useful or with business justification?

Roles cardinality is for the entire case instance life cycle - meaning it should protect from assigning more people/groups at any point in time and not only at start case.

cristianonicolai:
@mswiderski I don't disagree that we should promote concise designs and in the case of roles that problably means keeping to a single assingment. To my understanding that is just different from preventing user from starting a case with multiple assingments if they need to. As you mentioned, at the end of the day, it comes down to business need, and in this case if people have a genuine reason for having multiple assingments, they get blocked by the data model we impose. 
There are a few inconsistencies/issues I see here from the API perspective too:

- CaseFile Builder _addUserAssignments_ and _addGroupAssignments_ implies in adding multiple assigments but you end up replacing it if adding to the same role.
- CaseFile allows you to set assignments for both users and groups, letting you set at least one user and one group to the same role. At the _CaseManagementServiceBase_ level we process user firstly and groups later to create the same map with OrgEntities. What this causes is that the user is overriden by the Group that comes later in processing.
- @bsremac also mentioed another possible issue with auto start tasks, when the proper role assignment isnt defined at the start of the case.

Perhaps we want to look into making these changes backwards compatible somehow if that is the main concern.

mswiderski:
> CaseFile allows you to set assignments for both users and groups, letting you set at least one user and one group to the same role. At the CaseManagementServiceBase level we process user firstly and groups later to create the same map with OrgEntities. What this causes is that the user is overriden by the Group that comes later in processing.

this will happen anyway as you can't have user and groups with same id so once you attempt to do that it's already invalid.

Anyway, if it will be possible to make it in backward compatible way then I am fine with it, though I would not like to break backward compatibility because of this.

MarianMacik:
Having read this conversation, I agree with @mswiderski. We should do it in a backward compatible way.

mswiderski:
since this requires either work or change in the approach, I am closing it for now to avoid accidental merge ...

PR #1216, opened 2018-05-07
JBPM-7213: Tasks First Due Date update writes null to task log

nmirasch:
Updated the task field update log message

### DIFF
--- a/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
+++ b/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -591,6 +591,12 @@ public class JPATaskLifeCycleEventListener extends PersistableEventListener impl
         
     }
 
+    public String getUpdateFieldLog(String fieldName, String previousValue, String value){
+        return "Updated "+ fieldName
+                + " {From: '"+ (previousValue!=null ? previousValue :  "" )
+                + "' to: '"+ (value!=null ? value :  "" ) + "'}" ;
+    }
+
     @Override
     public void afterTaskUpdatedEvent(TaskEvent event) {
         String userId = event.getTaskContext().getUserId();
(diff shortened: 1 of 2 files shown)


nmirasch:
@cristianonicolai, @mswiderski  updated the task log message to take in account the null value. Changed the date format to match newvalue and old value. 
Could you take a look? Thanks!

jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/TaskAuditBaseTest.java
```diff
@@ -319,8 +319,10 @@ private void testDescriptionUpdate(String oldDescription, String newDescription,
         List<TaskEvent> taskEvents = taskAuditService.getAllTaskEvents(taskId, new QueryFilter());
         if (changeExpected) {
             Assertions.assertThat(taskEvents).hasSize(2);
-            Assertions.assertThat(taskEvents.get(1).getMessage()).contains(String.valueOf(oldDescription),
-                    String.valueOf(newDescription));
+            assertEquals(getUpdateFieldLogMessage("Description",
```
cristianonicolai: @nmirasch just a minor comment here, I think would be better if in the expected value we had the whole String instead of using an auxiliary method to build it. It would make the test easier to read as well, wdyt?

nmirasch:
@cristianonicolai changed the test to see full update log test result.

cristianonicolai:
@nmirasch tests look good now but I've to confess that it feels strange to have from: or to: followed by an empty String.  I know we don't want to add null there but maybe we need something in between. Any thoughts?

nmirasch:
@cristianonicolai It looks strange to me too. There are two points here.. How to differentiate between empty(null) value and '', Or between empty(null) and 'empty' as string value . Another consideration is how to find the way to express that with dates and strings.
Possible midterm solutions would be generate something like: 
  "Updated Name {From: value to: empty}",
  "Updated Name {From: value to: ''}" (add '' as a empty value?)
  "Updated Name {removed: value}"
Another thought is related with the internationalization for the future :  We could think in passing in this case the fieldname, old value and the new value to (TaskEventImpl) and let the ui manage this and other taskEvents log generation..

cristianonicolai:
@nmirasch good points, I think my preference would be something along the lines of the last option. @mswiderski  wdyt?
Regarding i18n, totally agree, we need the data model available in the event to be able to translate the message into different languages. Perhaps that could be another enhancement JIRA?

mswiderski:
yeah, the last one looks the most complete but to some extend I have a feeling we just do too much ... especially adding i18n for it, maybe we could just make it like

```
From 'value' to 'value'
```

so in case there is no value or is null it will end up to be

```
From 'aaa' to ''
```

as that will actually represent current state - I would not differentiate between null or empty string at all.

cristianonicolai:
Agree, perhaps to keep it simple just add the single quotes around the values so null end up as ''.

jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/TaskAuditBaseTest.java
```diff
@@ -319,45 +321,65 @@ private void testDescriptionUpdate(String oldDescription, String newDescription,
         List<TaskEvent> taskEvents = taskAuditService.getAllTaskEvents(taskId, new QueryFilter());
         if (changeExpected) {
             Assertions.assertThat(taskEvents).hasSize(2);
-            Assertions.assertThat(taskEvents.get(1).getMessage()).contains(String.valueOf(oldDescription),
-                    String.valueOf(newDescription));
+            assertEquals(expectedMessage,
+                         taskEvents.get(1).getMessage());
```
MarianMacik: I would not mix jUnit Asserts with AssertJ. Would you mind changing it back to AssertJ, please? `Assertions.assertThat(taskEvents.get(1).getMessage()).isEqualTo(expectedMessage);` And in other places too, please. Thanks.

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
```diff
@@ -591,6 +591,12 @@ public void beforeTaskUpdatedEvent(TaskEvent event) {
         
     }
 
+    public String getUpdateFieldLog(String fieldName, String previousValue, String value){
+        return "Updated "+ fieldName + " {"+
+                (previousValue != null? "From: '"+ previousValue + "' to: " :  "" ) +
+                "'"+ (value != null ? value :  "" ) + "'}" ;
```
MarianMacik: Correct me if I am wrong, but in case previousValue is null and new value is 'value' it will output:
`Updated fieldName {'value'}`. Don't we want to include the 'From... to' clause in all cases?

jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/TaskAuditBaseTest.java
nmirasch: @MarianMacik in this test there is a lot of assertEquals applied, you mean change that on the ones that my changes introduced, right?

MarianMacik: Sure. No need to rewrite whole class :) In general we try to use AssertJ in newly added tests as it not only provides more methods but also it is easier to read.

nmirasch:
Once the char ' is introduced, we need this related fix https://github.com/kiegroup/jbpm-wb/pull/1124

@cristianonicolai  @MarianMacik Changed the logs generation including  always the From even there is an empty value refered. Updated the test including @MarianMacik comments.

PR #1219, opened 2018-05-14
Fix TotalCompletionTimeLoadCalculator

MarianMacik:
@mswiderski As we discussed, here is the bug fix with a test. I have also created common abstract class since I wanted to reuse some of the existing code.

PR #1220, opened 2018-05-14
[7.7.x] Fix TotalCompletionTimeLoadCalculator

MarianMacik:
Same as #1219, but for 7.7.x.


mswiderski:
@MarianMacik I don't think this is on list of blockers so not sure we should really include it here... severity of this rather low anyway, right?

MarianMacik:
Ah, sorry. You are right. I was too quick :) We can close this.

PR #1221, opened 2018-05-14
Downgrade Resteasy to fix jbpm-container tests on WebLogic

MarianMacik:
There is an issue https://issues.jboss.org/browse/RESTEASY-1748 with Resteasy on weblogic. The issue prevents using REST services so REST WIH tests will fail on WebLogic. I will be notified when the bug is fixed, so it should be only a temporary fix.


MarianMacik:
Jenkins retest this

PR #1223, opened 2018-05-15
JBPM-7277 - Node SLA is not updated for the right node

MarianMacik:
@mswiderski As discussed, here is the fix. I have tested it on our Jenkins with all databases and it passed 3 times in a row. The reason why I don't include a test is that this test which was able to reproduce it in plain jBPM engine took more than minute for some databases (generating a big amount of data so query can clash wrong data), so I think it is not worth including it. We will still be able to verify it in KIE Server tests though. Thanks!

Marian


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
test failures not related

mswiderski:
Jenkins retest this

PR #1165, opened 2018-03-23
RHBA-611 - The system property org.jbpm.rm.init.timer should be set, JBPM-7071 - Get workitems from REST API call doesn't show up all

mswiderski:

PR #1225, opened 2018-05-16
[7.7.x] RHBA-611 - The system property org.jbpm.rm.init.timer should be set a…

MarianMacik:
…s false by default in jbpm ejb timer service

Backport of one commit from #1165 for 7.7.x.

@mswiderski AFAIK from JIRA, the second commit (JBPM-7071) is supposed to be fixed only on 7.8.x, so I think it is not needed on 7.7.x, right?


MarianMacik:
Jenkins retest this

PR #1226, opened 2018-05-21
[JBPM-7296] Signal on boundary event causes an unwanted event listener message

elguardian:
for boundary events, the cancel method should avoid firing and after node


mswiderski:
could you please add a test here https://github.com/kiegroup/jbpm/blob/master/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/IntermediateEventTest.java#L158

you can register custom ProcessEventListener on ksession and see if the boundary node instance will invoke it or not when being canceled.

mswiderski:
ok to test

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1233, opened 2018-05-29
JBPM-7323 - CMMN excution support for case management

mswiderski:
depends on 
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/762
https://github.com/kiegroup/droolsjbpm-knowledge/pull/311
https://github.com/kiegroup/drools/pull/1930


mswiderski:
Jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
```diff
@@ -152,8 +152,12 @@ public void removeAll() {
     }
 
     @Override
-    public Object getData(String name) {
-        return this.data.get(name);
+    public Object getData(String name) {        
+        Object result = this.data.get(name);
+        if (result == null && name != null && name.equalsIgnoreCase("caseid")) {
```
MarianMacik: I believe we can change ``` name != null && name.equalsIgnoreCase("caseid")  ``` for ``` "caseid".equalsIgnoreCase(name) ```. Will be much easier to read.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
```diff
@@ -376,6 +376,9 @@ public void signalEvent(String type, Object event) {
             workItemCompleted((WorkItem) event);
         } else if ("workItemAborted".equals(type)) {
             workItemAborted((WorkItem) event);
+        }else if (type.equals("RuleFlow-Activate" + getProcessInstance().getProcessId() + "-" + getNode().getMetaData().get("UniqueId"))) {
+            
+         trigger(null, org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
```
MarianMacik: Formatting.

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/java/org/jbpm/casemgmt/cmmn/xml/MilestoneHandler.java
```diff
@@ -0,0 +1,68 @@
...
+        super.handleNode(node, element, uri, localName, parser);
+        MilestoneNode milestoneNode = (MilestoneNode) node;        
+        ProcessBuildData buildData = (ProcessBuildData) parser.getData();
+        Map<String, PlanItem> planItems = (Map<String, PlanItem>) buildData.getMetaData("PlanItems");
+        
+        
+        PlanItem milestonePlanItem = planItems.get(milestoneNode.getMetaData("UniqueId"));
+        if (milestonePlanItem != null && milestonePlanItem.getExitCriterion() != null) {
+            
+            milestoneNode.setConstraint(milestonePlanItem.getExitCriterion().getExpression());
```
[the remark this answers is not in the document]
mswiderski: @MarianMacik  the issue is that milestone condition is to complete it or to say in different words to achieve it so it's actually exit criterion - as soon as it is met it will complete (and by that exit) milestone node. Moreover, we need to support entry criterion to activate the node so it can start tracking the exit condition.

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/java/org/jbpm/casemgmt/cmmn/xml/StageHandler.java
```diff
@@ -0,0 +1,86 @@
...
+        stageNode.setAutoComplete(false);
+        stageNode.setLanguage("rule");
+        
+        ProcessBuildData buildData = (ProcessBuildData) parser.getData();
+        Map<String, PlanItem> planItems = (Map<String, PlanItem>) buildData.getMetaData("PlanItems");
+                
+        PlanItem stagePlanItem = planItems.get(stageNode.getMetaData("UniqueId"));
+        if (stagePlanItem != null && stagePlanItem.getEntryCriterion() != null) {
+            if(AUTOCOMPLETE_EXPRESSIONS.contains(stagePlanItem.getEntryCriterion().getExpression())) {
+                stageNode.setAutoComplete(true);
```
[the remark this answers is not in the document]
mswiderski: you're right it should actually check - autostart instead of auto complete on entry and then check auto complete on exit, fixed.

mswiderski:
@MarianMacik thanks for the review, I think I addressed all the comments, please have a look

mswiderski:
Jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/java/org/jbpm/casemgmt/cmmn/xml/MilestoneHandler.java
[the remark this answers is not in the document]
MarianMacik: OK, just asking because in CMMN spec there is only milestone with entry criterion depicted, so is on the internet :) But I agree it makes sense the way we do it.

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/java/org/jbpm/casemgmt/cmmn/xml/SentryHandler.java
```diff
@@ -0,0 +1,107 @@
...
+		if ((this.validParents == null) && (this.validPeers == null)) {
+			this.validParents = new HashSet<>();
+			this.validParents.add(RuleFlowProcess.class);
+
+			this.validPeers = new HashSet<>();
+			this.validPeers.add(null);
+			this.validPeers.add(PlanItem.class);
+			this.validPeers.add(Sentry.class);
+
+			this.allowNesting = false;
```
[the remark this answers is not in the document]
MarianMacik: Hmm, I cannot see any change here, but GitHub shows it as if there was a change.

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/resources/META-INF/DI.xsd
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsd:schema xmlns:xsd="http://www.w3.org/2001/XMLSchema" 
+            xmlns:dc="http://www.omg.org/spec/CMMN/20151109/DC" 
+            xmlns:di="http://www.omg.org/spec/CMMN/20151109/DI"
+            targetNamespace="http://www.omg.org/spec/CMMN/20151109/DI"
+            elementFormDefault="qualified" 
+            attributeFormDefault="unqualified">
+	<xsd:import namespace="http://www.omg.org/spec/CMMN/20151109/DC" 
+                    schemaLocation="DC.xsd"/>
```
MarianMacik: One more thing I forgot. I see these schemas in both src/main/resources/META-INF and src/main/xsd. Why is it needed to have it duplicated? Because if I found it correctly, ExtensibleXMLParser in Drools looks for it in both directories anyway.

mswiderski: The xsd location is explicitly for OSGi

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/java/org/jbpm/casemgmt/cmmn/xml/SentryHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            this.validParents = new HashSet<>();
+            this.validParents.add(RuleFlowProcess.class);
+
+            this.validPeers = new HashSet<>();
+            this.validPeers.add(null);
+            this.validPeers.add(PlanItem.class);
+            this.validPeers.add(Sentry.class);
+
+            this.allowNesting = false;
+        }
```
MarianMacik: I think this comment has disappeared somehow and I haven't found a reply to it.

Is the ```if ((this.validParents == null) && (this.validPeers == null)) {``` relevant in these handlers? Because by looking at the code I cannot think of case where they are not null. BaseAbstractHandler does not initialize them nor there are any setters to set them. So I believe if we initialized them without checking, we might make it cleaner and quicker as well :) The IF is also omitted in AbstractCaseNodeHandler...

mswiderski: in general I simply followed the same as it was done for BPMN2 parsers so wanted to stay as close to it as possible ...

MarianMacik: OK. So should I understand it that don't touch it if it works? :)

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
```diff
@@ -152,8 +152,12 @@ public void removeAll() {
     }
 
     @Override
-    public Object getData(String name) {
-        return this.data.get(name);
+    public Object getData(String name) {        
+        Object result = this.data.get(name);
+        if (result == null && "caseid".equalsIgnoreCase(name)) {
+            result = getCaseId();
```
MarianMacik: What happens if I define my custom caseid as a case file item? Would I break it? And can't we call CaseFileInstance#getCaseId() method for that?

mswiderski: it won't break, this is just a shortcut mainly for CMMN that do not define "local" variables and only case file items. The idea here is to allow to use caseId as expressions #{caseId} for tasks names, descriptions, data inputs etc.

jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/java/org/jbpm/casemgmt/cmmn/xml/SentryHandler.java
mswiderski: yeah, that kind of thing :)

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
MarianMacik: OK, thanks.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
```diff
@@ -60,9 +60,11 @@ public void internalTrigger(NodeInstance from, String type) {
...
-    	agenda.activateRuleFlowGroup(
-			getRuleFlowGroupName(), getProcessInstance().getId(), getUniqueId());
+    	String ruleFlowGroup = getRuleFlowGroupName();
+    	if (ruleFlowGroup != null && !agenda.getRuleFlowGroup(ruleFlowGroup).isActive()) {
+        	agenda.getRuleFlowGroup(ruleFlowGroup).setAutoDeactivate(false);
+        	agenda.activateRuleFlowGroup(ruleFlowGroup, getProcessInstance().getId(), getUniqueId());
+    	}
 //    	if (getDynamicNode().isAutoComplete() && getNodeInstances(false).isEmpty()) {
 //    		triggerCompleted(NodeImpl.CONNECTION_DEFAULT_TYPE);
 //    	}
```
MarianMacik: I have just noticed this one when doing final review :)

PR #1234, opened 2018-05-30
RHPAM-1122 - Not possible to migrate process with recurring timer

mswiderski:


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
```diff
@@ -579,10 +584,19 @@ public Void execute(Context context) {
...
                             timerInstance.setDelay(delay);
-
+                            
+                            Trigger trigger = ((DefaultJobHandle)timerInstance.getJobHandle()).getTimerJobInstance().getTrigger();
+                            if (trigger instanceof IntervalTrigger && timerInstance.getPeriod() > 0) {
+                                if (timerInstance.getRepeatLimit() > 0) {
+                                    timerInstance.setRepeatLimit(((IntervalTrigger) trigger).getRepeatLimit() - ((IntervalTrigger) trigger).getRepeatCount());
+                                }
+                                delay = timerInstance.getPeriod() - (System.currentTimeMillis() - ((IntervalTrigger) trigger).getLastFireTime().getTime());
+                                timerInstance.setDelay(delay);
```
MarianMacik: Just to be sure, the delay will be reset to period value after the current trigger is fired, right?

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
```diff
@@ -592,6 +606,15 @@ public Void execute(Context context) {
...
                                 timerInstance.setDelay(delay);
+                                
+                                Trigger trigger = ((DefaultJobHandle)timerInstance.getJobHandle()).getTimerJobInstance().getTrigger();
+                                if (trigger instanceof IntervalTrigger && timerInstance.getPeriod() > 0) {
+                                    if (timerInstance.getRepeatLimit() > 0) {
+                                        timerInstance.setRepeatLimit(((IntervalTrigger) trigger).getRepeatLimit() - ((IntervalTrigger) trigger).getRepeatCount());
+                                    }
+                                    delay = timerInstance.getPeriod() - (System.currentTimeMillis() - ((IntervalTrigger) trigger).getLastFireTime().getTime());
+                                    timerInstance.setDelay(delay);
+                                }
```
MarianMacik: I think we could extract this to a method, wdyt?

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
mswiderski: as soon as the timer is cycle based it should rely on period

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
mswiderski: extracted to method

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
MarianMacik: So this will create an ordinary timer and only after that a cycle based one?

mswiderski: correct

PR #1235, opened 2018-05-30
Use PPI strategy for jbpm-container EJB tests

MarianMacik:
\+ minor README update.

Should I open another one for 7.7.x? We can merge it when we are able to.


mswiderski:
@MarianMacik since this is only test related go ahead and prepare 7.7.x PR too

PR #1236, opened 2018-05-31
[RHBPMS-5156] Exisiting deployments from 6.4.2 are always deactivated after server restart in 6.4.6+

elguardian:
active flag in the deployment unit descriptor is false by default overruling the
state in the database, causing the deployment not to start upon restart.


mswiderski:
Ok to test

mswiderski:
Jenkins retest this

PR #1237, opened 2018-06-01
[JBPM-4947] Add userInfo() for RuntimeEnvironmentBuilder to be able to set a custom UserInfo

elguardian:
Modified runtime environment to pass the information from the builder to the human task.

### DIFF
--- a/jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/RuntimeTimeTaskConfiguratorTest.java
+++ b/jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/RuntimeTimeTaskConfiguratorTest.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.runtime.manager.impl;
+
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.Persistence;
+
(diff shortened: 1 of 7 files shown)


elguardian:
depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/312

mswiderski:
Would be good to add a test case to make sure it works as expected

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/RuntimeTimeTaskConfiguratorTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+        EntityManagerFactory emf = Persistence.createEntityManagerFactory("org.jbpm.persistence.jpa");
+        RuntimeEnvironment environment = RuntimeEnvironmentBuilder.Factory.get()
+                 .newDefaultBuilder()
+                 .entityManagerFactory(emf)
+                 .get();
+
+        LocalTaskServiceFactory factory = new LocalTaskServiceFactory(environment);
+        InternalTaskService service = (InternalTaskService) factory.newTaskService();
+        // the instance provided shoulw be the one setup at custom and not the default one
```
MarianMacik: Just a small typo shoulw -> should.

elguardian:
@MarianMacik  done :)

elguardian:
ok to test

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/RuntimeTimeTaskConfiguratorTest.java
```diff
@@ -0,0 +1,101 @@
...
+    @BeforeClass
+    public static void enter() {
+        oldUserInfoValue = System.setProperty("org.jbpm.ht.userinfo", "custom");
+        oldUserInfoImpl = System.setProperty("org.jbpm.ht.custom.userinfo", "org.jbpm.runtime.manager.impl.CustomUserInfoImpl");
+    }
+
+    @AfterClass
+    public static void exit() {
+        if(oldUserInfoValue != null) {
+            System.setProperty("org.jbpm.ht.userinfo", oldUserInfoValue);
```
mswiderski: you should most likely clear system property even if it was not set otherwise all other tests will use your custom impl

elguardian: using System.clearProperty; nice catch.

PR #1240, opened 2018-06-05
RHPAM-131: Processes & Task Dashboard don't show data from undeployed…

nmirasch:
… Kie containers

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
@@ -18,6 +18,8 @@ package org.jbpm.kie.services.impl.query;
 
 import static org.jbpm.services.api.query.QueryResultMapper.COLUMN_DEPLOYMENTID;
 import static org.jbpm.services.api.query.QueryResultMapper.COLUMN_EXTERNALID;
+import static org.jbpm.services.api.query.QueryResultMapper.COLUMN_PROCESSINSTANCEID;
+import static org.jbpm.services.api.query.QueryResultMapper.COLUMN_TASKID;
 
 import java.util.ArrayList;
 import java.util.Collection;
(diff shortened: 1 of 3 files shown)


nmirasch:
@cristianonicolai @mswiderski @bsremac Here I have created this PR to be able to discuss what is the best approach to fix the problem. The point is when there are not allowed containers, the filter (IN) condition is not applied and the generated query returns all the values without taking that field in account. This code adds an impossible condition in that case, to the lookup to return an empty result. Could you take a look? any suggestion is more than welcome. Thanks!

mswiderski:
@nmirasch I'd ask first why do we filter dashboard data by deployments? this is not really a case for dashboard charts... So I would actually opt at removing this check for dashboard completely - maybe by including new target of the queries used by those charts.

nmirasch:
@mswiderski Looking at the implementation this preprocessor is added to show only items related with the user available deployments. Maybe for the dashboards make sense to show all the kie-server items removing this check , but my concern is this DeploymentIdsPreprocessor is used in other queries that potentially would have the same issue with the IN filter target with no values like FILTERED_BA_TASK, and FILTERED_PO_TASK.
Then I think we have two points here: 
1. Be sure if dashboards have to show items from stopped containers.
2. Find the way to make this preprocessor return empty results under specific situation.(the proposed fix for this is to include that impossible condition regarding the dataset doesn't have any possible configuration to not to execute the query and return empty
Does this make sense?

mswiderski:
> Be sure if dashboards have to show items from stopped containers.

for dashboards we should not filter it at all

> Find the way to make this preprocessor return empty results under specific situation.(the proposed fix for this is to include that impossible condition regarding the dataset doesn't have any possible configuration to not to execute the query and return empty

I am not sure we need deployment id filter for tasks - tasks should already be protected by user/group assignments isn't it?

cristianonicolai:
+1 for showing all the data in the dashboards. @nmirasch  I guess that means we would need to check if the container is running or not in order to show the details in the dashboard, plus remove the query target for these data set queries ( process and task dashboard ).

As for tasks and processes where the deployment pre-processor is used, that also helps to prevent to list data from containers that are not running which is something we still want to avoid.

nmirasch:
Maybe could be a good idea create another jira including  the dashboard different query without the contailerId filter addition but right now, I think we have to fix this preprocessor regarding it is being used by other queries at least for 7.0.1. to keep consistency. The current behaviour is that filter is applied if there is at least one started container, otherwise it show all the kieserver items and throw an ugly exception when the user try to access the details because the container is stopped.  

With that new jira (maybe for 7.0.2) we can work on see if that preprocessor is really needed and see the impact on the UI, accessing to details of items with its container stopped, giving access to items of all containers even the user could not have the right role to access to them, things like that...

@mswiderski  @cristianonicolai Do you think this is a good way to follow?

cristianonicolai:
jenkins please retest this

cristianonicolai:
@nmirasch +1 from me. @mswiderski  although we agree that we should list all data for the dashboard, that has implications on how the UI behaves in terms of handling data from stopped containers as well as security concerns as we present the task and process details. @nmirasch 's suggestion is to firstly contain the problem and open an enhancement to properly deal with these consequences.

mswiderski:
@cristianonicolai fine with me, if you see this as better approach go ahead

nmirasch:
@mswiderski  @cristianonicolai The dashboards jira issue has been fixed on jbpm-wb, and they are not using this preprocessor any more. Now, we have to decide what to do with the rest of queries that are using this preprocessor that is not working consistently (in my opinion) when all the containers are stopped. Do you think we have to open a jira with this, or we can consider it works as it's expected?
Thanks in advanced!

cristianonicolai:
jenkins please retest this

nmirasch:
jenkins please retest this

cristianonicolai:
jenkins please retest this

PR #1244, opened 2018-06-06
SimpleRuntimeEnvironmentTest shouldn't extend SimpleRuntimeEnvironment

MarianMacik:
Hi, @mswiderski,

I came through this test class which extended the implementation class because of the access to protected field. I switched it for Mockito injection as it properly should be.


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1246, opened 2018-06-08
RHPAM-1192: Error in the tasks search by potential owners

mcivantos-tribalyte:
@mswiderski  We've detected some incoherence in the tasks search by pot-owner. When the group is set:
- With status=null: the groupIds set in the input parameters isn't taken into account.
- With status not null: the groupIds set in the input parameters is taken into account.

And in all the queries used in this method (for example: https://github.com/kiegroup/jbpm/blob/4b5895028198c2ba6d3abf06ce2bd8c1a05d3c33/jbpm-human-task/jbpm-human-task-jpa/src/main/resources/META-INF/Taskorm.xml#L225) the where clause takes into account as pot-owner the userId OR the groups, but when the group is set in the input parameters the userId's groups aren't taken into account.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
@@ -836,7 +836,7 @@ public class RuntimeDataServiceImpl implements RuntimeDataService, DeploymentEve
 	public List<TaskSummary> getTasksAssignedAsPotentialOwner(String userId, List<String> groupIds, QueryFilter filter) {
 	    Map<String, Object> params = new HashMap<String, Object>();
         params.put("userId", userId);        
-        params.put("groupIds", getCallbackUserRoles(userGroupCallback, userId));
+        params.put("groupIds", mergeLists(groupIds, getCallbackUserRoles(userGroupCallback, userId)));
         
         applyQueryContext(params, filter);
         applyQueryFilter(params, filter);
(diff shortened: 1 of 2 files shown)


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -1337,12 +1337,26 @@ protected void applyQueryFilter(Map<String, Object> params, QueryFilter queryFil
...
-             
              return data;
          }
+         
          return source;
      }
      
+     protected List<?> mergeLists(List<?> source, List<?> values) {
+         List<Object> data = new ArrayList<Object>();            
+         for (Object value : values) {
```
mswiderski: why not using addAll?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -1337,12 +1337,26 @@ protected void applyQueryFilter(Map<String, Object> params, QueryFilter queryFil
...
+             data.add(value);
+         }
+         
+         if (source != null ) {
+             for (Object value : source) {
```
mswiderski: same here, why not use addAll?

mcivantos-tribalyte: Done

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
mcivantos-tribalyte: Done

mswiderski:
ok to test

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1250, opened 2018-06-11
[RHBPMS-4140] Automated Task reassignment is not working

elguardian:
Add a recalculatioin during claim operation. so it reschedules again the ExecuteDeadlinesCommand

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/command/CommandBasedTaskService.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/command/CommandBasedTaskService.java
@@ -165,7 +165,10 @@ public class CommandBasedTaskService implements InternalTaskService, EventServic
 	}
 
 	public void claim(long taskId, String userId) {
-		executor.execute(new ClaimTaskCommand(taskId, userId));
+        executor.execute(new CompositeCommand<Void>(
+             new ClaimTaskCommand(taskId, userId), // this is executed last
+             new CancelDeadlineCommand(taskId, true, true)
+             ));
 	}
 
 	// TODO: does not filter on language
--- a/jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/LocalTaskAuditWithDeadlineTest.java
+++ b/jbpm-human-task/jbpm-human-task-audit/src/test/java/org/jbpm/services/task/audit/service/LocalTaskAuditWithDeadlineTest.java
@@ -94,7 +94,10 @@ public class LocalTaskAuditWithDeadlineTest extends HumanTaskServicesBaseTest {
     
         Reader reader = new InputStreamReader(getClass().getResourceAsStream(MvelFilePath.DeadlineWithReassignment));
         Task task = (InternalTask) TaskFactory.evalTask(reader, vars);
-        taskService.addTask(task, new HashMap<String, Object>());
+
+        Map<String, Object> inputVars = new HashMap<String, Object>();
+        inputVars.put("NotCompletedReassign", "[users:Tony Stark,Bobba Fet,Jabba Hutt|groups:]@[500ms]");
+        taskService.addTask(task, inputVars);
         long taskId = task.getId();
         
         taskService.claim(taskId, "Tony Stark");
@@ -117,7 +120,7 @@ public class LocalTaskAuditWithDeadlineTest extends HumanTaskServicesBaseTest {
     
         // should have re-assigned by now
         countDownListener.waitTillCompleted();
-        
+
         task = taskService.getTaskById(taskId);
         assertNull(task.getTaskData().getActualOwner());
         assertEquals(Status.Ready, task.getTaskData().getStatus());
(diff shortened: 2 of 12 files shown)


mswiderski:
@elguardian could you move this PR to master? I don't think there is any support issue for it and thus should not target any other branch than master

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/command/CommandBasedTaskService.java
```diff
@@ -165,7 +165,10 @@ public void activate(long taskId, String userId) {
 	}
 
 	public void claim(long taskId, String userId) {
-		executor.execute(new ClaimTaskCommand(taskId, userId));
+        executor.execute(new CompositeCommand<Void>(
+             new ClaimTaskCommand(taskId, userId), // this is executed last
+             new CancelDeadlineCommand(taskId, true, true)
```
mswiderski: why claim should cancel any deadlines?

elguardian: addTask schedule the deadlines... so it could happen that if you add a task and you don't cancel you will have two different triggers.

mswiderski: I don't get this one... this PR should provide reschedule of the deadlines if they are defined and task was released so why claim needs to do anything?

elguardian: Deadlines are canceled and deleted once the ExecuteDeadlines or CancelDeadlineCommand happens.

> Set the reassignment property on a human task
>
> [users:|groups:manager, user]@[1m]@not-completed
> Start process and go to task list.
> 
> Claim the task, wait for the timer to expire.
> Refresh the task list.
> 
> Reclaim the task.

Deadlines are calculated based on relative time (AFAIK) @500ms or @1h ...etc...
So if you want to trigger those again you have to calculate deadlines again.
The Jira problem was that if you do 
addTask
Claim
<- deadline expired so it add again the potential owners and then the task can be claimed again.
Claim
<- no more deadlines because they are not calculated. -> so it never expired

That is the original problem. I thought that the way to solve this, was to restart the deadline once is claimed based on the relative time provided. 
addTask
Claim
<- cancel old deadlines and calculate new ones
<- schedule the ExecuteDeadlinesCommand

Reclaim
<- cancel old deadlines and calculate new ones
<- schedule the ExecuteDeadlinesCommand
.... and so on.

Point me out what would be the proper approach. If I understand that deadlies should be calculated upon realease ?

mswiderski: I'd actually put this in two places:
- in ReleaseTaskCommand
- in ExecuteDeadlinesCommand in section of handling reassignment 

so in these two places the deadlines should be rescheduled as only these operations should reset the counter so to say. Wdyt?

elguardian: I would go to a more general question:
1) When should we start the clock regarding the assignment ? ; right now it starts during addTask independently if it is claimed or not.

I see two things:
if the task is not claimed at any point you will be executed the same deadlines over and over again. I don't know if it makes sense.

Start task
deadline 1 -> reschedule
deadline 2 -> reschedule
deadline ...

(never claimed)

My guess is that there is no need to ReleaseTaskCommand to have that logic as execute deadlines will chain the next scheduled command.

Start task
claimed
claim
release (cancel old deline and reset a new one)

In this case the deadline would be never executed (you can enter a loop). It would be easy for the user to trick the system.

I guess it makes sense to only reschedule during ExecuteDeadlinesCommand

mswiderski: the clock is ticking as soon as task is created - regardless if it was claimed or not.

deadlines - which could be reassignment or notification should not be executed more than once unless task was released/reassigned which would then sort of reset the clock. Yes, it's true that someone could use this to reset the timer but that is still visible in the audit trail of the task.

Keep in mind that execute deadlines do not have to happen. Moreover, deadlines are of two types:
- not started - meaning task was not started - was not moved into InProgress state
- not completed - meaning task was not completed - was not moved to Completed state

So with that said deadlines should be scheduled when:
- task is created - initial state
- task is released - but only if the tasks was in InProgress state and this should only schedule not started deadlines because not completed is still there
- task is reassigned - since the deadline has expired and task was assigned to another user a new deadline should be set 

the last bullet is questionable in general as if the same rule should be applied to reassigned task or not. Don't have strong preference to be honest as both are valid to me.

deadlines should be canceled only if:
- task is started - this will cancel not started deadlines
- task is completed - this will cancel both - not started and not completed deadlines

does this make sense?

mswiderski:
closing as this should not go to 6.5.x branch

elguardian:
ok I will retake this tomorrow.

PR #1257, opened 2018-06-19
Stabilize EventEmitter tests

MarianMacik:
@mswiderski I have found the testIntegrationWithEventManager3 test failing in case MapBasedPersistenceTest is run before it. I think the issue was that these tests introduce their own Transaction Manager [here](https://github.com/kiegroup/jbpm/blob/master/jbpm-persistence/jbpm-persistence-jpa/src/test/java/org/jbpm/persistence/map/impl/MapBasedPersistenceTest.java#L53) which is actually ManualProcessTransactionManager. In such a case, MapBased tests registered their own TX manager and when TransactionalPersistenceEventManager creates its tx manager [here](https://github.com/kiegroup/jbpm/blob/master/jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/base/TransactionalPersistenceEventManager.java#L48) it does not find all thing it needs via JNDI, so then tx sync is not synchronized since everything is null [there](https://github.com/kiegroup/drools/blob/master/drools-persistence/drools-persistence-api/src/main/java/org/drools/persistence/jta/JtaTransactionManager.java#L303). Therefore, no events are persisted. Now when it is the other way around, during MapBased tests we get no tx active to register sync, since it registers sync on different tx manager than the manual one used by tests, but that's not an issue since we don't need it there.

So I have created a TestTransactionalPersistenceEventManager which is an extension of an ordinary one which we can reset to recreate the right jta tx manager.

Moreover, I slightly altered one compensation test method since on some DBs it returns wrong values if they are within the same timestamp. 

I hope it is understandable and my findings are correct. But I can say that it works on our Jenkins after this change.

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/CompensationTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/CompensationTest.java
@@ -16,17 +16,22 @@
 
 package org.jbpm.bpmn2;
 
+import static org.junit.Assert.assertThat;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.assertj.core.api.Assertions;
 import org.hamcrest.core.AnyOf;
 import org.hamcrest.core.Is;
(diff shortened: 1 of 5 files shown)


MarianMacik:
jenkins retest this

PR #1261, opened 2018-06-26
RHPAM-1257 Process Definitions: Project and Deployment Name coupled

bxf12315:
@bsremac 
Please check it.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
@@ -264,7 +264,7 @@ public class RuntimeDataServiceImpl implements RuntimeDataService, DeploymentEve
 						return o1.getName().compareTo(o2.getName());
 					} else if ("ProcessVersion".equals(queryContext.getOrderBy())) {
 						return o1.getVersion().compareTo(o2.getVersion());
-					} else if ("Project".equals(queryContext.getOrderBy())) {
+					} else if ("DeploymentId".equals(queryContext.getOrderBy())) {
 						return o1.getDeploymentId().compareTo(o2.getDeploymentId());
 					}
 					return 0;


bsremac:
@bxf12315 Could you attach _kie-server.war_ off of this PR, if you already have it?

bxf12315:
@bsremac Np, But I update repo of droolsjbpm-intergration yesterday, so I need update drools and jbpm and rebuild it, I will try to rebulid in the afternoon.

bsremac:
@bxf12315 Have tested it. Seems to be working fine now.
I think we are good.

MarianMacik:
Jenkins retest this

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -264,7 +264,7 @@ public int compare(ProcessDefinition o1, ProcessDefinition o2) {
 						return o1.getName().compareTo(o2.getName());
 					} else if ("ProcessVersion".equals(queryContext.getOrderBy())) {
 						return o1.getVersion().compareTo(o2.getVersion());
-					} else if ("Project".equals(queryContext.getOrderBy())) {
+					} else if ("DeploymentId".equals(queryContext.getOrderBy())) {
```
cristianonicolai: I'm concerned this could be a breaking change, we could perhaps keep both strings (Project and DeploymentId) instead. @mswiderski wdyt?

MarianMacik:
Actually, I have now found that the same "Project" is [here](https://github.com/kiegroup/jbpm/blob/master/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/CaseDefinitionComparator.java#L47) and [here](https://github.com/kiegroup/jbpm/blob/master/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/ProcessDefinitionComparator.java#L49). We should have it aligned I guess.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
MarianMacik: @cristianonicolai Well, it is not a fully breaking change in a way that it breaks your code, but yes, if somebody had used QueryContext's orderBy with "Project", it will no longer filter properly. I agree with keeping both.

bxf12315:
@MarianMacik @cristianonicolai  So , In here, We need keep orderBy variable "Project" and "DeploymentId", and keeping them sorted logic in here?

cristianonicolai:
@bxf12315 Is this change even necessary? Could we still use _Project_ as sort attribute from UI? If not I would say to keep both.

bxf12315:
@cristianonicolai It is not necessary, I think It just leads to confusion. 
@MarianMacik  I want to create test for a situation ,  sort by project(deployment), but I find all deployment of definitions is "org.jbpm.test:test-module:1.0.0".
https://github.com/kiegroup/jbpm/blob/master/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java#L287

Please tell me how to change it.

Thanks

MarianMacik:
@bxf12315 It's because here you deploy only one kjar https://github.com/kiegroup/jbpm/blob/c4eda77138c976016b301383db1369416b76029b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java#L78

To have another one you just need to deploy one more with different version for example.

As @bxf12315  said, if it can be fixed in a rather nice way I would fix it so it is not confusing.

cristianonicolai:
@bxf12315 sure, please do.

MarianMacik:
@cristianonicolai @bxf12315 So we are not doing this?

cristianonicolai:
@MarianMacik changes have been addressed in the UI, we continue to use _Project_ as the sort key, so no need to change the service.

bxf12315:
@MarianMacik Based on our previous discussion, we keep "Project" for this.
I think , we can create new issue to track refactor work.

MarianMacik:
OK, thanks.

PR #1253, opened 2018-06-13
[6.5.x] RHBPMS-5167 - Not possible to migrate process with recurring timer (#…

mswiderski:
…1234)

PR #1262, opened 2018-06-27
RHPAM-1122 - Follow-up: Not possible to migrate process with recurrin…

MarianMacik:
…g timer
@agiertli @mswiderski 

Here is the proposal on how to fix the issue. I found the very same approach in [drools](https://github.com/MarianMacik/drools/blob/36eccb8a33550eb2049255f058fe016c6a1fa2f7/drools-core/src/main/java/org/drools/core/time/impl/IntervalTimer.java#L136-L141).

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
@@ -623,10 +623,18 @@ public class MigrationManager {
     protected void updateBasedOnTrigger(TimerInstance timerInstance) {
         Trigger trigger = ((DefaultJobHandle)timerInstance.getJobHandle()).getTimerJobInstance().getTrigger();
         if (trigger instanceof IntervalTrigger && timerInstance.getPeriod() > 0) {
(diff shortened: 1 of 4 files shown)


MarianMacik:
Jenkins retest this

MarianMacik:
This should be also included in #1253.

MarianMacik:
I have added a reproducer test. And while adding it I have found another issue.
```timerInstance.setRepeatLimit(intervalTrigger.getRepeatLimit() - intervalTrigger.getRepeatCount()); ```
had to be changed to
```timerInstance.setRepeatLimit(intervalTrigger.getRepeatLimit() - intervalTrigger.getRepeatCount() + 1); ```

Since repeatCount also counts in last scheduled trigger which is cancelled before migration, we have to put it back.

The reason why the previously added test passed was that after timer no. 3 was scheduled, getRepeatLimit() - getRepeatCount() produced 4 - 3 = 1. And repeatLimit is controlled only in nextFireTime() method which is only called after the first invocation of the trigger. Where it correctly produced false as 2>=1. It is not checked in setFirstFireTime() method. So basically when somebody wants to repeat trigger n times, repeatLimit has to be set to n+1 :) (Because there is a repeatCount >= repeatLimit check.)

PR #1263, opened 2018-07-02
JBPM-7407: RESTWorkItemHandler should support custom HTTP Headers

jesuino:
JBPM-7407: RESTWorkItemHandler should support custom HTTP Headers


tsurdilo:
@jesuino could you fix the formatting of your changes please?

tsurdilo:
@jesuino looks ok to me (just the formatting comment above). One thing thats could simplify use for community is allow for easier ways to set up these custom headers to not have to create a 
Map<String, List<String>> httpHeaders in process. As you only give option to set custom headers via parameters. If user want to just have a single custom header name/value pair i think passing them in as simple strings (which can be done easily via data input editors) would be much much simpler imo.

jesuino:
Hello @tsurdilo I like the idea of passing headers using direct parameters in the WIH. What do you think about using a suffix for the header key, like Header-, so users can pass parameter for headers using: Header-token=abc would result in a header in the HTTP request  -> token: abc

tsurdilo:
@jesuino i think that may work. another idea could be fore example two string parameters:
HeaderNames,
HeaderValues

which can be comma-separated 
then users could specify simple strings in data inputs for the workitem for example:

HeaderNames=token,token2,token3
HeaderValues=abc,xyz,123

wdyt?

jesuino:
The problem is if users want to specify more than one parameter for a header, a header may have multiple parameters. token: abc, xyz

Another possible way is a parameter called Headers that would accept all headers separated by semi colon:

Headers="header1=val1,val2,val3;header2=val1,val2,val3"

Let me know which one you chose and I will modify the PR. Thanks for your comments :)

tsurdilo:
@jesuino OK sorry didn't know. I think this last one would be great. Thanks!

jesuino:
@tsurdilo I updated the PR, please me know if it looks fine to you.

Thanks!

tsurdilo:
@jesuino +1 thanks!

PR #1276, opened 2018-07-11
JBPM-7401: Show all process instances when all containers are stopped

nmirasch:
Introduce an impossible condition when all containers are stopped to avoid show all elements

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/DeploymentIdsPreprocessor.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/DeploymentIdsPreprocessor.java
@@ -15,6 +15,7 @@
  */
 package org.jbpm.kie.services.impl.query.preprocessor;
 
+import static org.dashbuilder.dataset.filter.FilterFactory.equalsTo;
 import static org.dashbuilder.dataset.filter.FilterFactory.in;
 
 import java.util.List;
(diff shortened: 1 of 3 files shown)


nmirasch:
@cristianonicolai @mswiderski  Bringing again the fix for the problem of that deploimentIdPreprocessor that only applies the filter when there are at least one container started and otherwise don't add any container restriction there. Could you take a look? Thanks!

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/DeploymentIdsPreprocessor.java
```diff
@@ -49,7 +55,12 @@ public void preprocess(DataSetLookup lookup) {
         }
 
         final List<String> deploymentIds = deploymentRolesManager.getDeploymentsForUser(identityProvider);
-        final ColumnFilter columnFilter = in(columnId, deploymentIds);
+        ColumnFilter columnFilter;
+        if(deploymentIds == null  || deploymentIds.size()==0){
```
MarianMacik: rather use deploymentIds.isEmpty(), please. It can be much quicker to check.

MarianMacik:
@nmirasch Just to clarify, JIRA says "Show all process instances when all containers are stopped" but actually this fix won't show anything if there are no active deployments, right? Thanks.

nmirasch:
@MarianMacik the jira title was trying to explain the problem. But it's true, the time tense have to be past.  The explanation was at jira. The inconsistent behaviour is that the filter over containers is applied only when any of the container are available if not, before this code, the query returned all elements not having any container consideration.

PR #1277, opened 2018-07-12
RHPAM-1271 KIE database creation does not include indexes

calvinzhuca:
Signed-off-by: calvin zhu <calvinzhu@hotmail.com>

This PR is for 
https://issues.jboss.org/browse/RHPAM-1271

It use JPA 2.1 annotation to create the indexes (index list is from jbpm-installer/src/main/resources/db/ddl-scripts)


mswiderski:
ok to test

PR #1281, opened 2018-07-18
[DROOLS-2740] Move ExecutableCommand to public API

danielezonca:
@mariofusco 
Changed import to use public KIE API instead of the old Drools internal


kkufova:
jenkins please retest this

PR #1282, opened 2018-07-19
[7.5.x] RHPAM-1154 - Signal on boundary event causes an unwanted event listener message

elguardian:
the cancel method should avoid firing and after node

cherry picked from #1226


mswiderski:
Jenkins retest this

MarianMacik:
jenkins retest this

mswiderski:
@mareknovotny shall this be merged? does it target any patch release?

mareknovotny:
@mswiderski no, if we have this on master we are ok and done.

PR #1283, opened 2018-07-19
[6.5.x] RHBPMS-5173 - Signal on boundary event causes an unwanted event listener message

elguardian:
the cancel method should avoid firing and after node

cherry picked from #1226


mswiderski:
Jenkins retest this

mareknovotny:
Jenkins retest this

MarianMacik:
jenkins retest this

mswiderski:
same here @mareknovotny shall this be merged? does it target any patch release?

mareknovotny:
We haven't agreed on inclusion in JIRA, so this is now obsolete I think until someone stand up with the request to merge still. I will keep that decision on you @mswiderski, please follow up on JIRA.

mswiderski:
@mareknovotny I am closing this PR as there is no response in jira on what to do with it, it can always be reopen so @elguardian please don't delete your branch

PR #1287, opened 2018-07-24
JBPM-7469 - Improve CaseData expression readability (auto import)

mswiderski:


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

jbpm-bpmn2/src/test/resources/BPMN2-SimpleXMLProcess.bpmn2
```diff
@@ -12,7 +12,10 @@
              xmlns:tns="http://www.jboss.org/drools">
 
   <process processType="Private" isExecutable="true" id="Minimal" name="&lt; &amp; &gt;" tns:packageName="com.sample" >
-
+    <extensionElements>
+     <tns:import name="org.kie.api.runtime.process.CaseAssignment" />
+     <tns:import name="org.kie.api.runtime.process.CaseData" />
+    </extensionElements>
```
MarianMacik: Was this addition needed? I mean this process definition is used only in BPMN2XMLTest.java and I cannot see any reason why this addition is needed. Or does it just test a use case with custom imports?

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ProcessImpl.java
```diff
@@ -49,7 +51,7 @@
     private ContextContainer contextContainer = new ContextContainerImpl();
     private Map<String, Object> metaData = new HashMap<String, Object>();
     private transient Map<String, Object> runtimeMetaData = new HashMap<String, Object>();
-    private Set<String> imports;
+    private Set<String> imports = new HashSet<>(Arrays.asList("org.kie.api.runtime.process.CaseData", "org.kie.api.runtime.process.CaseAssignment"));
```
MarianMacik: What if I add other custom imports to my process definition? It seems that it will override these automatic imports too, won't it?

jbpm-bpmn2/src/test/resources/BPMN2-SimpleXMLProcess.bpmn2
mswiderski: It compares two bpmn by xml and since this pr adds two imports automatically they will show up in dump of the process

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ProcessImpl.java
MarianMacik: Hmm, the only way I can change imports field is by setImports() method which will override this initialization. But maybe it cannot happen, just wanted to discuss it.

MarianMacik: Thanks, now I see that there is `list.add(name);` on a list returned by getter.

PR #1290, opened 2018-07-27
JBPM-7520 - Start conditional event do not start process instance eve…

mswiderski:
…n though activation was triggered

@hasys here is a fix for the start conditional event - it turned out to be a bug ... no wonder it was not found before ... test case for it had no assertions :(

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
@@ -413,23 +413,7 @@ public class ProcessRuntimeImpl implements InternalProcessRuntime {
 	            }
 	        	}
 	        }
-	        RuntimeManager manager = (RuntimeManager) kruntime.getEnvironment().get(EnvironmentName.RUNTIME_MANAGER);
-            if (manager != null) {
-                org.kie.api.runtime.manager.Context<?> context = ProcessInstanceIdContext.get();
-                
-                String caseId = (String) kruntime.getEnvironment().get(EnvironmentName.CASE_ID);
-                if (caseId != null) {
-                    context = CaseContext.get(caseId);
-                }
-                
-                RuntimeEngine runtime = manager.getRuntimeEngine(context);
-                KieSession ksession = runtime.getKieSession();
-                
-                ksession.execute(new StartProcessWithTypeCommand(processId, params, type));
-                
-            } else {
-            	startProcess( processId, params, type );
-            }
+	        startProcessWithParamsAndTrigger(processId, params, type, false);
 	    }
 	}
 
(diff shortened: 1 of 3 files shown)


hasys:

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -457,6 +460,30 @@ public void matchCreated(MatchCreatedEvent event) {
...
+                            RuntimeEngine runtime = manager.getRuntimeEngine(context);
+                            KieSession ksession = runtime.getKieSession();
+                            try {
+                                ksession.execute(new StartProcessWithTypeCommand(processId, null, "conditional"));
+                            } finally {
+                                manager.disposeRuntimeEngine(runtime);
+                            }
+                        } else {
+                            startProcess( processId, null, "conditional");
+                        }
```
MarianMacik: Looks good but I have just noticed that this part of code (starting with `if (manager != null)`) is exactly the same as this part here:
https://github.com/kiegroup/jbpm/blob/4afa414e0159e5727bccb358323905e363ef7136/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java#L417-L435
So maybe it might be good to export it to method.

mswiderski: good point, extracted to new method

MarianMacik: BTW why does it have to be done separately (in case there is / is not a runtime manager available) when in the end it calls the same method? We just want to be sure that we call it via right runtime manager in case of PPI strategy for example? Thanks.

mswiderski: th reason is to use the right context - like to not use the same ksession for new process instance but start a new one - mainly for strategies different than singleton

MarianMacik: Hmm. So what if it is a Singleton strategy - do we want to create PPIContext anyway?

mswiderski: that does not matter as singleton does not care about the context, it has always loaded single instance of runtime engine

hasys:

mswiderski:
Jenkins retest this

hasys:
Works good now, thank you @mswiderski!

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1291, opened 2018-08-02
JBPM-7541 - Error while sending email notifications for users defined…

mswiderski:
… in properties file


mswiderski:
Jenkins retest this

PR #1292, opened 2018-08-02
[6.5.x] RHBPMS-5204 - Cannot send email notifications for users defined in pr…

mswiderski:
…operties file

depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/316

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
@@ -65,7 +65,7 @@ public class EmailNotificationListener implements NotificationListener {
     
     @Override
     public void onNotification(NotificationEvent event, UserInfo userInfo) {
-        
+        logger.debug("User info implementation {} and mail session {}", userInfo, mailSession);
         if (userInfo == null || mailSession == null) {
             logger.info("Missing mail session or userinfo - skipping email notification listener processing");
             return;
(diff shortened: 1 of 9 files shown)


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
```diff
@@ -114,11 +114,12 @@ public void onNotification(NotificationEvent event, UserInfo userInfo) {
                     for (User user : entry.getValue()) {
     
                         String emailAddress = userInfo.getEmailForEntity(user);
-                        if (emailAddress != null && !toAddresses.contains(emailAddress)) {                        	
-                        	msg.addRecipients( Message.RecipientType.TO, InternetAddress.parse( emailAddress, false));
-                        	toAddresses.add(emailAddress);
+                        if (emailAddress != null) {                        	
+                        	if (toAddresses.add(emailAddress)) {
```
MarianMacik: I think this can be combined into ```if (emailAddress != null && toAddresses.add(emailAddress))```

jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/util/HumanTaskHandlerHelper.java
```diff
@@ -145,7 +145,7 @@ public static Deadlines setDeadlines(WorkItem workItem, List<OrganizationalEntit
 			EmailNotification emailNotification = TaskModelProvider.getFactory().newEmialNotification();
 			notifications.add(emailNotification);
 
-			emailNotification.setBusinessAdministrators(businessAdministrators);
+			emailNotification.setBusinessAdministrators(new ArrayList<>(businessAdministrators));
```
MarianMacik: Just curious if the collection is modified anywhere so we have to do a copy of it...

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
mswiderski: that was the whole point to split it to avoid warning saying that email was not found in case the set already has the email address

jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/util/HumanTaskHandlerHelper.java
mswiderski: this is required so hibernate will not fail saying there is a shared collection used for several entities.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/email/EmailNotificationListener.java
MarianMacik: Sorry, I didn't notice the else part :)

PR #1293, opened 2018-08-02
JBPM-7549 - Allow the use of stage-name to trigger adhoc fragment

mswiderski:
@DuncanDoyle if you find some time have a look

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/TriggerAdHocNodeInStageCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/TriggerAdHocNodeInStageCommand.java
@@ -17,7 +17,10 @@
 package org.jbpm.casemgmt.impl.command;
 
 import org.drools.core.command.impl.RegistryContext;
+import org.jbpm.casemgmt.api.AdHocFragmentNotFoundException;
 import org.jbpm.casemgmt.api.StageNotFoundException;
+import org.jbpm.casemgmt.api.model.CaseDefinition;
+import org.jbpm.casemgmt.api.model.CaseStage;
 import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
 import org.jbpm.workflow.instance.node.DynamicNodeInstance;
 import org.kie.api.runtime.Context;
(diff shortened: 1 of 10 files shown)


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/TriggerAdHocNodeInStageCommand.java
```diff
@@ -57,11 +62,19 @@ public Void execute(Context context) {
         ProcessInstance processInstance = ksession.getProcessInstance(processInstanceId);
         
         DynamicNodeInstance dynamicContext = (DynamicNodeInstance) ((WorkflowProcessInstanceImpl) processInstance).getNodeInstances(true).stream()
-                .filter(ni -> (ni instanceof DynamicNodeInstance) && stageId.equals(ni.getNode().getMetaData().get("UniqueId")))
+                .filter(ni -> (ni instanceof DynamicNodeInstance) && stage.equals(ni.getNode().getMetaData().get("UniqueId")) || stage.equals(ni.getNodeName()))
```
MarianMacik: Probably missing parentheses? What happens if NodeInstance is not a dynamic one and its nodeName is the one I am looking for? I guess there will be a ClassCastException.
Shouldn't there be
```(ni instanceof DynamicNodeInstance) && (stage.equals(ni.getNode().getMetaData().get("UniqueId")) || stage.equals(ni.getNodeName()))```

File: jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java [hunk 2]
```diff
@@ -3236,4 +3238,100 @@ public void testGetCaseFileInstanceForNotExistingCase() {
...
+        assertNotNull(caseId);
+        assertEquals(FIRST_CASE_ID, caseId);
+        try {
+            CaseInstance cInstance = caseService.getCaseInstance(caseId);
+            assertNotNull(cInstance);
+            assertEquals(deploymentUnit.getIdentifier(), cInstance.getDeploymentId());
+            
+            Map<String, Object> data = new HashMap<>();
+            data.put("type", "basic");
+            caseService.triggerAdHocFragment(caseId, "Stage #{type}", data);
```
MarianMacik: Just a question on how this work. So if we don't have NodeInstance created yet, we have to call it as if directly from case definition, i.e. without variables resolved. Once the instance of the stage is created, variables are resolved, right? Just thinking if it makes sense/is possible to do variable resolution even when triggering adHoc fragment, although I see that it can cause troubles since once it would be Stage basic, once Stage advanced, but it would still be the same Stage being invoked. So just an idea...

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/TriggerAdHocNodeInStageCommand.java
mswiderski: good catch, fixed

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
mswiderski: in general the stage needs to be created based on the definition - adhoc subprocess and if that has the name with expression it needs to be used as the way to created it but at the time it's created it will resolve the variable to make sure that stage instance has the correct name instead of expression.

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1294, opened 2018-08-03
[JBPM-7506] prevent manual start if timer start event is used

elguardian:


jbpm-flow/src/test/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceTest.java
```diff
@@ -0,0 +1,70 @@
...
+    private static String PROCESS_ID = "process.test";
+    static {
+        ProcessRuntimeFactory.setProcessRuntimeFactoryService(new ProcessRuntimeFactoryServiceImpl());
+    }
+
+    public void addLogger() { 
+        logger = LoggerFactory.getLogger(this.getClass());
+    }
+
+    @Test(expected=IllegalArgumentException.class)
```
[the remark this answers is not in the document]
triceo: Or AssertJ's `assertThatThrownBy()`. :-)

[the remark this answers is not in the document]
mswiderski: +1 on using AssertJ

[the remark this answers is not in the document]
MarianMacik: Yep, that's the ultimate and the best solution :)

elguardian:
done @MarianMacik

jbpm-flow/src/test/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstanceTest.java
```diff
@@ -0,0 +1,70 @@
...
+        ProcessRuntimeFactory.setProcessRuntimeFactoryService(new ProcessRuntimeFactoryServiceImpl());
+    }
+
+    public void addLogger() { 
+        logger = LoggerFactory.getLogger(this.getClass());
+    }
+
+    @Test
+    public void testStartProcessThrowException() throws Exception {
+    	Assertions.assertThatThrownBy(() -> {
```
MarianMacik: I really don't want to be nitpicking, but to mitigate false positives I would add there also ```.isInstanceOf(IllegalArgumentException.class)``` and possibly even ```.hasMessage("Runtime exception occurred")``` or ```.hasMessageContaining(...)```

elguardian:
done @MarianMacik 
Just quick note, I didn't add about hasMessage because it depends on the locale and I had problems in the past with other projects using that sort of test. (false positive when you are working in a different locale  -> my case es_ES)

mswiderski:
@elguardian this failure seems to be related
https://kie-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/jbpm-pullrequests/1939/testReport/junit/org.jbpm.test.regression.event/TimerEventTest/testStartTimerCycle/

elguardian:
@mswiderski the error seems legit.
Happens when the workflow is not dynamic but there is a timer  set. 
I set a new condition if there are not timers set.

elguardian:
ok to test

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
```diff
@@ -34,9 +34,11 @@ public RuleFlowProcess getRuleFlowProcess() {
 
     public void internalStart(String trigger) {
     	StartNode startNode = getRuleFlowProcess().getStart(trigger);
-    	if (startNode != null) {
-    		((NodeInstance) getNodeInstance(startNode)).trigger(null, null);
-    	}
+        if (startNode != null) {
+            ((NodeInstance) getNodeInstance(startNode)).trigger(null, null);
+        } else if (!getRuleFlowProcess().isDynamic() && this.getRuleFlowProcess().getTimerStart().isEmpty()) {
+            throw new IllegalArgumentException("There is no start node that matches the trigger " + trigger == null ? "none" : trigger);
```
MarianMacik: I just realized when I checked the failing test that there are parentheses missing around the ternary operator, string concatenation has higher priority.
So it should be:
`IllegalArgumentException("There is no start node that matches the trigger " + (trigger == null ? "none" : trigger));`

elguardian:
@MarianMacik  
nvm the last update is just wrong. I need to check better the test case. It is the actual problem we are trying to fix in here.

start time -> task -> end.

elguardian:
ok to test

elguardian:
ok to test

elguardian:
@mswiderski I think now it is ok.

jbpm-flow/pom.xml
```diff
@@ -162,6 +162,10 @@
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.assertj</groupId>
+      <artifactId>assertj-core</artifactId>
```
mswiderski: this should be in scope test

elguardian:
done @mswiderski

jbpm-flow/src/main/java/org/jbpm/ruleflow/instance/RuleFlowProcessInstance.java
```diff
@@ -34,9 +34,11 @@ public RuleFlowProcess getRuleFlowProcess() {
 
     public void internalStart(String trigger) {
     	StartNode startNode = getRuleFlowProcess().getStart(trigger);
-    	if (startNode != null) {
-    		((NodeInstance) getNodeInstance(startNode)).trigger(null, null);
-    	}
+        if (startNode != null) {
+            ((NodeInstance) getNodeInstance(startNode)).trigger(null, null);
+        } else if (!getRuleFlowProcess().isDynamic()) {
+            throw new IllegalArgumentException("There is no start node that matches the trigger " + trigger == null ? "none" : trigger);
```
MarianMacik: @elguardian There are still parentheses missing around the ternary operator, string concatenation has higher priority.
So it should be:
IllegalArgumentException("There is no start node that matches the trigger " + (trigger == null ? "none" : trigger));

In other words, 
`"There is no start node that matches the trigger " + trigger == null` is always false.

elguardian:
done @MarianMacik

mswiderski:
test failure not related

PR #1295, opened 2018-08-05
JBPM-7542: Create better response for user without permissions to do REST Forward task operation

mcivantos-tribalyte:
@mswiderski - Created better responses for user without permissions or tasks with status which doesn't match with the operation performed. The following task operations have been changed:

- Claim.
- Completed.
- Forward.
- Delegate.
- Release.
- Nomitate.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
@@ -187,9 +187,7 @@ public class UserTaskServiceImpl implements UserTaskService, VariablesAware {
 			TaskService taskService = engine.getTaskService();
 			// perform actual operation
 			taskService.claim(taskId, userId);
-		} catch(PermissionDeniedException e){
-            throw new TaskNotFoundException(e.getMessage());
-        } finally {
+		} finally {
 			disposeRuntimeEngine(manager, engine);
 		}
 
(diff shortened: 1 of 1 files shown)


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
```diff
@@ -188,6 +188,8 @@ public void claim(String deploymentId, Long taskId, String userId) {
 			// perform actual operation
 			taskService.claim(taskId, userId);
 		} catch(PermissionDeniedException e){
+		    throw new PermissionDeniedException(e.getMessage());
+        } catch(TaskNotFoundException e){
             throw new TaskNotFoundException(e.getMessage());
```
MarianMacik: The part
```
} catch(PermissionDeniedException e){
    throw new PermissionDeniedException(e.getMessage());
} catch(TaskNotFoundException e){
    throw new TaskNotFoundException(e.getMessage());
```
can be completely omitted I believe. Since you catch the same type of exception and then throwing it again with the same message, there is no reason to do it, Java will do it automatically. So I think we can delete these 4 lines completely and just leave try {...} finally {...} block.

mswiderski: the change itself is sort of needed as currently the PermissionDeniedException is caught and rethrown is TaskNotFound so both needs to be now handled separately ... but indeed, we might not need the catch at all...

mswiderski:
ok to test

mswiderski:
Jenkins retest this

PR #1296, opened 2018-08-06
JBPM-7383: Unable to access completed processes with undeployed conta…

nmirasch:
…iner. Removed DeploymentIdPreprocessor


nmirasch:
@mswiderski @cristianonicolai removed DeploymentIdPreprocessor, could you take a look? Thanks!

cristianonicolai:
jenkins please retest this

cristianonicolai:

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
```diff
@@ -177,14 +171,10 @@ public void replaceQuery(QueryDefinition queryDefinition) {
                     dataSetDefRegistry.registerPreprocessor(sqlDef.getUUID(), new BusinessAdminTasksPreprocessor(identityProvider, userGroupCallback, metadata));
                 } else if (queryDefinition.getTarget().equals(Target.PO_TASK)) {
                     dataSetDefRegistry.registerPreprocessor(sqlDef.getUUID(), new PotOwnerTasksPreprocessor(identityProvider, userGroupCallback, metadata));
-                } else if (queryDefinition.getTarget().equals(Target.FILTERED_PROCESS)) {
-                    dataSetDefRegistry.registerPreprocessor(sqlDef.getUUID(), new DeploymentIdsPreprocessor(deploymentRolesManager, identityProvider, COLUMN_EXTERNALID,COLUMN_PROCESSINSTANCEID));
```
MarianMacik: Just realized that `Target.FILTERED_PROCESS` can probably be deleted since I cannot see any other place where it is used in codebase.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/QueryServiceImplTest.java
```diff
@@ -1019,15 +1019,6 @@ public void testGetFilteredProcessInstances() {
...
 
-        // let's now change the roles so user should not see instances
-        roles.clear();
-        roles.add("employees");
-        identityProvider.setRoles(roles);
-        identityProvider.setName("anotherUser2");
-
-        instances = queryService.query(query.getName(), ProcessInstanceQueryMapper.get(), new QueryContext());
-        assertNotNull(instances);
-        assertEquals(0, instances.size());
```
MarianMacik: What about leaving this use case but just changing the number in the assert? So we are sure that anotherUser2 can see the instances as well?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/QueryServiceImpl.java
cristianonicolai: @MarianMacik @mswiderski it is used in the custom queries UI, see: https://github.com/kiegroup/jbpm-wb/blob/master/jbpm-wb-dashboard/jbpm-wb-dashboard-client/src/main/java/org/jbpm/dashboard/dataset/editor/impl/RemoteDataSetDefAttributesEditor.java#L95
Should it be removed from both places?

mswiderski: that's a good question @cristianonicolai - thinking from backwards compatibility point of view it might not be safe to remove the target as there might be queries using that - even our own will refer to that target. So maybe in the end it's better to keep the FILTERED_PROCESS target and the preprocessor as well? We are not going to use it but it might still be useful to some scenarios - like building dashboards based on user access to deployments... wdyt?

cristianonicolai: @mswiderski good point, for _Target.FILTERED_PROCESS_ I think is fine to keep as is as we now use a CUSTOM target anyway.

mswiderski: correct, but if we keep the target we should keep the preprocessor as well to preserve the behaviour

cristianonicolai: definitely, I'll create a new PR to reflect that.

cristianonicolai:
Create a new PR to handle this changes, see https://github.com/kiegroup/jbpm/pull/1300

PR #1297, opened 2018-08-07
[DO NOT MERGE][JBPM-7598] MigrationManager allows nodeInstance which …

tkobayas:
…cannot find upgradedNode

This PR is just a Unit test to explain the issue. If we let MigrationManager fail in this case, the unit test would change to assert the failure.


mswiderski:
incorporated in https://github.com/kiegroup/jbpm/pull/1299

PR #1301, opened 2018-08-14
JBPM-7646 - Broadcast signal with expression is not received by event…

mswiderski:
… subprocess

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
@@ -529,6 +529,9 @@ public abstract class WorkflowProcessInstanceImpl extends ProcessInstanceImpl
                 List<String> events = ((EventSubProcessNode) node).getEvents();
                 for (String type : events) {
                     addEventListener(type, EMPTY_EVENT_LISTENER, true);
+                    if (isVariableExpression(type)) {
+                        addEventListener(resolveVariable(type), EMPTY_EVENT_LISTENER, true);
+                    }
                 }
             }  else if (node instanceof DynamicNode) {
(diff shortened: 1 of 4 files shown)


mswiderski:
Jenkins retest this

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -529,6 +529,9 @@ private void registerExternalEventNodeListeners() {
                 List<String> events = ((EventSubProcessNode) node).getEvents();
                 for (String type : events) {
                     addEventListener(type, EMPTY_EVENT_LISTENER, true);
+                    if (isVariableExpression(type)) {
```
tsurdilo: if isVariableExpression we add two event listeneres - is that right thing to do or should do this check first?

mswiderski: yes, that's on purpose to allow variable expression to be resolved on demand - for instance when process instance is going to be persisted, duplicates will be removed as all events are kept in set.

tsurdilo: Ok. thanks for explaining

MarianMacik: Hmm, so we have one listener with the current value of variable and the second one with unresolved expression? When we update the variable, the old one get's removed or not?

mswiderski: these two are only registered for the time process instance is being executed so once it's done (tx is being committed) it will the store only one.

PR #1306, opened 2018-08-21
[RHPAM-153] parentProcessInstanceId in ProcessEventListener is 0

elguardian:
setting process instance parent to -1 by default

### DIFF
--- a/jbpm-flow/src/test/java/org/jbpm/event/process/ProcessEventSupportTest.java
+++ b/jbpm-flow/src/test/java/org/jbpm/event/process/ProcessEventSupportTest.java
@@ -21,6 +21,7 @@ import static org.junit.Assert.assertEquals;
 import java.util.ArrayList;
 import java.util.List;
 
+import org.apache.commons.lang3.mutable.MutableLong;
 import org.drools.core.definitions.InternalKnowledgePackage;
 import org.drools.core.definitions.impl.KnowledgePackageImpl;
 import org.drools.core.impl.InternalKnowledgeBase;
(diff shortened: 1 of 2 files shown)


jbpm-flow/src/test/java/org/jbpm/event/process/ProcessEventSupportTest.java
```diff
@@ -543,4 +542,88 @@ public void afterVariableChanged(ProcessVariableChangedEvent event) {
...
+        );
+
+        final InternalKnowledgePackage pkg = new KnowledgePackageImpl( "org.drools.test" );
+        pkg.addProcess(process);
+        List<KiePackage> pkgs = new ArrayList<KiePackage>();
+        pkgs.add( pkg );
+        kbase.addPackages( pkgs );
+
+        KieSession session = kbase.newKieSession();
+        final List<ProcessEvent> processEventList = new ArrayList<ProcessEvent>();
```
mswiderski: what's the purpose of this list? I don't see it being checked/asserted in anyway?

jbpm-flow/src/test/java/org/jbpm/event/process/ProcessEventSupportTest.java
```diff
@@ -543,4 +542,88 @@ public void afterVariableChanged(ProcessVariableChangedEvent event) {
...
+        final MutableLong parentProcessId = new MutableLong(0L);
+        final ProcessEventListener processEventListener = new ProcessEventListener() {
```
mswiderski: you can use DefaultProcessEventListener and override only method(s) that you need

elguardian:
just clean up those two things @mswiderski

PR #1307, opened 2018-08-21
DROOLS-1393 - Investigate why JpaOptLockPersistentStatefulSessionTest…

mswiderski:
… is failing with Hibernate 5

depends on https://github.com/kiegroup/drools/pull/2019

this change is needed as on per request strategy the session is destroyed and thus causes now additional error from narayana about failure in delisting connection from transaction.


mswiderski:
Jenkins retest this

PR #1310, opened 2018-08-22
JBPM-7671 - Verification, build and deploy doesn't notify user about …

tsurdilo:
…empty subprocess

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/validation/RuleFlowProcessValidator.java
+++ b/jbpm-flow/src/main/java/org/jbpm/ruleflow/core/validation/RuleFlowProcessValidator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2018 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


tsurdilo:
Sorry about lots of formatting changes. The additions are:

https://github.com/kiegroup/jbpm/pull/1310/files#diff-3f817daf9ffde3edd8cfc8fc3e9dd0b3R576

https://github.com/kiegroup/jbpm/pull/1310/files#diff-146fe70debee31f1424f7bf1299c17e8R174

hasys:

hasys:
Hi @tsurdilo, not sure, will test it with Full Downstream Build, but it looks like new check will test only not connected Subprocesses, like Event subprocess but Embedded subprocess mentioned in Steps to Reproduce will be still not validated.

tsurdilo:
@hasys your scenario mentioned should also be covered.

tsurdilo:
@hasys thanks for that info - yes lets create a separate issue so we can discuss it. Thanks!

tsurdilo:
@hasys looked into your findings some more:
* for multi-intsance - yes that is right behavior
* for ad-hoc - this is also right behavior. this is a dynamic node and should have the ability to start and then users can add nodes to it. this behavior of completing if empty is specifically removed actually: https://github.com/kiegroup/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java#L76
User do have option to cancel an empty ad-hoc subprocess via ProcessInstanceAdminService but that is not yet exposed in the workbench. So maybe if want a jira to open that would be something, to add that service/expose it.

MarianMacik:
@mswiderski sorry, it seems that I missed the notification, but looks good :)

hasys:
Hi @tsurdilo, thank you for great explanation. I reported related issue against  Workbench component there: https://issues.jboss.org/browse/JBPM-7697

PR #1311, opened 2018-08-23
replace JPAAuditService use in MigrationManager with direct EntityMan…

mswiderski:
…ager to avoid issues with spring setups

@MarianMacik here it is what we talked about

### DIFF
--- a/jbpm-human-task/jbpm-human-task-jpa/pom.xml
+++ b/jbpm-human-task/jbpm-human-task-jpa/pom.xml
@@ -66,11 +66,6 @@
     </dependency>
 
     <dependency>
-      <groupId>org.codehaus.jackson</groupId>
-      <artifactId>jackson-core-asl</artifactId>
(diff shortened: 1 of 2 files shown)


jbpm-human-task/jbpm-human-task-jpa/pom.xml
```diff
@@ -65,11 +65,6 @@
       <scope>provided</scope>
     </dependency>
 
-    <dependency>
-      <groupId>org.codehaus.jackson</groupId>
-      <artifactId>jackson-core-asl</artifactId>
-    </dependency>
```
mswiderski: @MarianMacik I added one more comment to remove old and not used Jackson dependency as it then is packaged into kieserver/spring boot apps

MarianMacik: Ok, I think we are ready to merge here.

PR #1313, opened 2018-08-23
JBPM-7556 Typed Process Variables

evacchi:
VariableInstances is a new method for accessing instance variables, adding typed access. Typed variable instances encapsulate validation for different kind of variables (e.g. Case Variables)

A test case is provided with example usages of the new API with the expected behavior

The change is quite self-contained. More details on the rationale in the JIRA.

The change does not affect the way serialization and lazy loading is handled at the boundary, but only the way processes are created internally (`createProcess()`)


evacchi:
Jenkins please retest this

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
```diff
@@ -0,0 +1,102 @@
...
+        public T get() {
+            Collection<CaseData> caseFiles = (Collection<CaseData>)
+                    knowledgeRuntime.getObjects(new ClassObjectFilter(CaseData.class));
+            if (caseFiles.size() == 1) {
+                CaseData caseFile = caseFiles.iterator().next();
+                // check if there is case file prefix and if so remove it before checking case file data
+                if (caseFile != null) {
+                    return (T) caseFile.getData(caseFileName);
+                }
+            }
```
mswiderski: Shouldn’t this throw an exception? Assuming we are sure the case file should be there it should fail in case these is no case file in working memory

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
```diff
@@ -0,0 +1,102 @@
...
+            Collection<CaseData> caseFiles = (Collection<CaseData>)
+                    knowledgeRuntime.getObjects(new ClassObjectFilter(CaseData.class));
+            if (caseFiles.size() == 1) {
+                CaseData caseFile = caseFiles.iterator().next();
+                FactHandle factHandle = knowledgeRuntime.getFactHandle(caseFile);
+                Objects.requireNonNull(factHandle, "factHandle for " + caseFileName + " was null");
+                caseFile.add(caseFileName, value);
+                knowledgeRuntime.update(factHandle, caseFile);
+                ((KieSession) knowledgeRuntime).fireAllRules();
+            }
```
mswiderski: Same here I guess

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+    };
+
+    public CaseVariableInstance(VariableScopeInstance parentScopeInstance, Variable variable) {
+        this.knowledgeRuntime = parentScopeInstance.getProcessInstance().getKnowledgeRuntime();
+        this.variable = variable;
+        if (variable.getName().startsWith(VariableScope.CASE_FILE_PREFIX)) {
+            this.caseFileName = variable.getName().substring(VariableScope.CASE_FILE_PREFIX.length());
+        } else {
+            // should probably issue a warning
```
mswiderski: What do you mean by warning?

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
```diff
@@ -0,0 +1,102 @@
...
+
+    @Override
+    public void setReference(ValueReference<T> value) {
+        // shall we log this? "Cannot setReference on a CaseVariableInstance". will set(value.get())
+        set(value.get());
+    }
+
+    @Override
+    public ValueReference<T> getReference() {
+        // fixme: is a CaseVariable really an InstanceVariable? Shall we use a ValueReference?
```
mswiderski: I’d say it should be InstanceVariable, the main difference is the scope of it - case variable is broader than process variable

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ProcessImpl.java
```diff
@@ -96,6 +97,17 @@ public void setPackageName(String packageName) {
 		this.packageName = packageName;
 	}
 
+    public String getRuleUnit() {
```
mswiderski: What is this used for?

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessVariables.java
```diff
@@ -0,0 +1,120 @@
...
+import org.kie.api.runtime.rule.RuleUnit;
+
+public abstract class ProcessVariables implements Serializable {
+
+    public static <T extends RuleUnit> ProcessVariables.Typed<T> typed(T value) {
+        return new Typed<>(value);
+    }
+
+    public static ProcessVariables.Untyped untyped(Map<String, Object> parameters) {
+        return new Untyped(parameters == null ? Collections.emptyMap() : parameters);
```
mswiderski: This can be source of issues as emptyMap is unmodifiable if I am not mistaken so might be better to put simply hashmap instead

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessVariables.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.stream.Collectors;
+
+import org.jbpm.process.core.context.variable.LambdaValueReference;
+import org.jbpm.process.core.context.variable.SimpleValueReference;
+import org.jbpm.process.core.context.variable.ValueReference;
+import org.kie.api.runtime.rule.RuleUnit;
+
+public abstract class ProcessVariables implements Serializable {
+
+    public static <T extends RuleUnit> ProcessVariables.Typed<T> typed(T value) {
```
mswiderski: Same question here, what rule unit is used for? In general maybe we should abstract it bit to have a unit instead rule unit as it might make more sense to refer to unit or process unit in jbpm

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -81,76 +87,177 @@ public Object getVariable(String name) {
...
+        }
 
         return null;
     }
 
     public Map<String, Object> getVariables() {
-        return Collections.unmodifiableMap(variables);
+        Map<String, Object> result = new HashMap<>();
+        variables.values().stream()
+                .filter(v -> !(v instanceof CaseVariableInstance))
```
mswiderski: Why filtering out case variables?

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -81,76 +87,177 @@ public Object getVariable(String name) {
...
+    /**
+     * Assign a value reference to a declared variable instance.
+     * @throws IllegalArgumentException if the given variableName has not been declared in the process
+     */
+    public VariableInstance<?> assignVariableInstance(String variableName, ValueReference reference) {
+        VariableScope variableScope = getVariableScope();
+        Variable variable = variableScope.findVariable(variableName);
+        if (variable == null) {
+            variable = new Variable(variableName);
+            variable.setMetaData("on-the-fly", true);
```
mswiderski: What is this on-the-fly used for?

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -81,76 +87,177 @@ public Object getVariable(String name) {
...
+                getProcessInstance().getProcessName(),
+                variableName,
+                reference.get());
+        VariableInstance<?> instance = addVariableInstance(variableName);
+        instance.setReference(reference);
+        return instance;
+    }
+
+    private String instancePrefixed(String name) {
+        return (variableInstanceIdPrefix == null ? "" : variableInstanceIdPrefix + ":") + name;
```
mswiderski: What’s this prefix?

evacchi:
Thanks a lot @mswiderski for feedback, will go through it later

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
evacchi: I think I agree with this, although I'm pretty sure we are currently defaulting to returning `null`. But I'm all for throwing.

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
evacchi: so currently, when a variable begins with `CASE_FILE_PREFIX` we remove the prefix, then look it up as a case variable. Put apparently there are situations (see the `else` branch) where the case variable has **no prefix** and yet we look it up.
https://github.com/kiegroup/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L52

The procedure in master goes like this:

- look for the given var name in the map of variables
- if no result: see if it's a "magic" name ("processInstanceId" or "parentProcessInstanceId")
- no result: see it's a global
- no result: see if it's a case file -- if there is a prefix, strip it, otherwise just use that name

so in the last case, we just lookup the given name in the case file... even if it's not prefixed. So, the question is: shall we raise a warning? e.g.: "Cannot find $name in scope, looking for $name in case files"

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
evacchi: this is rather a question on the class hierarchy/class relations here.

A CaseVariable here is a subclass of an InstanceVariable; but each InstanceVariable was defined to hold a `ValueReference`, which also has different types. I was unsure here whether `Case` should be a subclass of a  `ValueReference` or a subclass of `InstanceVariable`. It probably doesn't make much sense to leave that comment there, though.

mswiderski:
Jenkins retest this

evacchi:
Jenkins retest this

evacchi:
Apparently this test now breaks expecting 13 entries and it gets 31 (!!!)

https://github.com/evacchi/jbpm/blob/121be21c6291331ac369395be09eb307677633d4/jbpm-audit/src/test/java/org/jbpm/process/audit/AbstractAuditLogServiceTest.java#L457

I am not sure about what's happening here. The indexer seems not to de-duplicate correctly the entries (?) what do you think is wrong there?

evacchi:
Fixed the test above, now investigating more failures

mswiderski:

mswiderski:
executing full downstream to make sure it works on droolsjbpm-integration repo (spring, Karaf and kie server) and if all goes well I think it's good to be merged

@MarianMacik please have a look as well.

evacchi:
cool, I'll take a second look today to purge from some leftover (e.g. unused methods/classes)

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
evacchi: so this is currently not used anywhere, but I figured we'd better mark "undeclared" variables somehow (i.e. variables that are defined "on the fly"). This should (of course) be an internal implementation detail. Do we want to 

1) add a new field in Variable ?
2) use some specific meta data field instead of "on-the-fly" ?

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessVariables.java
evacchi: removed

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ProcessImpl.java
evacchi: removed

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
evacchi: what do you guys think about this? Should we log? Should we throw?

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
mswiderski: let's ignore I guess this is what we do now, isn't it?

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/CaseVariableInstance.java
mswiderski: let's log ... just to stay on safe side for now

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
mswiderski: I'd go for new field so it is more explicit and easier to follow

evacchi:

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessVariables.java
evacchi: good catch

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessVariables.java
```diff
@@ -0,0 +1,116 @@
...
+            try {
+                BeanInfo beanInfo = Introspector.getBeanInfo(value.getClass());
+                this.propertyDescriptors = new HashMap<>();
+                for (PropertyDescriptor propertyDescriptor : beanInfo.getPropertyDescriptors()) {
+                    propertyDescriptors.put(propertyDescriptor.getName(), propertyDescriptor);
+                }
+            } catch (IntrospectionException e) {
+                throw new IllegalArgumentException(e);
+            }
+        }
```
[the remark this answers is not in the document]
evacchi: the Typed version (not yet exposed in a public interface) will support an object instance as a container for variables. The Untyped version uses the usual map backing (this is currently the default)

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -81,76 +86,181 @@ public Object getVariable(String name) {
...
+        if (variable == null) {
+            variable = new Variable(variableName);
+            variable.setMetaData("on-the-fly", true);
+        }
+        variableScope.validateVariable(
+                getProcessInstance().getProcessName(),
+                variableName,
+                reference.get());
+        VariableInstance<?> instance = getVariableInstance(variableName);
+        instance.setReference(reference);
```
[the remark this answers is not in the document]
evacchi: yes exactly. This is all done internally; the API requiring a ProcessVariables instance instead of a Map<String,Object> is not not published to kie-api for now

evacchi:
I am momentarily closing this because I am focusing on something else, and it might not be the best design (maybe overkill for the task at hand)

PR #1315, opened 2018-08-27
[RHPAM-2] NullPointerException during transaction rollback when using CMT

elguardian:
integration test for RHPAM-2

### DIFF
--- a/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/AbstractEJBServicesTest.java
+++ b/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/AbstractEJBServicesTest.java
@@ -22,6 +22,7 @@ import java.util.List;
 import javax.ejb.EJB;
 
 import org.assertj.core.api.Assertions;
+import org.assertj.core.api.ThrowableAssert.ThrowingCallable;
 import org.jboss.arquillian.container.test.api.Deployment;
 import org.jboss.arquillian.container.test.api.TargetsContainer;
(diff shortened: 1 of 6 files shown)


jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/AbstractEJBServicesTest.java
```diff
@@ -95,9 +96,20 @@ public boolean accept(File dir, String name) {
             }
         }
     }
+ 
+    protected static Throwable catchRootCause(ThrowingCallable shouldRaiseThrowable) {
+        Throwable result = Assertions.catchThrowable(shouldRaiseThrowable);
+        Throwable cause = null; 
+
+        while(null != (cause = result.getCause())  && (result != cause) ) {
```
MarianMacik: Interesting, so down the stacktrace it can happen that `result == clause`?

elguardian: It shouldn't,  but it covers possible mistakes when somebody sets the cause of the exceptions to itself (that would cause an infinite loop). I haven't seen it here but it does not hurt

MarianMacik: @elguardian yeah, definitely

tsurdilo:
Hi @elguardian, can this be merged?

elguardian:
@tsurdilo the dependency is not yet merged.

PR #1321, opened 2018-09-06
JBPM-7721 - add toFile method to Document

tsurdilo:


jbpm-document/src/main/java/org/jbpm/document/service/impl/DocumentImpl.java
```diff
@@ -174,6 +178,16 @@ public void setContent(byte[] content) {
         return content;
     }
 
+    @Override
+    public File toFile() throws IOException {
+        File file = File.createTempFile(getName().split("\\.")[0], "." + getName().split("\\.")[1]);
```
mswiderski: would be best to do the split just once

jbpm-document/src/main/java/org/jbpm/document/service/impl/DocumentImpl.java
```diff
@@ -174,6 +178,16 @@ public void setContent(byte[] content) {
...
+        file.setLastModified(getLastModified().getTime());
+        BufferedWriter writer = new BufferedWriter(new FileWriter(file));
+        writer.write(new String(getContent(), "UTF-8"));
```
mswiderski: I don't think we can assume the document is always a string/text document. Instead with should always write bytes.

tsurdilo:
@mswiderski updated - thanks

PR #1322, opened 2018-09-10
[JBPM-7731] SLA tracking command signals ended process or nodes

elguardian:
Adding filter to not signal processes that are completed or aborted

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/SLATrackingCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/SLATrackingCommand.java
@@ -100,8 +100,9 @@ public class SLATrackingCommand implements Command, Reoccurring {
         parameters.put("now", new Date());
         StringBuilder lookupQuery = new StringBuilder();
 
-        lookupQuery.append("select log from NodeInstanceLog log where ");
-        lookupQuery.append("log.nodeInstanceId in ( select nil.nodeInstanceId from NodeInstanceLog nil where nil.slaDueDate < :now and nil.slaCompliance = 1 ");
+        lookupQuery.append("select log from NodeInstanceLog log, ProcessInstanceLog proc ");
+        lookupQuery.append("where proc.processInstanceId = log.processInstanceId and proc.status = 1 ");
+        lookupQuery.append("and log.nodeInstanceId in ( select nil.nodeInstanceId from NodeInstanceLog nil where nil.slaDueDate < :now and nil.slaCompliance = 1 ");
         lookupQuery.append("GROUP BY nil.nodeInstanceId ");
         lookupQuery.append("HAVING sum(nil.type) = 0) ");
         lookupQuery.append("and log.type = 0 ");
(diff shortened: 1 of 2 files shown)


jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/SLATrackingCommand.java
```diff
@@ -100,11 +100,11 @@ public ExecutionResults execute(CommandContext ctx) throws Exception {
...
         StringBuilder lookupQuery = new StringBuilder();
 
-        lookupQuery.append("select log from NodeInstanceLog log where ");
-        lookupQuery.append("log.nodeInstanceId in ( select nil.nodeInstanceId from NodeInstanceLog nil where nil.slaDueDate < :now and nil.slaCompliance = 1 ");
+        lookupQuery.append("select log from NodeInstanceLog log, ProcessInstanceLog proc ");
+        lookupQuery.append("where log.nodeInstanceId in ( select nil.nodeInstanceId from NodeInstanceLog nil where nil.slaDueDate < :now and nil.slaCompliance = 1 ");
         lookupQuery.append("GROUP BY nil.nodeInstanceId ");
         lookupQuery.append("HAVING sum(nil.type) = 0) ");
-        lookupQuery.append("and log.type = 0 ");
+        lookupQuery.append("and log.type = 0 and proc.status NOT IN(2,3) and proc.processInstanceId = log.processInstanceId ");
```
mswiderski: I believe better would be to use status = 1 as that is the only status that indicates active instance and I would move this join and condition part as the first ones, does not matter to look up node instance info if the process instance is not active

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/SLATrackingCommand.java
```diff
@@ -144,7 +144,7 @@ public ExecutionResults execute(CommandContext ctx) throws Exception {
         parameters.put("now", new Date());
         lookupQuery = new StringBuilder();
 
-        lookupQuery.append("select log from ProcessInstanceLog log where log.slaDueDate < :now and log.slaCompliance = 1 ");
+        lookupQuery.append("select log from ProcessInstanceLog log where log.slaDueDate < :now and log.slaCompliance = 1 and log.status NOT IN(2,3)");
```
mswiderski: same here, use status = 1 instead of NOT IN

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/SLATrackingCommandTest.java
```diff
@@ -250,6 +256,48 @@ public void testSLATrackingOnUserTaskSLAMet() throws Exception {
...
+        assertEquals(ProcessInstance.STATE_ACTIVE, processInstance.getState());
+        
+        List<TaskSummary> tasks = runtime.getTaskService().getTasksAssignedAsPotentialOwner("john", "en-UK");
+        assertEquals(1, tasks.size());
+        
+
+        Thread.sleep(3000);
+
+
+
```
mswiderski: maybe reduce the empty lines here...

elguardian:
@mswiderski  Just to double check. Just change the state to ACTIVE. 
possible states: STATE_PENDING, STATE_ACTIVE, STATE_COMPLETED, STATE_ABORTED, STATE_SUSPENDED
Is it ok to only signal the process the SLA is only running during ACTIVE state ? what about the individual timer logic ?

mswiderski:
@elguardian correct, just change to look up for ACTIVE - other states like pending or suspended are irrelevant - pending is only used before process instance is started and suspended is not implemented

elguardian:
@mswiderski any progress with the review... I performed the changes

mswiderski:
Jenkins retest this

PR #1323, opened 2018-09-10
[RHPAM-2] NullPointerException during transaction rollback when using CMT (#1315)

elguardian:
integration test for RHPAM-2


mswiderski:
this was not approved as far as I know for 7.1 so closing

PR #1326, opened 2018-09-12
JBPM-7639- Provide Test Helper Classes for Case Management and jBPM …

mswiderski:
…services


jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/ProcessServiceImplPerProcessInstanceTest.java
```diff
@@ -16,89 +16,58 @@
...
 import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
-import org.kie.scanner.KieMavenRepository;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import static org.junit.Assert.*;
-import static org.kie.scanner.KieMavenRepository.getKieMavenRepository;
 
-public class ProcessServiceImplPerProcessInstanceTest extends AbstractKieServicesBaseTest {
+public class ProcessServiceImplPerProcessInstanceTest extends AbstractKieServicesTest {
```
MarianMacik: Why not to extend `AbstractKieServicesBaseTest` and override desired methods? GAV is the same so we won't need to specify it here I think, since it is already defined in `*BaseTest` class.

mswiderski: the whole point here was to simulate how end user would make use of this AbstractKieServicesTest class so to make sure only these two (optionally three) methods needs to be implemented.

MarianMacik: Ah, OK. So AbstractKieServicesBaseTest is only "our" approach, but user is encouraged to use directly AbstractKieServicesTest in case he doesn't have so complex test hierarchy as we have?

mswiderski: yup

PR #1327, opened 2018-09-14
JBPM-7763 - jBPM Executor Service is not registered in ServiceRegistry

mswiderski:


mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorServiceImpl.java
```diff
@@ -152,6 +153,7 @@ public void init() {
     		try {
 		        executor.init();
 		        this.executorStarted = true;
+		        ServiceRegistry.get().register(org.kie.api.executor.ExecutorService.class.getSimpleName(), this);
```
MarianMacik: Why not to use newly added ServiceRegistry.EXECUTOR_SERVICE also here?

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorServiceImpl.java
```diff
@@ -163,7 +165,8 @@ public void destroy() {
     	if (executorStarted) {
     		ExecutorServiceFactory.resetExecutorService(this);
 	    	this.executorStarted = false;
-	        executor.destroy();	        
+	        executor.destroy();	
+	        ServiceRegistry.get().remove(org.kie.api.executor.ExecutorService.class.getSimpleName());
```
MarianMacik: Same as above.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorServiceImpl.java
mswiderski: replaced

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorServiceImpl.java
mswiderski: replaced

mswiderski:
Jenkins retest this

PR #1335, opened 2018-09-25
JBPM-7806 - Handle async jobs by Migration service

mswiderski:


jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/MigrationAsyncWorkItemHandlerTest.java
```diff
@@ -0,0 +1,344 @@
...
+public class MigrationAsyncWorkItemHandlerTest extends AbstractExecutorBaseTest {
+
+    private PoolingDataSource pds;
+    private UserGroupCallback userGroupCallback;  
+    private RuntimeManager manager;
+    private RuntimeManager manager2;
+    private ExecutorService executorService;
+    private EntityManagerFactory emf = null;
+    
+    private EntityManagerFactory emfErrors = null;
```
MarianMacik: This name might be confusing since it is not really EMF for errors, but for whole jBPM. I would rename emf to emfExecutor and emfErrors to just emf. Does that make sense?

PR #1344, opened 2018-10-09
[JBPM-7837] ExecutorServiceImpl.requeue(Long olderThan) doesn't calcu…

tkobayas:
…late a correct value

- Unit test

It contains non-preferred Thread.sleep(). Hmm, if it's really better to avoid, I would write one more test with requestInfo.setTime(new Date() - 2000ms). Please let me know.


MarianMacik:
First run passed. I will try it 2 more times just to be sure.

jenkins retest this

MarianMacik:
Run 2/3 has passed.

jenkins retest this

MarianMacik:
3 runs have passed in a row, merging

PR #1352, opened 2018-10-16
JBPM-7849 - Process instance uses very long autogenerated Correlation…

mswiderski:
… Keys for sub processes


mswiderski:
Jenkins retest this

PR #1356, opened 2018-10-22
[JBPM-7879] Mvel Script doesn't resolve dependency

tkobayas:
- Unit test

testScriptTaskJava: Success
testScriptTaskMvel: Failure

### DIFF
--- a/jbpm-test-coverage/pom.xml
+++ b/jbpm-test-coverage/pom.xml
@@ -85,6 +85,15 @@
       <artifactId>jbpm-human-task-core</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-kie-services</artifactId>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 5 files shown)


tsurdilo:
@tkobayas hello :) I have updated your pr and made some small changes, as you can see now your tests run fine. 

Whats the difference is that with MVEL you have to do just:

```xml
<bpmn2:script><![CDATA[testpojo = new defaultPackage.TestPojo(); testpojo.say();
]]></bpmn2:script>
```
so dont declare the type before the variable
With java you need to declare the type.

I updated your test class as well as it was failing (initializing runtime manager 2 times). 

Btw I would not commit this pr as-is as it duplicates a ton of methods that are already available in abstract test classes in jBPM. If you want to feel free to update your test class to extend AbstractKieServicesBaseTest and then i think it would make a nice test to be added .

tsurdilo:
@tkobayas looks good to me. thanks

PR #1357, opened 2018-10-24
JBPM-7880 - ElasticSearch event emitter for runtime events

mswiderski:
depends on 
https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/861
https://github.com/kiegroup/droolsjbpm-knowledge/pull/337
https://github.com/kiegroup/drools/pull/2120
https://github.com/kiegroup/jbpm/pull/1357


jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/test/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitterTest.java
```diff
@@ -0,0 +1,327 @@
...
+        // use latch to wait for async processing of the emitter
+        CountDownLatch latch = new CountDownLatch(1);
+        ElasticSearchEventEmitter emitter = new ElasticSearchEventEmitter() {
+
+            @Override
+            protected ExecutorService buildExecutorService() {
+                return createExecutor(latch);
+            }
+            
+        };
```
MarianMacik: Theoretically the entire emitter creation can be done in a separate method.

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -132,49 +135,53 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
...
+                
+                ((ProcessInstanceImpl) processInstance).reconnect();
+                if (readOnly) {
+                    internalRemoveProcessInstance(processInstance);
+                }
+            }
+            return processInstance;
+        } finally {
+            if (!readOnly && processInstance != null) {
+                EventManagerProvider.getInstance().get().update(getInstanceViewFor(processInstance));
```
MarianMacik: So this is for situations when something goes wrong? In such cases we want to push updates anyway?

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/ProcessInstanceView.java
```diff
@@ -186,7 +186,7 @@ public void copyFromSource() {
         this.correlationKey = ((WorkflowProcessInstanceImpl)source).getCorrelationKey();
         this.date = new Date();
         this.id = source.getId();
-        this.initiator = (String) ((WorkflowProcessInstance)source).getMetaData().get("Initiator");
+        this.initiator = (String) ((WorkflowProcessInstanceImpl)source).getVariable("initiator");
```
MarianMacik: What is the difference between these two approaches?

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
mswiderski: it should be added as long as it's not readonly and not null and to avoid adding same line in few places adding that in finally block made sense. It must be done once the process instance was reconnected.

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/ProcessInstanceView.java
mswiderski: variable is always available while meta data only on start of process instance - since meta data is not persisted.

PR #1359, opened 2018-10-29
[DO NOT MERGE] unit test to explain 'import' in Mvel Script Task

tkobayas:
Firstly, I'll send an email to an internal email thread to discuss. Will followup on this PR comments.


tkobayas:
Confirmed with @mswiderski , "import" statement in ScriptTask is not supported. Users should use imports on process definition level (<drools:import name="defaultPackage.MyUtil"/>).

Closing this PR.

PR #1360, opened 2018-10-29
[BAQE-439] Introducing Narayana and tomcat-dbcp into engine tests

rsynek:
@mswiderski @MarianMacik sorry for such a big PR, I reworked the PoolingDataSource API that was spread across the entire code base.

The PR depends on https://github.com/kiegroup/drools/pull/1939.


rsynek:
jenkins retest this

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/HumanTaskServicesBaseTest.java
```diff
@@ -185,10 +185,10 @@ protected JaxbContent xmlRoundTripContent(Content content) {
     protected static final String PASSWORD = "password";
     protected static final String JDBC_URL = "url";
 
-    protected static PoolingDataSource setupPoolingDataSource() {
-        Properties dsProps = getDatasourceProperties();
-        PoolingDataSource pds = PersistenceUtil.setupPoolingDataSource(dsProps, "jdbc/jbpm-ds", false);
-        pds.init();
+    protected static PoolingDataSourceWrapper setupPoolingDataSource() {
+        //Properties dsProps = getDatasourceProperties();
```
mswiderski: remove the commented code if not needed

rsynek:
@mswiderski good question, there are other utilities in the package, but probably not all the modules use them. Will find out where the dependency can be removed.

mswiderski:
@rsynek would be worth checking if we can move them to kie-test-util and thus get rid of one module

rsynek:
@mswiderski there are classes closely related to testing processes (e.g. various listeners); so this doesn't look like an option - there is nothing related to Drools.

mswiderski:
@rsynek ok, let's keep it as is, thanks let me know when this is ready to be merged

rsynek:
@mswiderski the jbpm-test-util is still required by other modules due to the classes I mentioned. Let's merge.

PR #1361, opened 2018-10-29
[JBPM-7824] dataInputAssociation sourceRef unable to resolve property /variable

elguardian:
transform id to variable name during message tag process xml processing using
id metadata


mswiderski:
@elguardian have you seen my comment above? wdyt?

elguardian:
yep. added the test and did the corrections If i didn't miss anything.

elguardian:
rebased @mswiderski

PR #1372, opened 2018-11-09
[JBPM-5215] Reuse deployment descriptor classes from common jBPM module

elguardian:
removing the common classes provided by kie internal and refactor imports in other classes


elguardian:
depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/341

PR #1380, opened 2018-11-19
RHPAM-1648 - Improving process compiler so 'XOR split could not find …

mswiderski:
…at least one valid outgoing connection for split' error is thrown during compile time and not runtime


mswiderski:
Jenkins retest this

PR #1382, opened 2018-11-26
JBPM-7982 - Case Management: CaseFileDataLog table is not populated a…

mswiderski:
…fter task completion, RHPAM-1729 - Errors in server log when importing sample projects - renamed CMMN spec xsd files


sutaakar:
Causes failures in Kie server tests - https://issues.jboss.org/browse/JBPM-8018

PR #1386, opened 2018-11-30
RHPAM-1691 - Error while cancelling command, RHPAM-1687 - Reoccuring …

mswiderski:
…command not executed after temporary database unavailability


mswiderski:
Jenkins retest this

PR #1393, opened 2018-12-11
DROOLS-3355: Improve Assembler/Weaver API

evacchi:
(Maybe I should link this to a jBPM issue)

- ProcessBuilderImpl is not called directly anymore from drools-core, we hook into KnowledgeBuilder via the "assembler/weaver API" (see the JIRA)

this is done for BPMN2, DRF and also CMMN. You'll notice the change amounts to very little code.

Strictly depends on 
- https://github.com/kiegroup/droolsjbpm-knowledge/pull/351
- https://github.com/kiegroup/drools/pull/2181


evacchi:
Jenkins please execute full downstream build

evacchi:
Jenkins please execute full downstream build

evacchi:
Jenkins please execute full downstream build

pszubiak:
@evacchi I need to restart Jenkins and abandon full downstream build of this PR. Let me restart that build just after Jenkins is restarted. Sorry for the inconvenience!

pszubiak:
Jenkins please execute full downstream build

jbpm-flow-builder/src/main/resources/META-INF/kie.conf
```diff
@@ -1 +1,5 @@
-org.drools.compiler.compiler.ProcessBuilderFactoryService=org.jbpm.process.builder.ProcessBuilderFactoryServiceImpl
\ No newline at end of file
+org.drools.compiler.compiler.ProcessBuilderFactoryService=org.jbpm.process.builder.ProcessBuilderFactoryServiceImpl
+org.kie.api.internal.assembler.KieAssemblers = +org.jbpm.assembler.BPMN2AssemblerService
+org.kie.api.internal.weaver.KieWeavers = +org.jbpm.weaver.BPMN2WeaverService
+org.kie.api.internal.assembler.KieAssemblers = +org.jbpm.assembler.DRFAssemblerService
+org.kie.api.internal.weaver.KieWeavers = +org.jbpm.weaver.DRFWeaverService
```
mswiderski: don't we need CMMN WeaverService as well?

evacchi: it's in `jbpm-case-mgmt/jbpm-case-mgmt-cmmn/src/main/resources/META-INF/kie.conf `

evacchi:
thanks @mswiderski you can already take a look but I'm still figuring out why full dowstream is failing :( it seems like it can't find an interface I have generified, but I didn't move it! :/

evacchi:
let's try rebasing :/

Jenkins please execute full downstream build

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/ProcessBuildContext.java
```diff
@@ -31,7 +31,7 @@
     private ProcessDescr processDescr;
     private DialectCompiletimeRegistry dialectRegistry;
 
-    public ProcessBuildContext(final KnowledgeBuilderImpl pkgBuilder,
+    public ProcessBuildContext(final DroolsAssemblerContext pkgBuilder,
```
tsurdilo: From readability perspective this confuses me (the use of "DroolsAssemblerContext" for a "package builder" field inside ProcessBuilderContext...what?

jbpm-flow-builder/src/main/java/org/jbpm/weaver/BPMN2WeaverService.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.weaver;
+
+import org.kie.api.io.ResourceType;
+
+public class BPMN2WeaverService extends AbstractWeaverService {
```
tsurdilo: this info is already available in BPMN2AssemblerService (the resource type). Any way to combine and not have two?

jbpm-flow-builder/src/main/java/org/jbpm/weaver/DRFWeaverService.java
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.weaver;
+
+import org.kie.api.io.ResourceType;
+
+public class DRFWeaverService extends AbstractWeaverService {
```
tsurdilo: same here -- resource type defined in DRFAssemblerService

jbpm-flow-builder/src/main/java/org/jbpm/assembler/AbstractProcessAssembler.java
```diff
@@ -0,0 +1,80 @@
...
+import org.drools.core.impl.InternalKnowledgeBase;
+import org.jbpm.compiler.ProcessBuilderImpl;
+import org.kie.api.definition.process.Process;
+import org.kie.api.internal.assembler.KieAssemblerService;
+import org.kie.api.io.Resource;
+import org.kie.api.io.ResourceConfiguration;
+import org.kie.api.io.ResourceType;
+import org.kie.internal.builder.KnowledgeBuilderResult;
+
+public abstract class AbstractProcessAssembler implements KieAssemblerService {
```
tsurdilo: Take this as nitpicking it's ok :)
kb, kbase,e,es, pkg
doesnt intellij have like auto-suggestions for naming fields based on their type? 

Real question - why are we adding all exceptions except RuntimeException which is being re-thrown? Just curious.

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/ProcessBuildContext.java
evacchi: Yes I agree. I'll change the variable name. This needs a bit of an explanation, anyway. 

Here in jBPM an internal (`Impl`) of Drools was used. `KnowledgeBuilderImpl` has a *huge* API surface. We'll have to **deprecate** that. The `Assembler` mechanism should help with that, but it's unfeasible at this point to avoid using all those methods; as an interim measure, I have extracted all and only the methods of `KnowledgeBuilderImpl` that are used in `ProcessBuildContext` (and a bunch of other places) to the [`DroolsAssemblerContext`](https://github.com/kiegroup/drools/pull/2181/files#diff-8536ce258aa279001c9e66571938b099) interface. This interface should disappear in the future, in favor of a more general API. The name `Assembler` comes from the fact that we have now moved all of this processing in an `Assembler`. and `Drools`, because it's still relying on Drools internals.

jbpm-flow-builder/src/main/java/org/jbpm/weaver/BPMN2WeaverService.java
evacchi: I agree. I may think of a way, but unfortunately weaver and assembler are supposed to be separate. TBF I think this API should benefit another round of review, but this is what we have available right now. See the related JIRA for my thoughts on that.

jbpm-flow-builder/src/main/java/org/jbpm/assembler/AbstractProcessAssembler.java
evacchi: probably an oversight; the `addResource()` signature includes `throws Exception`, and `e` here have not been casted. Will fix. (In fact: notice this originally was a quasi-verbatim copy of [KnowledgeBuilderImpl#addProcessFromXml()](https://github.com/kiegroup/drools/blob/8092938bc75726481deb26e94c70d8167c7ba6e9/drools-compiler/src/main/java/org/drools/compiler/builder/impl/KnowledgeBuilderImpl.java#L703-L737) where the cast exists)

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/ProcessBuildContext.java
evacchi: > Here in jBPM

heh, just to be clear: that's no blaming, we were reusing some code from Drools that relied *itself* too much on internal impl details...

evacchi:
wheel of fortune

Jenkins please execute full downstream build

evacchi:
I really hope these failures are unlelated: `java.net.BindException: Address already in use (Bind failed)`

Jenkins please retest this

evacchi:
Jenkins please execute full downstream build

evacchi:
No related failures in either branch so I think we can declare this ready to merge. Please coordinate with @mariofusco as it requires the other related PRs to be merged at the same time

PR #1400, opened 2018-12-19
JBPM-8087 - Avoid duplicate listeners call when retriggering node ins…

cristianonicolai:
…tance


jsoltes:
Jenkins please execute full downstream build

cristianonicolai:
@jsoltes I just checked the zip dist from the full downstream build here, and here is what I get after retriggering the Self Evaluation node:

selection_011

Looks fine to me.

PR #1403, opened 2018-12-20
RHPAM-1761 Bulk abort throws an exception when called on several subp…

bxf12315:
…rocess instances

@cristianonicolai  I not sure it is right. Please review it.
I will try to add unit test for.(I do no know how to add parent process and sub process).

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
@@ -193,26 +193,40 @@ public class ProcessServiceImpl implements ProcessService, VariablesAware {
 
     @Override
     public void abortProcessInstances(List<Long> processInstanceIds) {
-        processInstanceIds.stream().filter(processInstanceId -> !hasParentProcessInProcessInstanceList(null, processInstanceId, processInstanceIds))
+        processInstanceIds.stream().filter(processInstanceId -> !hasAbortedParentProcessInProcessInstanceList(null,
+                                                                                                              processInstanceId,
+                                                                                                              processInstanceIds))
                 .forEach(processInstanceId -> abortProcessInstance(processInstanceId));
     }
 
     @Override
     public void abortProcessInstances(String deploymentId, List<Long> processInstanceIds) {
-        processInstanceIds.stream().filter(processInstanceId -> !hasParentProcessInProcessInstanceList(deploymentId, processInstanceId, processInstanceIds))
+        processInstanceIds.stream().filter(processInstanceId -> !hasAbortedParentProcessInProcessInstanceList(deploymentId,
+                                                                                                              processInstanceId,
+                                                                                                              processInstanceIds))
                 .forEach(processInstanceId -> abortProcessInstance(deploymentId, processInstanceId));
     }
 
-    private boolean hasParentProcessInProcessInstanceList(String deploymentId, long processInstanceId, List<Long> processInstanceIds) {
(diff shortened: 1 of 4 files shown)


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/ProcessServiceImpl.java
```diff
@@ -190,21 +190,33 @@ public void abortProcessInstance(String deploymentId, Long processInstanceId) {
...
         }
 	}
-	
-	@Override
+
+    @Override
     public void abortProcessInstances(List<Long> processInstanceIds) {
-	    for (long processInstanceId : processInstanceIds) {
-            abortProcessInstance(processInstanceId);
+        processInstanceIds.stream().filter(processInstanceId -> !hasParentProcessInProcessInstanceList(null, processInstanceId, processInstanceIds))
```
cristianonicolai: what if you just sort the process instances based on their id, and start aborting the highest ones first? a child process should have a higher id than its parent.

bxf12315: as your said , after close parent process then sub proccess also will close. but the sub process closed, we will get same issue. so I think we just need close parent process , skip all sub processes of it.

cristianonicolai: what if you just want to abort the sub-process in here? That wouldn't be possible anymore.

bxf12315: just abort sub process, maybe can't abort   
parent process. the issue is case is very special

bxf12315: just abort sub process, maybe can't abort   
parent process. the issue is case is very special

cristianonicolai: if you abort the subprocess then the parent process, it should work. The issue now is that we abort the parent, then abort the child process, which has been aborted already.
Or you can check if a parent process has been aborted already, in the list of ids.

bxf12315: you means, after abort subprocesses, if the parent process already aborted, we can abort it again ? 
I find use getProcessInstance of ProcessServiceImpl, I can not get aborted process, it will be null.

cristianonicolai: No, what I mean is that you should check if the parent process id of a subprocess is in the list of process ids to be aborted. If it is, you can skip aborting that particular instance as you know that the parent will be aborted anyway and consequently that sub process.

bxf12315: As you said, I try to refactor these code. You can review the code.
I find ProcessServiceImpl  can not get aborted process instance , only RuntimeDataServiceImpl can get aborted and completed process instance. if we want to get aborted process instance , we must call RuntimeDataServiceImpl, I not sure it is right  and these is do not work.

bxf12315: ProcessServiceImpl can get process instance in ProcessInstance table , RuntimeDataServiceImpl can get process instance in ProcessInstancelog table.
In ProcessInstance, if the process aborted and completed, the instance will remove.

bxf12315:
@cristianonicolai  I think I can close the PR.

PR #1415, opened 2019-01-21
Updated timeout to 15 sec (stabilization)

gmunozfe:
BAQE-822 - Increase timeout for testExecuteProcessWithAsyncHandler 
testExecuteProcessWithAsyncHandler fails frequently at slower machines (w2k16 mainly) due to execution time is close to defined timeout. Increased for stabilization.


jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/BasicExecutorBaseTest.java
```diff
@@ -622,7 +622,11 @@ public void testPrioritizedJobsExecution() throws InterruptedException {
         assertNotNull(executedLow);
         assertEquals("low priority", executedLow.getKey());
         
-        assertTrue(executedLow.getTime().getTime() > executedHigh.getTime().getTime());
+        logger.info("executedLow: {}", executedLow.getTime().getTime());
+        logger.info("executedHigh: {}", executedHigh.getTime().getTime());
+        logger.info("exec difference: {}", (executedLow.getTime().getTime() - executedHigh.getTime().getTime()));
```
MarianMacik: Can you please delete this logging? We don't need this anymore since we verified our theory about the issue.

PR #1422, opened 2019-01-24
JBPM-8000 - update jbpm workitems so can be used in workitem repository

tsurdilo:


mswiderski:
replaced by https://github.com/kiegroup/jbpm/pull/1423 so we have complete PR build

PR #1424, opened 2019-01-25
AF-1685: Development Lifecycle Streamline

pefernan:
AF-1687: Align with 'kie-server' with development mode

Added extra param to decide if existing process instances must be aborted or not when undeploying.

@mswiderski @tomasdavidorg could you please take a look?

This is part of an ensemble, please merge with:
https://github.com/kiegroup/appformer/pull/614
https://github.com/kiegroup/jbpm/pull/1424
https://github.com/kiegroup/droolsjbpm-integration/pull/1698
https://github.com/kiegroup/kie-wb-common/pull/2422

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
@@ -16,20 +16,22 @@
 
 package org.jbpm.kie.services.impl;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.function.Function;
 
 import javax.persistence.EntityManagerFactory;
 
 import org.jbpm.kie.services.api.DeploymentIdResolver;
 import org.jbpm.kie.services.impl.audit.ServicesAwareAuditEventBuilder;
 import org.jbpm.kie.services.impl.security.IdentityRolesSecurityManager;
+import org.jbpm.kie.services.impl.utils.PreUndeployOperations;
 import org.jbpm.persistence.api.integration.EventManagerProvider;
 import org.jbpm.process.audit.event.AuditEventBuilder;
 import org.jbpm.runtime.manager.impl.AbstractRuntimeManager;
@@ -41,19 +43,13 @@ import org.jbpm.services.api.ListenerSupport;
 import org.jbpm.services.api.RuntimeDataService;
 import org.jbpm.services.api.model.DeployedUnit;
 import org.jbpm.services.api.model.DeploymentUnit;
-import org.jbpm.services.api.model.ProcessInstanceDesc;
 import org.kie.api.runtime.KieContainer;
-import org.kie.api.runtime.KieSession;
-import org.kie.api.runtime.manager.RuntimeEngine;
 import org.kie.api.runtime.manager.RuntimeEnvironment;
 import org.kie.api.runtime.manager.RuntimeManager;
 import org.kie.api.runtime.manager.RuntimeManagerFactory;
-import org.kie.api.runtime.process.ProcessInstance;
-import org.kie.api.runtime.query.QueryContext;
 import org.kie.internal.identity.IdentityProvider;
 import org.kie.internal.runtime.conf.DeploymentDescriptor;
 import org.kie.internal.runtime.manager.InternalRuntimeManager;
-import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -176,50 +172,30 @@ public abstract class AbstractDeploymentService implements DeploymentService, Li
         }
         
     }
-    
+
     @Override
     public void undeploy(DeploymentUnit unit) {
-        undeploy(unit, false);
+        undeploy(unit, PreUndeployOperations.checkActiveProcessInstances(runtimeDataService));
     }
 
     @Override
-    public void undeploy(DeploymentUnit unit, boolean abortInstances) {
+    public void undeploy(DeploymentUnit unit, Function<DeploymentUnit, Boolean> beforeUndeploy) {
 
-        List<Integer> states = new ArrayList<Integer>();
-        states.add(ProcessInstance.STATE_ACTIVE);
-        states.add(ProcessInstance.STATE_PENDING);
-        states.add(ProcessInstance.STATE_SUSPENDED);
-        Collection<ProcessInstanceDesc> activeProcesses = runtimeDataService.getProcessInstancesByDeploymentId(unit.getIdentifier(), states, new QueryContext());
-        if (!abortInstances && !activeProcesses.isEmpty()) {
-            throw new IllegalStateException("Undeploy forbidden - there are active processes instances for deployment "
-                                                    + unit.getIdentifier());
-        }
+        Function<DeploymentUnit, Boolean> beforeUpdate = Optional.ofNullable(beforeUndeploy).orElse(PreUndeployOperations.checkActiveProcessInstances(runtimeDataService));
 
-        synchronized (this) {
-            DeployedUnit deployed = deploymentsMap.remove(unit.getIdentifier());
+        if (Boolean.TRUE.equals(beforeUpdate.apply(unit))) {
+            synchronized (this) {
+                DeployedUnit deployed = deploymentsMap.remove(unit.getIdentifier());
 
-            if (deployed != null) {
-
-                RuntimeManager manager = deployed.getRuntimeManager();
-
-                if (abortInstances && !activeProcesses.isEmpty()) {
-                    activeProcesses = runtimeDataService.getProcessInstancesByDeploymentId(unit.getIdentifier(), states, new QueryContext(0, -1));
-                    activeProcesses.forEach(processInstanceDesc -> {
-                        RuntimeEngine engine = manager.getRuntimeEngine(ProcessInstanceIdContext.get(processInstanceDesc.getId()));
-                        try {
-                            KieSession ksession = engine.getKieSession();
-                            ksession.abortProcessInstance(processInstanceDesc.getId());
-                        } catch (Exception ex) {
-                            logger.error("Error aborting process instance {} due to: {}", processInstanceDesc.getId(), ex.getMessage());
-                        } finally {
-                            manager.disposeRuntimeEngine(engine);
-                        }
-                    });
+                if (deployed != null) {
+                    RuntimeManager manager = deployed.getRuntimeManager();
+                    ((AbstractRuntimeManager) manager).close(true);
                 }
-
-                ((AbstractRuntimeManager)manager).close(true);
+                notifyOnUnDeploy(unit, deployed);
             }
-            notifyOnUnDeploy(unit, deployed);
+        } else {
+            logger.warn("Couldn't undeploy unit '{}'", unit.getIdentifier());
+            throw new IllegalStateException("Couldn't undeploy unit '" + unit.getIdentifier() + "'");
         }
     }
 
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
@@ -185,6 +185,16 @@ public abstract class AbstractDeploymentService implements DeploymentService, Li
     @Override
     public void undeploy(DeploymentUnit unit, boolean abortInstances) {
 
+        List<Integer> states = new ArrayList<Integer>();
+        states.add(ProcessInstance.STATE_ACTIVE);
+        states.add(ProcessInstance.STATE_PENDING);
+        states.add(ProcessInstance.STATE_SUSPENDED);
+        Collection<ProcessInstanceDesc> activeProcesses = runtimeDataService.getProcessInstancesByDeploymentId(unit.getIdentifier(), states, new QueryContext());
+        if (!abortInstances && !activeProcesses.isEmpty()) {
+            throw new IllegalStateException("Undeploy forbidden - there are active processes instances for deployment "
+                                                    + unit.getIdentifier());
+        }
+
         synchronized (this) {
             DeployedUnit deployed = deploymentsMap.remove(unit.getIdentifier());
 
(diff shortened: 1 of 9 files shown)


pefernan:
jenkins restest this please

pefernan:
@mswiderski thanks! better now?

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/DeploymentService.java
```diff
@@ -43,6 +43,14 @@
 	 * @throws RuntimeException in case of problems encountered while deploying unit
 	 */
     void undeploy(DeploymentUnit unit);
+
+    /**
+     * Performs undeployment operation of given <code>unit</code> aborting existing process instances if needed.
+     * @param unit deployment unit to be undeployed from runtime
+     * @param abortInstances determines if existing process instances should be aborted or not.
+     * @throws RuntimeException in case of problems encountered while deploying unit
```
MarianMacik: should be while **un**deploying unit

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
```diff
@@ -176,19 +179,40 @@ public void commonDeploy(DeploymentUnit unit, DeployedUnitImpl deployedUnit, Run
     
     @Override
     public void undeploy(DeploymentUnit unit) {
-        List<Integer> states = new ArrayList<Integer>();
-        states.add(ProcessInstance.STATE_ACTIVE);
-        states.add(ProcessInstance.STATE_PENDING);
-        states.add(ProcessInstance.STATE_SUSPENDED);
-        Collection<ProcessInstanceDesc> activeProcesses = runtimeDataService.getProcessInstancesByDeploymentId(unit.getIdentifier(), states, new QueryContext());
-        if (!activeProcesses.isEmpty()) {
-            throw new IllegalStateException("Undeploy forbidden - there are active processes instances for deployment "
```
MarianMacik: So there will be no more exception when undeploying in production for example? I get it for "dev mode" but for production we probably should inform user that there are still active process instances.

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/DeploymentService.java
pefernan: @MarianMacik touché!

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
pefernan: @MarianMacik yes that's correct. The kie-sever is the one that takes care of the mode (development or regular) and will call the right method depending on it. The jBPM runtime won't have any knowledge about the mode and will rely on the params received.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
```diff
@@ -176,19 +179,40 @@ public void commonDeploy(DeploymentUnit unit, DeployedUnitImpl deployedUnit, Run
...
+
         synchronized (this) {
             DeployedUnit deployed = deploymentsMap.remove(unit.getIdentifier());
+
             if (deployed != null) {
+
                 RuntimeManager manager = deployed.getRuntimeManager();
+
+                if(abortInstances) {
+                    List<Integer> states = new ArrayList<>();
```
tomasdavidorg: I think, this `states` list could be moved outside of this method to avoid repetitive initialization.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
```diff
@@ -176,19 +179,40 @@ public void commonDeploy(DeploymentUnit unit, DeployedUnitImpl deployedUnit, Run
...
+    public void undeploy(DeploymentUnit unit, boolean abortInstances) {
+
         synchronized (this) {
             DeployedUnit deployed = deploymentsMap.remove(unit.getIdentifier());
+
             if (deployed != null) {
+
                 RuntimeManager manager = deployed.getRuntimeManager();
+
+                if(abortInstances) {
```
tomasdavidorg: Missing space behind `if`

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
MarianMacik: @pefernan I discussed this with @mswiderski and we came to a conclusion that if `abortInstances` is false, then it should throw exception as it did until now if there are active processes. If `abortInstances` is true, then it should abort them and not throw exception.

mswiderski: looks better now

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/KModuleDeploymentServiceTest.java
```diff
@@ -254,7 +253,61 @@ public void testUnDeploymentWithActiveProcesses() {
...
+        DeployedUnit deployedGeneral = deploymentService.getDeployedUnit(deploymentUnit.getIdentifier());
+        assertNotNull(deployedGeneral);
+        assertNotNull(deployedGeneral.getDeploymentUnit());
+        assertNotNull(deployedGeneral.getRuntimeManager());
+
+        RuntimeManager manager = deploymentService.getRuntimeManager(deploymentUnit.getIdentifier());
+        assertNotNull(manager);
+
+        RuntimeEngine engine = manager.getRuntimeEngine(EmptyContext.get());
+        assertNotNull(engine);
```
mswiderski: please dispose the engine once done with processes

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/KModuleDeploymentServiceTest.java
```diff
@@ -254,7 +253,61 @@ public void testUnDeploymentWithActiveProcesses() {
 
         engine.getKieSession().abortProcessInstance(processInstance.getId());
 
-        checkFormsDeployment( deploymentUnit.getIdentifier() );
+        checkFormsDeployment(deploymentUnit.getIdentifier());
+    }
+
+    @Test
+    public void testUnDeploymentWithoutActiveProcesses() {
```
mswiderski: not sure this test has much value to be honest ....

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
```diff
@@ -176,19 +179,46 @@ public void commonDeploy(DeploymentUnit unit, DeployedUnitImpl deployedUnit, Run
...
         }
+
         synchronized (this) {
             DeployedUnit deployed = deploymentsMap.remove(unit.getIdentifier());
+
             if (deployed != null) {
+
                 RuntimeManager manager = deployed.getRuntimeManager();
+
+                if(abortInstances && !activeProcesses.isEmpty()) {
```
jhrcek: Formatting..
```suggestion
                if (abortInstances && !activeProcesses.isEmpty()) {
```

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
```diff
@@ -176,19 +179,46 @@ public void commonDeploy(DeploymentUnit unit, DeployedUnitImpl deployedUnit, Run
...
+                    activeProcesses = runtimeDataService.getProcessInstancesByDeploymentId(unit.getIdentifier(), states, new QueryContext(0, -1));
+                    if (!activeProcesses.isEmpty()) {
```
jhrcek: The if check is redundat. You can safely remove it, because `forEach(..)` method called on empty Iterable just does nothing.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
jhrcek: You don't have to reformat the whole class, but adding spaces after if / for etc shouldn't need reformatting the whole class. Up to you..

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/KModuleDeploymentServiceTest.java
pefernan: @mswiderski done, I also disposed it on other tests...

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
pefernan: @jhrcek done!

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
```diff
@@ -176,19 +179,44 @@ public void commonDeploy(DeploymentUnit unit, DeployedUnitImpl deployedUnit, Run
...
+
                 RuntimeManager manager = deployed.getRuntimeManager();
+
+                if (abortInstances && !activeProcesses.isEmpty()) {
+                    activeProcesses = runtimeDataService.getProcessInstancesByDeploymentId(unit.getIdentifier(), states, new QueryContext(0, -1));
+                    activeProcesses.forEach(processInstanceDesc -> {
+                        RuntimeEngine engine = manager.getRuntimeEngine(ProcessInstanceIdContext.get(processInstanceDesc.getId()));
+                        try {
+                            KieSession ksession = engine.getKieSession();
+                            ksession.abortProcessInstance(processInstanceDesc.getId());
```
pefernan: @mswiderski we should reconsider throwing this Exception, if we throw it it won't be possible to update a snapshot without aborting process instances. WDYT?

pefernan: maybe we can add an extra param, to skip the check.

mswiderski: I see your point, instead of adding another parameter maybe we could use Function<Collection<ProcessInstanceDesc>, Boolean> that would be responsible for dealing with these checks and operations. We could then have few implementation
- default that checks if the list is empty and if so disallow underlay
- one that aborts all process instances from the list and if all were done successfully returns true to proceed with undeploy
- another one that simply ignores the list of instances and ask to underplay anyway

then Kie Server based on all the info it has can select one of them to make the undeploy/update/etc

wdyt?

mswiderski:
@pefernan just a side note, if you're good with the PR from your side please remove DO NOT MERGE as I tend to ignore such PRs...

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractDeploymentService.java
pefernan: Probably what's the best solution. What do you think about instead of using a Function use a Function<DeploymentUnit, Boolean> and let the consumer be the one that gets the active instances and decides what to do, there might be cases that we don't need any check and we can just skip the query.

pefernan: @mswiderski ^^

mswiderski: maybe it is more concise but it does not reflect the need as it is expected to perform actions inside it. So fine to use DeploymentUnit to avoid extra query) but we should not use Predicate

pefernan: @mswiderski +1

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
```diff
@@ -222,12 +223,12 @@ protected RegisterableItemsFactory getRegisterableItemsFactory(AuditEventBuilder
     }
 
     @Override
-	public void undeploy(DeploymentUnit unit) {
+	public void undeploy(DeploymentUnit unit, Function<DeploymentUnit, Boolean> beforeUndeploy) {
```
jhrcek: I'd recommend using `Predicate<DeploymentUnit>` instead of `Function<DeploymentUnit, Boolean>` for this. Apart from being shorter you don't have to deal with the case when user potentially returns null (which is valid Boolean value)

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/DeploymentService.java
```diff
@@ -40,9 +42,18 @@
...
-	 * @throws RuntimeException in case of problems encountered while deploying unit
+	 * @throws RuntimeException in case of problems encountered while undeploying unit
 	 */
     void undeploy(DeploymentUnit unit);
+
+    /**
+     * Performs undeployment operation of given <code>unit</code> aborting existing process instances if needed.
+     * @param unit deployment unit to be undeployed from runtime
+     * @param beforeUndeploy a function to run custom actions before undeploying a given {@link DeploymentUnit} . It should return if the unit can be deployed or not
+     * determining if the deployment can continue.
```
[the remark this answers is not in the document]
pefernan: @MarianMacik heh! yea sometimes my english...

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
mswiderski: there is no need to deal with null as only true is the important one ... everything else will produce an error

MarianMacik: I think if you return Boolean as null, then if (null) will fail with NPE in the undeploy method...

pefernan:
@mswiderski better now?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
mswiderski: that's true but we are not doing defencing programming to guard any possible user mistakes, if someone returns null then it's expected to fail with NPE

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/utils/PreUndeployOperations.java
```diff
@@ -0,0 +1,108 @@
...
+                    manager.disposeRuntimeEngine(engine);
+                }
+            }
+
+            return true;
+        };
+    }
+
+    /**
+     * Returns a function that bypasses the check and always always allows to undeploy.
```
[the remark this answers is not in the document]
pefernan: @MarianMacik applied all your changes... your translations saved me! thanks!

pefernan:
Fixed conflicts... jenkins retest this please!

pefernan:
jenkins retest this please

PR #1426, opened 2019-01-29
Index changes to address deadlocks on MySQL based on support feedback…

martinweiler:
… (KBase #3063771)


MarianMacik:
Looks good to me, but are we ware of losing a bit of referential integrity since we are deleting one foreign key? Moreover, I see @sutaakar has this JIRA tested in the past, so probably he is a better contact to review.

sutaakar:
Ok for me apart from point brought by Marian. Not sure what consequences it may bring.

mswiderski:
I'd say it bring more good than potential issues so I opt at merging as it was already verified by several users, @MarianMacik @sutaakar any objections?

sutaakar:
@mswiderski Fine for me

PR #1427, opened 2019-01-30
Fix NullPointerException when undeploy

meadlai:
When call KModuleDeploymentService.undeploy#230, it will unregister forms, if the kjar doesn't contains any forms, then it throws NullPointerException


kiereleaseuser2:

PR #1429, opened 2019-01-30
[BAPL-1070] RHPAM graceful shutdown (DO NOT MERGE)

elguardian:
registering listener to destroy the executor once is shutdown.


mswiderski:
@elguardian not sure we achieve a lot with this as it simply delays the shutdown as currently we rely on sort of PreDestroy behaviour which is similar. I think the request was about to sort of allow to finish active jobs but disallow scheduling new... @tkobayas could you please comment?

tkobayas:
@elguardian @mswiderski 

> I think the request was about to sort of allow to finish active jobs but disallow scheduling new...

Yes, exactly. I also posted a comment on BAPL-1070

PR #1431, opened 2019-01-31
add new feature to mask the original sender's address. 

ghjboss:
see https://issues.jboss.org/browse/RHPAM-1854 for details.

This new feature introduces a new variable name called "DisplayName" passed in EmailWorkItemHandler. 
This variable holds the value to show the name for displaying instead of the sender's original email address.  
The main change in the code is in SendHtml.java:

previous code:
            msg.setFrom(new InternetAddress(from));

proposed changes:

            if(null != displayName) {
                msg.setFrom(new InternetAddress(from, displayName));
            } else {
                msg.setFrom(new InternetAddress(from));
            }


krisv:
ok to test

PR #1434, opened 2019-02-08
RuleSetNodeInstance should pass null variable values to DMN

SimonRinguette:
In the DMN engine, there is a difference between passing a variable with a null value and not passing a variable. If a DMN model would support receiving a null input. it is currently impossible to invoke this model from BPMN. The DMN model will complain about a missing input requirement (since the null value variable was not passed) and the execution will not append. With this fix, null value variables are passed to DMN and the execution works.

### DIFF
--- a/jbpm-workitems/jbpm-workitems-bpmn2/src/test/java/org/jbpm/process/workitem/bpmn2/BusinessRuleTaskTest.java
+++ b/jbpm-workitems/jbpm-workitems-bpmn2/src/test/java/org/jbpm/process/workitem/bpmn2/BusinessRuleTaskTest.java
@@ -125,6 +125,48 @@ public class BusinessRuleTaskTest {
                      processInstance.getState());
     }
 
+    @Test
+    public void testDecisionPassingNullToDMN() throws Exception {
+        //This test make sure that null variables are passed to the DMN execution context, otherwise DMN will throw an exception that an input requirement is missing
+        KieBase kbase = readKnowledgeBase();
+        KieSession ksession = createSession(kbase);
+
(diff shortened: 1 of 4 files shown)


etirelli:
jenkins test this

etirelli:
@SimonRinguette any chance you have a unit test for this? 

The change absolutely makes sense, it is just that we need a unit test for each PR that gets merged.

SimonRinguette:
Let me get something together

SimonRinguette:
@etirelli There you go, without my fix, the test case fails. Its a simplified (as much as possible) way of reproducing it.

mswiderski:
ok to test

jbpm-workitems/jbpm-workitems-bpmn2/src/test/java/org/jbpm/process/workitem/bpmn2/BusinessRuleTaskTest.java
```diff
@@ -125,6 +125,27 @@ public void testDecisionTaskProcess() throws Exception {
...
+        params.put("Input",
+                   null);
+
+        WorkflowProcessInstance processInstance = (WorkflowProcessInstance) ksession.startProcess("passthru",
+                                                                                                  params);
+        Object output = processInstance.getVariable("Output");
+
+        assertEquals(null,
+                     output);
+    }
```
MarianMacik: Just to extend the coverage a little bit, would it be possible to add another test with non-null values? I think this kind of test is not present here yet.

SimonRinguette: I have added a simple test with a non-null value using the same model

PR #1438, opened 2019-02-14
[JBPM-8038] Upgrading to 7.15: AsyncTaskLifeCycleEventProducer

elguardian:
excluding org.jbpm.services.task.audit.jms.AsyncTaskLifeCycleEventProducer for only cdi applications


elguardian:
@mswiderski added the test so you can see where the errors happens. if the task audit is included this happens.
I only thinks this is only expected to be used by newJMSInstance

PR #1441, opened 2019-02-15
[JBPM-8077] When a case goes straight through from start to end, case file data is not persisted in the logs.

elguardian:
adding variables as parameters upon process instance start
this avoids the problem of start/complete in the same sentence causing vars not logged

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
@@ -22,12 +22,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 
-import org.drools.core.common.InternalKnowledgeRuntime;
 import org.drools.core.definitions.rule.impl.RuleImpl;
-import org.drools.core.event.ProcessEventSupport;
 import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
 import org.jbpm.casemgmt.impl.event.CaseEventSupport;
-import org.jbpm.process.instance.InternalProcessRuntime;
+import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.services.api.ProcessService;
 import org.kie.api.KieServices;
 import org.kie.api.command.BatchExecutionCommand;
@@ -38,7 +36,6 @@ import org.kie.api.event.rule.DefaultAgendaEventListener;
 import org.kie.api.event.rule.MatchCreatedEvent;
 import org.kie.api.runtime.Context;
 import org.kie.api.runtime.EnvironmentName;
-import org.kie.api.runtime.KieRuntime;
 import org.kie.api.runtime.KieSession;
 import org.kie.api.runtime.process.ProcessInstance;
 import org.kie.internal.KieInternalServices;
@@ -123,9 +120,14 @@ public class StartCaseCommand extends CaseCommand<Void> {
         Map<String, Object> params = new HashMap<>();
         // set case id to allow it to use CaseContext when creating runtime engine
         params.put(EnvironmentName.CASE_ID, caseId);
+        final Map<String, Object> caseData = caseFile.getData();
+
+        for (Map.Entry<String, Object> entry : caseData.entrySet()) {
+            params.put(VariableScope.CASE_FILE_PREFIX + entry.getKey(), entry.getValue());
+        }
+
         final long processInstanceId = processService.startProcess(deploymentId, caseDefinitionId, correlationKey, params);
         logger.debug("Case {} successfully started (process instance id {})", caseId, processInstanceId);        
-        final Map<String, Object> caseData = caseFile.getData();
 
         processService.execute(deploymentId, CaseContext.get(caseId), new ExecutableCommand<Void>() {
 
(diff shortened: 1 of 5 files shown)


MarianMacik:
jenkins retest this

mswiderski:
@elguardian can you add a test case? I'd like to look into it more ...

elguardian:
done @mswiderski . added test case.

mswiderski:
@elguardian after looking into it more I'd say this is kind of invalid use case. The main reason why the variables are not stored is that process instance is already completed when it tries to update the variables in log. Though this is only the initial state of variables. If variables are updated as part of case instance that will be properly reflected in the logs. 

Moreover, this use case is like minimalistic straight through process rather than a case definition.

We cannot just copy case file data into process variables just to populate the logs as that means variables will shadow case file data and thus will get out of sync and in most of the cases will result in inconsistent state. SO I opt at rejecting this jira

elguardian:
@mswiderski isn't it a bit inconsistent the behavior ? seems to me that depending on the situation you will get logged the case data or not (actually that is the complain in the jira).  I get the feeling of some sort of race condition.
you are right that if in the middle of the process there is a change of var this will be set on logs.
regarding the shadowing ? I think this is taking into account int the var scope instance:
https://github.com/kiegroup/jbpm/blob/7.17.0.Final/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L121

mswiderski:
> @mswiderski isn't it a bit inconsistent the behavior ? seems to me that depending on the situation you will get logged the case data or not (actually that is the complain in the jira). I get the feeling of some sort of race condition.

correct, but the thing here is that the use case like shown here is not really case management use case and that's why I said I opt at skipping it.

> you are right that if in the middle of the process there is a change of var this will be set on logs.
regarding the shadowing ? I think this is taking into account int the var scope instance:
https://github.com/kiegroup/jbpm/blob/7.17.0.Final/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L121

we had a talk with @MarianMacik about it and you're both right that the shadow problem should not happen as the variable prefixed with caseFile_ is written to case file and not to process instance. So assuming this has been verified and confirmed to work like that we can take this in.

elguardian:
@mswiderski what is the status of this ? do I need to perform any changes here ?

mswiderski:
Just to make the build green

mswiderski:
Jenkins retest this

MarianMacik:
@mswiderski Just one question - if you then try to query variable in a process instance, e.g. caseFile_myVar. Then according to this code:

https://github.com/kiegroup/jbpm/blob/c6a9e77ce1468da6a31c3888c6c77a0703918438/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L52-L88

It will be obtained from the variables map (since it is not null), which means it will query process variables, not case file. As I understand it, since we put everything as a normal variable to a process by this fix, then we won't be able to read case file variables (just write to them) and the shadowing in the end happens?

elguardian:
@MarianMacik 
you cannot set variables that way. all variables with the prefix will end up in the case data

https://github.com/kiegroup/jbpm/blob/7.17.0.Final/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L121

so in theory the case prefix won't end up in the variables scope but case data. That is the reason you don't need any control in the get. just the set. the VariableScopeInstance acts as façade and I think all the ops goes through this class.

MarianMacik:
@elguardian yes, that's right. But if you pass all caseFile_ prefixed variables as a parameters to the process, they will be put in variables map like any other process variable, won't they? After that, when you call:

https://github.com/kiegroup/jbpm/blob/c6a9e77ce1468da6a31c3888c6c77a0703918438/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L52-L58

you will always query the prefixed variables from the process variables map. But in case of a write to these prefixed variables, you won't update variables map, but you will insert to the case data, as you pointed out. And then when you query for the actual value afterwards, you will get the value from variables map, which will contain the old value from the time when you started a process instance. Does that make sense? Maybe there is some catch which I am missing.

elguardian:
@MarianMacik, No they won't 
https://github.com/kiegroup/jbpm/blob/7.18.0.Final/jbpm-flow/src/main/java/org/jbpm/process/instance/AbstractProcessInstanceFactory.java#L59
It will use the VariableScopeInstance façade. You can try to debug the use case I added here.

MarianMacik:
@elguardian Sorry, I completely overlooked that internalSetVariable is called also when starting a process instance. Now it makes sense to me.

mswiderski:
Jenkins retest this

elguardian:
Jenkins retest this

MarianMacik:
jenkins retest this

elguardian:
those three errors don't look related.

MarianMacik:
hmm, according to the stackTrace it fails with NPE when setting a variable for a case. I don't think it is a random fail.

elguardian:
@MarianMacik you were right... if you try to set case variables before the process is started it fails. it is a subtle difference in behaviour regarding the case data.

elguardian:
@mswiderski any more thoughts ?

mswiderski:
@elguardian can you please sum up what is the change in behaviour?

elguardian:
@mswiderski 
When the case is started it forces to trigger the persistence of the variables in the log. Due to the fact that VariableScopeInstance façace we ensure that the data is stored in case file.

Because during the internal set of variables of a still pending process 
https://github.com/kiegroup/jbpm/pull/1441/files#diff-379e2da8beb6b3317d13309a5a53ffe0R137
this can only happen after the process is started.

That sum up both changes.

mswiderski:
so will the case file data given at start of the case still trigger rules?

elguardian:
nope during process initilization. Keep in mind that at that moment there is no factHandle.

mswiderski:
my question was more if rules based on initial case data will trigger or not. I guess they will as all tests seem to pass and I believe there are tests already covering this... so I think we are good with it

@MarianMacik any comments?

MarianMacik:
I think the answer is here:
https://github.com/kiegroup/jbpm/blob/c5b03f0d49881caa06f8425fab2a73bcd01bc119/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java#L84-L130

CaseFile is inserted into working memory before the process is even started. My question just is that this only works now for rules which inserts something else into working memory, meaning that if they signal something to process instance, it won't work since process instance is not started yet?

Anyway, I am completely ok with the current implementation.

PR #1443, opened 2019-02-15
Raise scope of jaxws-api for jbpm-workitems-bpmn2

MarianMacik:
org.apache.cxf:cxf-rt-frontend-jaxws depends on JAX-WS API (at compile time), but for JDK 11 these classes are not part of JDK. This can cause issues in any component which uses jbpm-workitems-bpmn2 artifact, i.e. Spring Boot.

It is interesting that although Apache CXF 3.3.0 officially supports JDK 11, the dependency on JAX-WS API still wasn't added, although now we use version 3.1.16.

### DIFF
--- a/jbpm-workitems/jbpm-workitems-bpmn2/pom.xml
+++ b/jbpm-workitems/jbpm-workitems-bpmn2/pom.xml
@@ -42,6 +42,10 @@
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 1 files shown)


MarianMacik:
jenkins execute compile downstream build

MarianMacik:
jenkins execute compile downstream build

PR #1444, opened 2019-02-18
Remove hibernate-entitymanager (obsolete)

gmunozfe:
Since hibernate 5.2, hibernate-entitymanager has been marked as deprecated and all its functionality can be obtained from hibernate-core component.

Besides this replacement, hibernate-entitymanager profile can be deleted as it is not used.


jbpm-container-test/jbpm-in-container-test/shrinkwrap-war-profiles/pom.xml
```diff
@@ -215,7 +205,7 @@
         </dependency>
         <dependency>
           <groupId>org.jbpm</groupId>
-          <artifactId>idep-hibernate-entitymanager</artifactId>
+          <artifactId>idep-hibernate-core</artifactId>
```
MarianMacik: If we change this for simple hibernate-core artifact, then idep-hibernate-entitymanager will not be used anywhere. The main reason why we had a wrapper for it was that before we used it directly with shrinkwrap when building an archive. But since now this is a part of tomcat-specific profile, we don't need a nice wrapper around it. What do you think?

gmunozfe: For the tomcat profile, I agree, we may directly include the hibernate-core dependency, without wrapper. I chose this option for changing as less as possible

mswiderski:
@gmunozfe @MarianMacik is this ready for merge?

MarianMacik:
@mswiderski yes

PR #1445, opened 2019-02-18
[JBPM-8252] Fail Fast ExecutorImpl

elguardian:
Check whether those sys properties can be empty or not, as well as found or not.
it does not stop kie-server from being deployed.

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorImpl.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/ExecutorImpl.java
@@ -40,7 +40,9 @@ import javax.jms.MessageProducer;
 import javax.jms.Queue;
 import javax.jms.Session;
 import javax.jms.TextMessage;
+import javax.naming.Context;
 import javax.naming.InitialContext;
+import javax.naming.NamingException;
 
 import org.apache.commons.io.input.ClassLoaderObjectInputStream;
 import org.drools.core.process.instance.WorkItem;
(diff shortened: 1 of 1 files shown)


mswiderski:
have you tried this on tomcat or spring boot deployments? I believe that it might fail as there is no JMS at all but property is on by default...

elguardian:
@mswiderski you were right... tried with spring boot samples in droosjbpm-integration and it was failing. I added some logic
1) if the property is set, we cannot overwrite
2) if the property is not set, only consider default value if there is a java ee container underneath and false otherwise. I use standard  "java:app/AppName" to detect the container. It should not exist in servlet container as this only exists within an ear.

mswiderski:
@elguardian the more I think about it and all the checks in the code I believe it's not worth it, @MarianMacik what do you guys think?

MarianMacik:
Well, according to: https://stackoverflow.com/questions/3431357/how-can-i-config-module-and-application-name-for-jndi-lookups

appName is available only for ears and modulename is available also for wars.

The other thing is that we are changing the backward compatibility behaviour right now. I bet that a large portion of users don't use JMS and they didn't set up anything explicitly. They just rely on falling back to ThreadPool. As a compromise we might fail fast only if a user explicitly specified to use JMS, i.e. system property `org.kie.executor.jms` is set to true.

elguardian:
@MarianMacik not really sure but before the change jms was used by default if the queue was found; e.g: EAP or any app server, but not other runtimes...
The problem here is that you cannot fail fast in environments like spring boot because by default before the use of jms was true but because of the fail fast this cannot be long the policy.

anycase if @mswiderski agrees with you I can change the behaviour.

mswiderski:
I personally opt as dropping this change altogether ... seems to be rather complex logic for a feature that is kind of nice to have. Objections?

elguardian:
none @mswiderski closing it.

PR #1446, opened 2019-02-19
JBPM-8254 - jBPM XES Exporter tool

cristianonicolai:
Depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/928

### DIFF
--- a/jbpm-xes/pom.xml
+++ b/jbpm-xes/pom.xml
@@ -1,0 +1,162 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright 2019 Red Hat, Inc. and/or its affiliates.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~      http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jbpm</groupId>
+    <artifactId>jbpm</artifactId>
+    <version>7.19.0-SNAPSHOT</version>
+  </parent>
+
+  <name>jBPM :: XES Exporter</name>
+  <description>jBPM XES Exporter</description>
+  <artifactId>jbpm-xes</artifactId>
+
+  <dependencies>
+
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie</groupId>
+      <artifactId>kie-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.soup</groupId>
+      <artifactId>kie-soup-dataset-sql</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.soup</groupId>
+      <artifactId>kie-soup-dataset-core</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.soup</groupId>
+      <artifactId>kie-soup-dataset-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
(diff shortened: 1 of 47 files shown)


jbpm-xes/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+
+    <dependency>
+      <groupId>org.jbpm</groupId>
+      <artifactId>jbpm-test</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.server</groupId>
+      <artifactId>kie-server-api</artifactId>
```
mswiderski: we cannot refer to kie-server-api from jbpm repo

jbpm-xes/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+  ~ limitations under the License.
+  -->
+
+<project xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.jbpm</groupId>
+    <artifactId>jbpm</artifactId>
+    <version>7.18.0-SNAPSHOT</version>
```
mswiderski: I guess we are already 7.19....

jbpm-xes/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+    </dependency>
+
+    <dependency>
+      <groupId>org.kie.soup</groupId>
+      <artifactId>kie-soup-dataset-api</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-classic</artifactId>
```
mswiderski: should be test scope

jbpm-xes/pom.xml
```diff
@@ -0,0 +1,157 @@
...
+
+    <dependency>
+      <groupId>org.mariadb.jdbc</groupId>
+      <artifactId>mariadb-java-client</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>org.postgresql</groupId>
+      <artifactId>postgresql</artifactId>
+    </dependency>
```
mswiderski: mark the drivers as either optional or move them into profile or remove completely with bit of docs

jbpm-xes/pom.xml
cristianonicolai: that was only because of the marshalling of query definitions, don't think is necessary so replaced with using Jackson ObjectMapper.

jbpm-xes/pom.xml
cristianonicolai: Fixed

jbpm-xes/pom.xml
cristianonicolai: That is actually used when running the CLI tool.

jbpm-xes/pom.xml
cristianonicolai: @mswiderski idea was to bundle most frequent used drivers to make easier for users when running the CLI tool. Not sure if making it optional or not including it could be an issue for people to get started.

jbpm-xes/pom.xml
```diff
@@ -0,0 +1,167 @@
...
+
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+
+    <dependency>
+      <groupId>simple-jndi</groupId>
+      <artifactId>simple-jndi</artifactId>
```
mswiderski: I'd opt at removing JNDI completely ... no need to make use of it here. Just push the data source directly to data service

jbpm-xes/pom.xml
mswiderski: fair enough

jbpm-xes/pom.xml
mswiderski: ok

jbpm-xes/src/main/java/org/jbpm/xes/XESExportMain.java
```diff
@@ -0,0 +1,136 @@
...
+
+public class XESExportMain {
+
+    private static final String DS_NAME = "jdbc/xesDs";
+
+    public static void main(String[] args) {
+        // create the Options
+        Options options = getOptions();
+
+        System.setProperty(Context.INITIAL_CONTEXT_FACTORY,
```
mswiderski: skip JNDI and supply data source directly

jbpm-xes/src/main/java/org/jbpm/xes/XESExportMain.java
```diff
@@ -0,0 +1,136 @@
...
+            ds.setUsername(line.getOptionValue("user"));
+            ds.setPassword(line.getOptionValue("password"));
+            ds.setDefaultReadOnly(true);
+//          Test connection
+            ds.getConnection().close();
+
+            context = new InitialContext();
+            context.bind(DS_NAME,
+                         ds);
+            DataSetService dataSetService = new DataSetServiceImpl(dsName -> DS_NAME);
```
mswiderski: looks like you're already setting data source so what's the point for JNDI bind?

jbpm-xes/src/main/java/org/jbpm/xes/XESExportMain.java
cristianonicolai: @mswiderski yeah, good catch, I removed the JNDI completely now.

cristianonicolai:
@mswiderski updated.

PR #1449, opened 2019-02-19
[JBPM-8097] jBPMHelper no longer sets default values for H2

elguardian:
set proper default values in jBPM.properties

### DIFF
--- a/jbpm-examples/src/main/java/org/jbpm/examples/checklist/impl/ChecklistExample.java
+++ b/jbpm-examples/src/main/java/org/jbpm/examples/checklist/impl/ChecklistExample.java
@@ -21,135 +21,134 @@ import java.util.List;
 
 import org.jbpm.examples.checklist.ChecklistItem;
 import org.jbpm.examples.checklist.ChecklistManager;
-import org.jbpm.test.JBPMHelper;
 import org.kie.api.KieServices;
 import org.kie.api.io.ResourceType;
 import org.kie.api.runtime.manager.RuntimeEnvironment;
 import org.kie.api.runtime.manager.RuntimeEnvironmentBuilder;
 import org.kie.api.task.UserGroupCallback;
+import org.kie.test.util.db.PersistenceUtil;
 
 public class ChecklistExample {
-	
-	public static void main(String[] args) {
-		try {
-			
-			JBPMHelper.startH2Server();
-			JBPMHelper.setupDataSource();
-			RuntimeEnvironment environment = RuntimeEnvironmentBuilder.Factory.get().newDefaultBuilder()
-	            .userGroupCallback(new UserGroupCallback() {
-	    			public List<String> getGroupsForUser(String userId) {
-	    				List<String> result = new ArrayList<String>();
-	    				if ("actor4".equals(userId)) {
-	    					result.add("group1");
-	    				}
-	    				return result;
-	    			}
-	    			public boolean existsUser(String arg0) {
-	    				return true;
-	    			}
-	    			public boolean existsGroup(String arg0) {
-	    				return true;
-	    			}
-	    		})
-	            .addAsset(KieServices.Factory.get().getResources().newClassPathResource("checklist/SampleChecklistProcess.bpmn"), ResourceType.BPMN2)
-	            .get();
-			ChecklistManager checklistManager = new DefaultChecklistManager(environment);
-	
-			long c1 = checklistManager.createContext("org.jbpm.examples.checklist.sample1", "actor1");
-			List<ChecklistItem> items = checklistManager.getTasks(c1, null);
-			printChecklistItems(items, c1);
-			
-			System.out.println("Completing Task1");
-			ChecklistItem item1 = findChecklistItem(items, "Task1");
-			checklistManager.completeTask("actor1", item1.getTaskId());
-			items = checklistManager.getTasks(c1, null);
-			printChecklistItems(items, c1);
-			
-			System.out.println("Adding Extra Task");
-			String[] actorIds = new String[] {
-				"actor5"
-			};
-			ChecklistItem itemExtra = checklistManager.addTask("actor5", actorIds, new String[0], "TaskExtra", "2+", c1);
-			items = checklistManager.getTasks(c1, null);
-			printChecklistItems(items, c1);
-			
-			System.out.println("Completing Task2");
-			ChecklistItem item2 = findChecklistItem(items, "Task2");
-			checklistManager.claimTask("actor4", item2.getTaskId());
-			checklistManager.completeTask("actor4", item2.getTaskId());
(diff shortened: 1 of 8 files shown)


MarianMacik:
It seems that JBPMHelper class is deprecated since 6.0:

https://github.com/kiegroup/jbpm/blob/a0db05af65bc88c0cc9053a6d12f1d2e993b0616/jbpm-test/src/main/java/org/jbpm/test/JBPMHelper.java#L45-L49

Wouldn't it be better to replace it with direct calls of PersistenceUtil as we do in other tests?

elguardian:
@MarianMacik up to you. don't really mind either solution is good to me. this only affects jbpm-examples AFAIK.

elguardian:
depends on https://github.com/kiegroup/drools/pull/2264/files

jbpm-examples/src/main/java/org/jbpm/examples/multipleinstance/MultipleInstanceExample.java
```diff
@@ -32,6 +31,7 @@
 import org.kie.api.runtime.manager.RuntimeManagerFactory;
 import org.kie.api.task.TaskService;
 import org.kie.api.task.model.TaskSummary;
+import org.kie.test.util.db.PersistenceUtil;
```
MarianMacik: There is also PersistenceUtil class in jbpm, though you would probably have to declare dependency:

    <dependency>
      <groupId>org.jbpm</groupId>
      <artifactId>jbpm-persistence-jpa</artifactId>
      <type>test-jar</type>
      <scope>test</scope>
    </dependency>

So it is up to you.

elguardian: wouldn't that be spreading the logic ? I think it makes more sense to use test utils

MarianMacik: we already use this org.jbpm.persistence.util.PersistenceUtil in all jbpm unit tests. But given that jbpm-examples are de facto not tests, maybe it is better to stay with org.kie.test.util.db.PersistenceUtil.

jbpm-test/src/main/resources/jBPM.properties
```diff
@@ -1,9 +1,9 @@
...
-persistence.datasource.name=jdbc/jbpm-ds
-persistence.datasource.user=sa
-persistence.datasource.password=
-persistence.datasource.url=jdbc:h2:tcp://localhost/~/jbpm-db
-persistence.datasource.driverClassName=org.h2.Driver
+name=jdbc/jbpm-ds
+driverClassName=org.h2.Driver
+user=sa
+password=
+url=jdbc:h2:tcp://localhost/~/jbpm-db
```
MarianMacik: Is JBPMHelper usable with these properties (although I know it is deprecated)? It still searches for properties with `persistence.datasource` prefix. Maybe we don't need to change this file at all.

jbpm-examples/src/main/java/org/jbpm/examples/checklist/impl/ChecklistExample.java
```diff
@@ -21,20 +21,19 @@
...
 
 public class ChecklistExample {
 	
 	public static void main(String[] args) {
 		try {
-			
-			JBPMHelper.startH2Server();
-			JBPMHelper.setupDataSource();
+            PersistenceUtil.setupPoolingDataSource();
+
```
MarianMacik: Can you fix the formatting here, please?

jbpm-test/src/main/resources/jBPM.properties
karlnicholas: I didn't see that JBPMHelper was searching for properties with `persistence.datasource` prefix. I was about to submit a PR for simply these changes but I see this PR has made more changes. Hopefully the requestor will fix the formatting and this PR can be merged.

elguardian: @MarianMacik change from tabs to spaces in ChecklistExample.java otherwise it was imposible to put everything align.
done

karlnicholas: The problem is that eclipse (or devstudio) displays tab's as 4 spaces wide and github displays them as 8 spaces wide. I typically change DevStudio (or eclipse) to match that format so that it looks right on github or the rest of the world. Of course, if someone brings the code back into eclipse with the original settings it's messed up for them. The best solution I've found is set the tabs to 8 spaces but tell eclipse (or DevStudio) to always use spaces in the source.

elguardian:
@MarianMacik dependency was merged. any progress with this part ?

jbpm-test/src/main/resources/jBPM.properties
MarianMacik: I still think these should be left as before, since JBPMHelper still uses format with `persistence.datasource` prefix. See:

https://github.com/kiegroup/jbpm/blob/5dfd109a778a9cde753302d527418655556154e9/jbpm-test/src/main/java/org/jbpm/test/JBPMHelper.java#L97-L102

Can we leave it as it was before? And after that in future we can remove JBPMHelper completely?

elguardian: ok. feel free to file a jira regarding removing the JBPMHelper...

elguardian:
this is ok to test

MarianMacik:
Thanks @elguardian. It's there: [JBPM-8380](https://issues.jboss.org/browse/JBPM-8380). Fix Version is FUTURE, that doesn't necessarily mean it will be excluded only in the next major version.

elguardian:
@MarianMacik @mswiderski is there anything else stopping this for getting merge ?

mswiderski:
@elguardian @MarianMacik how about this work https://github.com/kiegroup/jbpm/pull/1476? 

would it be better option?

elguardian:
@mswiderski he created a new class SetupExamplesDatasource. which is basically PersistenceUtil, so throwing new logic to the table. I don't really mind either way.
In any case that is the wrong jira JBPM-8380 is for removing completely the JBPMHelper... that includes stuff in integration project as well.

elguardian:
@MarianMacik any thoughts ^^

MarianMacik:
I think we should stick with this one first as another way of creating a data source in our code base does bring more complexity. I would definitely use PersistenceUtil, since this is now wide-spread both here and in drools.

mswiderski:
fair enough, could you please finalize the work around it and merge?

MarianMacik:
Well, my initial thought was to merge this and then remove jBPMHelper class in a separate PR. But now there is #1477. How do we want to proceed @mswiderski @elguardian?

elguardian:
let's close this one.

PR #1452, opened 2019-02-28
JBPM-8249 - Rules activation not canceled in Case Management subprocess

mswiderski:


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -706,6 +719,179 @@ public void testTriggerTaskAndMilestoneInCase() {
...
+            assertNotNull(milestones);
+            assertEquals(0, milestones.size());
+
+            milestones = caseRuntimeDataService.getCaseInstanceMilestones(HR_CASE_ID, false, new QueryContext());
+            assertNotNull(milestones);
+            assertEquals(2, milestones.size());
+
+            List<String> expectedMilestones = Arrays.asList("Milestone1", "Milestone2");
+            for (CaseMilestoneInstance mi : milestones) {
+                assertTrue("Expected mile stopne not found", expectedMilestones.contains(mi.getName()));
```
MarianMacik: Typo in message.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -706,6 +719,179 @@ public void testTriggerTaskAndMilestoneInCase() {
...
+            // trigger milestone node
+            caseService.triggerAdHocFragment(HR_CASE_ID, "Milestone1", null);
+
+            milestones = caseRuntimeDataService.getCaseInstanceMilestones(HR_CASE_ID, true, new QueryContext());
+            assertNotNull(milestones);
+            assertEquals(1, milestones.size());
+            CaseMilestoneInstance msInstance = milestones.iterator().next();
+            assertNotNull(msInstance);
+            assertEquals("Milestone1", msInstance.getName());
+            assertEquals(true, msInstance.isAchieved());
```
MarianMacik: Can be replaced with assertTrue.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -104,7 +104,7 @@ public void matchCreated(MatchCreatedEvent event) {
             // trigger node instances of that milestone node
             String ruleName = event.getMatch().getRule().getName();
             String milestoneName = "RuleFlow-Milestone-" + getProcessInstance().getProcessId() + "-" + getMilestoneNode().getUniqueId();
-            if (milestoneName.equals(ruleName) && checkProcessInstance((Activation) event.getMatch()) && checkDeclarationMatch(event.getMatch(), getMilestoneNode().getMatchVariable())) {
+            if (((Activation) event.getMatch()).isActive() &&  milestoneName.equals(ruleName) && checkProcessInstance((Activation) event.getMatch()) && checkDeclarationMatch(event.getMatch(), getMilestoneNode().getMatchVariable())) {
```
MarianMacik: Can you explain briefly what was going on in that case with a subprocess from JIRA? I mean when there was a match for the rule at the same time for 2 objects, and the second match should be cancelled because we need only one (we configured it in that way), does it mean that drools will fire a MatchCreatedEvent also for the second rule, but this time with the match set as not active?

mswiderski: that's exactly my understanding as well so you actually explained it yourself :)

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
mswiderski: fixed, there were actually two of those

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
mswiderski: changed

mswiderski:
Jenkins retest this

mswiderski:
Jenkins retest this

PR #1453, opened 2019-03-01
jBPM XES Exporter jdk11 support

gmunozfe:
JAXB dependencies are needed for jdk11, as JAXB APIs are considered to be Java EE APIs, and therefore, have been completely removed from the jdk11.

### DIFF
--- a/jbpm-xes/pom.xml
+++ b/jbpm-xes/pom.xml
@@ -34,6 +34,12 @@
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-test</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>javax.activation</groupId>
+          <artifactId>javax.activation-api</artifactId>
+        </exclusion>
+      </exclusions>
     </dependency>
 
     <dependency>
@@ -126,6 +132,27 @@
     <dependency>
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
+    </dependency>
+    
+    <!-- needed for jdk 11 -->
+    <dependency>
+      <groupId>org.jboss.spec.javax.xml.bind</groupId>
+      <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+    </dependency>
+
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+    </dependency>
+    
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-core</artifactId>
+    </dependency>
+    
+    <dependency>
+      <groupId>javax.activation</groupId>
+      <artifactId>activation</artifactId>
     </dependency>
 
   </dependencies>


jbpm-xes/pom.xml
```diff
@@ -127,6 +127,24 @@
...
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+      <scope>test</scope>
+    </dependency>
+    
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-core</artifactId>
+      <scope>test</scope>
+    </dependency>
```
MarianMacik: I think we can leave out the scope entirely (which means compile scope), since no other components depend on this one and if somebody wants to use the tool on JDK 11, they will need all these dependencies anyway.

gmunozfe: Good point, agreed. Also added a comment before for highlighting that this is needed because of jdk11

jbpm-xes/pom.xml
```diff
@@ -127,6 +127,23 @@
       <groupId>org.postgresql</groupId>
       <artifactId>postgresql</artifactId>
     </dependency>
+    
+    <!-- needed for jdk 11 -->
+    <dependency>
+      <groupId>org.jboss.spec.javax.xml.bind</groupId>
+      <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+      <scope>provided</scope>
```
MarianMacik: There is still scope provided.

gmunozfe:
@cristianonicolai 

When testing this tool, I had some issues:

1.- Excluding "commons-logging" provokes this exception at command line invocation (also for jdk 8):

_Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/commons/logging/LogFactory
	at org.apache.commons.dbcp2.BasicDataSource.<clinit>(BasicDataSource.java:73)
	at org.jbpm.xes.XESExportMain.main(XESExportMain.java:36)_

_Caused by: java.lang.ClassNotFoundException: org.apache.commons.logging.LogFactory
	at java.net.URLClassLoader.findClass(URLClassLoader.java:382)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:349)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	... 2 more_

Not sure about the main reason of this exclusion (enforcer plugin pushes to do it) but it's needed in the classpath for working (BasicDataSource depends on it).

2.- Status filtering is expecting 4 arguments always
https://github.com/kiegroup/jbpm/blob/b223713a01dcf676cd244a5a4f9a9c6d49b06629/jbpm-xes/src/main/java/org/jbpm/xes/XESExportMain.java#L107

therefore, only when status is the last parameter works as expected. If there are more options after "status", it takes them as own arguments.

_java -jar .....   -file jbpm.xes -status 0 1   -> works
java -jar .... -status 0 1  - file jbpm.xes  -> doesn't work, console output:_

_Invalid process status number for input: -file, valid status are number between 0 and 4.
Invalid process status number for input: jbpm.xes, valid status are number between 0 and 4_

My proposal is to use unlimited number of arguments for status, but comma-separated.

`options.addOption(Option.builder("status").argName("status").hasArgs().valueSeparator(',').desc("Process status to export").build());
`

But when testing it, I found that due to this bug at commons-cli-1.3.x:
 https://issues.apache.org/jira/browse/CLI-265
then, we should move to commons-cli-1.4.

Adding commons-cli-1.4 in the classpath, it works something like this:
_java -jar .... -status 0,1  - file jbpm.xes_

Perhaps, if you agree, I can write some unit tests for covering also command line execution with filtering, invoking directly to the _XESExportMain.main_

What do you think?

3.- When testing against RHPAM running with EAP 7.2, I found that h2 versions are clashing, and following exception is thrown:

_Caused by: org.h2.jdbc.JdbcSQLException: Connection is broken: "unexpected status 16777216" [90067-173]_

JBoss EAP 7.2 is using h2-1.4.193.redhat-2 vs. h2-1.3.173 in XES effective pom.

When both versions are aligned, it works fine.

jbpm-xes/pom.xml
gmunozfe: Done

cristianonicolai:
@gmunozfe thanks for taking a look, overall I think your proposals are good, let me try to address each one individually.

1. Looks fine to me to skip the banned dependency as you suggested.
2. Sounds like a good idea, moving to a newer commons-cli to support a list of statuses as well as adding new tests for it. I would just perhaps suggest doing that on a separated PR.
3. I'm fine to upgrade, if that also works with older versions. The only issue might be that this is defined in the jboss-bom which might affect other places. If that is not possible we could look into overriding _version.com.h2database_ only for this pom.

jbpm-xes/pom.xml
```diff
@@ -127,6 +133,27 @@
...
+
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-impl</artifactId>
+    </dependency>
+    
+    <dependency>
+      <groupId>com.sun.xml.bind</groupId>
+      <artifactId>jaxb-core</artifactId>
+    </dependency>
```
gmunozfe: javax.activation is needed for jdk11 command line execution, but activation-api must be excluded from jbpm-test to avoid duplicated classes

cristianonicolai:
@gmunozfe do you still need to merge this? Seems like you incorporated changes into https://github.com/kiegroup/jbpm/pull/1459

gmunozfe:
Right, #1459 contains all changes of this PR, no need to merge when pulling that (only comments may be useful to keep)

cristianonicolai:
@gmunozfe ok, I guess you can close this one then.

gmunozfe:
Closed, as #1459 contains also these changes

PR #1457, opened 2019-03-08
jbpm-remote-ejb: Disable deployment sync

MarianMacik:
This should help avoid false negatives in duplicate deploy/undeploy tests.


MarianMacik:
I deleted in on purpose since the cargo-maven2-plugin is not configured at all in the parent.

gmunozfe:
OK, then looks perfectly fine for me

PR #1459, opened 2019-03-12
jBPM XES Exporter: command line tests and upgrade to latest schema (BAPL-1214)

gmunozfe:
1. Added tests for XES Exporter command line

2. Schema updated to the latest at xes-standard.org: xes-ieee-1849-2016.xsd:
2.1- ClassifierType name is NCName type, so whitespaces like at "By Node Name", etc are not allowed and replaced by "ByNodeName", etc.
2.2- scope attribute is required at Classifier (set up to "event", default value).

3. Status option set up to unlimited number of arguments and comma-separated

4. Updated versions of following 3rd parties at pom:
- commons-cli-1.4 (because of bug https://issues.apache.org/jira/browse/CLI-265)
- h2-1.4.193 (to be aligned with JBoss EAP 7.2)


gmunozfe:
@MarianMacik @cristianonicolai  you can review it, thank you!

jbpm-xes/pom.xml
```diff
@@ -27,13 +27,29 @@
   <name>jBPM :: XES Exporter</name>
   <description>jBPM XES Exporter</description>
   <artifactId>jbpm-xes</artifactId>
+  
+  <properties>
+    <version.commons-cli>1.4</version.commons-cli>
+    <version.com.h2database>1.4.193</version.com.h2database>
```
MarianMacik: Don't we want to upgrade this globally? H2 we use right now is quite old.

jbpm-xes/pom.xml
```diff
@@ -27,13 +27,29 @@
...
   <dependencies>
 
     <dependency>
       <groupId>org.jbpm</groupId>
       <artifactId>jbpm-test</artifactId>
       <scope>test</scope>
+      <exclusions>
+        <exclusion>
+          <groupId>org.slf4j</groupId>
+          <artifactId>jcl-over-slf4j</artifactId>
```
MarianMacik: Just curious, why is this exclusion needed?

jbpm-xes/pom.xml
```diff
@@ -127,10 +137,57 @@
...
+              <goals>
+                <goal>enforce</goal>
+              </goals>
+              <configuration>
+                <rules>
+                  <bannedDependencies>
+                    <includes>
+                      <include>commons-logging:*</include>
+                    </includes>
+                  </bannedDependencies>
```
MarianMacik: Why was this banned?

jbpm-xes/pom.xml
gmunozfe: It is needed to avoid duplicated classes with commons-logging:

  Found in:
    commons-logging:commons-logging:jar:1.1.1:compile
    org.slf4j:jcl-over-slf4j:jar:1.7.25:test
  Duplicate classes:
    org/apache/commons/logging/impl/SimpleLog$1.class
    org/apache/commons/logging/Log.class
    org/apache/commons/logging/LogConfigurationException.class
    org/apache/commons/logging/impl/NoOpLog.class
    org/apache/commons/logging/impl/SimpleLog.class
    org/apache/commons/logging/LogFactory.class

jbpm-xes/pom.xml
gmunozfe: It's banned at kie-parent pom:
 
         
                      <exclude>commons-logging:commons-log*</exclude>

As it's needed for working (BasicDataSource depends on org.apache.commons.logging.LogFactory), Cristiano and me agreed to override the enforcer plugin at pluginManagement for skipping this banned dependency.

jbpm-xes/pom.xml
gmunozfe: upgrading h2 and commons-cli, perhaps can be done at kie-parent, instead of here, but Cristiano was concerned about the implications in other modules

gmunozfe:
As Marian intelligently suggests, it's convenient to switch commons-dbcp2 artifact to tomcat-dbcp, and consequently we can get rid off of enforcer plugin (ban dependency issue is not longer present) as well as jcl-over-slf4j exclusion from jbpm-test. Thanks a lot @MarianMacik !

gmunozfe:
jenkins retest this

PR #1460, opened 2019-03-13
[JBPM-8342] Newly added system property - org.jbpm.correlationkey.len gth - does not truncate the "value" column in the CorrelationPropertyInfo table

elguardian:
added CORRELATION_KEY_LOG_LENGTH to CorrelationPropertyInfo

PR #1463, opened 2019-03-20
[JBPM-8342] Newly added system property - org.jbpm.correlationkey.length - does not truncate the "value" column in the CorrelationPropertyInfo table

elguardian:
added CORRELATION_KEY_LOG_LENGTH to CorrelationPropertyInfo

cherry pick from (#1460)


elguardian:
@mareknovotny this is also for 7.3.1 need tag and merge (all acks)

PR #1464, opened 2019-03-21
JBPM-8116: It is not possible to see SLA violations in business centr…

nmirasch:
…al UI
Added sla data to UsertTaskInstanceDesc when getTaskInstanceById.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/mapper/UserTaskInstanceQueryMapper.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/mapper/UserTaskInstanceQueryMapper.java
@@ -77,7 +77,10 @@ public class UserTaskInstanceQueryMapper extends AbstractQueryMapper<UserTaskIns
                 getColumnStringValue(dataSetResult, COLUMN_TASK_PROCESSID, index),//processId,
                 getColumnLongValue(dataSetResult, COLUMN_TASK_PROCESSINSTANCEID, index),//processInstanceId,
                 getColumnDateValue(dataSetResult, COLUMN_CREATEDON, index),//createdOn,
-                getColumnDateValue(dataSetResult, COLUMN_DUEDATE, index)//dueDate
+                getColumnDateValue(dataSetResult, COLUMN_DUEDATE, index),//dueDate
+                getColumnLongValue(dataSetResult, COLUMN_WORKITEMID, index),//workItemId
+                getColumnDateValue(dataSetResult, COLUMN_TASK_SLA_DUE_DATE, index),//slaDueDate
+                getColumnIntValue(dataSetResult, COLUMN_TASK_SLA_COMPLIANCE, index)//slaCompliance
                 );
         return userTask;
     }
--- a/jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
+++ b/jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
@@ -714,12 +714,17 @@
 	      a.processId,
 	      a.processInstanceId,
 	      a.createdOn,
-	      a.dueDate
+	      a.dueDate,
+    	      a.workItemId,
+    	      log.slaDueDate,
+    	      log.slaCompliance
 	      )
         from 
-        	AuditTaskImpl a 
+        	AuditTaskImpl a,
+            	NodeInstanceLog log
         where 
-        	a.taskId = :taskId 
+        	a.taskId = :taskId
+            	and log.workItemId = a.workItemId
     </query>
     <!-- hint name="org.hibernate.timeout" value="200"/ -->
   </named-query>
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-impl/src/test/java/org/jbpm/services/ejb/test/RuntimeDataServiceEJBIntegrationTest.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-impl/src/test/java/org/jbpm/services/ejb/test/RuntimeDataServiceEJBIntegrationTest.java
@@ -817,6 +817,9 @@ public class RuntimeDataServiceEJBIntegrationTest extends AbstractTestSupport {
     	assertEquals(processInstanceId, userTask.getProcessInstanceId());
     	assertEquals("Write a Document", userTask.getName());
 
+		NodeInstanceDesc nodeInstanceDesc = runtimeDataService.getNodeInstanceForWorkItem(userTask.getWorkItemId());
+		assertEquals(nodeInstanceDesc.getSlaCompliance(), userTask.getSlaCompliance());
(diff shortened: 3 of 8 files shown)


nmirasch:
@mswiderski @cristianonicolai Retrieve sla compliance and duedate at getTaskInstanceById and fill the UserTaskInstanceDesc to avoid to do a second call to retrieve the task node instance related,I think make sense maintain this together. Please could you take a look on this? Thanks!

cristianonicolai:
jenkins please retest this

jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
```diff
@@ -714,12 +714,17 @@
 	      a.processId,
 	      a.processInstanceId,
 	      a.createdOn,
-	      a.dueDate
+	      a.dueDate,
+        a.workItemId,
```
cristianonicolai: small formatting off here

nmirasch: @cristianonicolai updated

jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
```diff
@@ -714,12 +714,17 @@
...
-	      a.dueDate
+	      a.dueDate,
+    	      a.workItemId,
+    	      log.slaDueDate,
+    	      log.slaCompliance
 	      )
         from 
-        	AuditTaskImpl a 
+        	AuditTaskImpl a,
+            	NodeInstanceLog log
```
mswiderski: Not sure I like this join... would it be possible to store that info in AuditTask table to avoid join? Just asking ... it might not be that easy ..,

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-impl/src/test/java/org/jbpm/services/ejb/test/RuntimeDataServiceEJBIntegrationTest.java
```diff
@@ -817,6 +817,9 @@ public void testGetTaskById() {
     	assertEquals(processInstanceId, userTask.getProcessInstanceId());
     	assertEquals("Write a Document", userTask.getName());
 
+		NodeInstanceDesc nodeInstanceDesc = runtimeDataService.getNodeInstanceForWorkItem(userTask.getWorkItemId());
+		assertEquals(nodeInstanceDesc.getSlaCompliance(), userTask.getSlaCompliance());
+		assertEquals(nodeInstanceDesc.getSlaDueDate(), userTask.getSlaDueDate());
```
mswiderski: Looks like we change UserTaskInstance class but we test nodeInstance ... shouldn’t we test usertask?

nmirasch: @mswiderski here we rely here in  runtimeDataService.getNodeInstanceForWorkItem(userTask.getWorkItemId()) returns the good values for sla info. This test try to assert that the getTaskById filled new sla values are equals that the ones returned at nodeInstanceDesc

jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
nmirasch: @mswiderski for sure it would be better avoiding this join an it would be easier the console dataset query definition. Then, the proposal would be add this new two column to audiTask table and look for nodeInstanceDesc sla updates calls and propagate them to the auditTask too. Is that the idea?

nmirasch: @mswiderski sorry I mean  nodeInstanceLog updates instead nodeInstanceDesc

mswiderski: yes, that was a thought, do you think that would be doable?

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-impl/src/test/java/org/jbpm/services/ejb/test/RuntimeDataServiceEJBIntegrationTest.java
mswiderski: ah ok, now I see, thanks

jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
nmirasch: @mswiderski let me give it a try. Thanks!

mswiderski: @nmirasch after looking it at more closely it will be rather difficult to make this without the join. Main reason is that SLA is tracked on process engine level (timers) and updates are only received within the process instance/node instance. So attempting to update the task tables will not really be an option without extra listeners etc. So please go ahead with the approach used in this PR, thanks!

nmirasch: @mswiderski ook, Thanks!

nmirasch:
rebased

nmirasch:
@MarianMacik @jstastny-cz rebased and updated revapi-config.json

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/mapper/UserTaskInstanceQueryMapper.java
```diff
@@ -77,7 +77,10 @@ protected UserTaskInstanceDesc buildInstance(DataSet dataSetResult, int index) {
                 getColumnStringValue(dataSetResult, COLUMN_TASK_PROCESSID, index),//processId,
                 getColumnLongValue(dataSetResult, COLUMN_TASK_PROCESSINSTANCEID, index),//processInstanceId,
                 getColumnDateValue(dataSetResult, COLUMN_CREATEDON, index),//createdOn,
-                getColumnDateValue(dataSetResult, COLUMN_DUEDATE, index)//dueDate
+                getColumnDateValue(dataSetResult, COLUMN_DUEDATE, index),//dueDate
+                getColumnLongValue(dataSetResult, COLUMN_WORKITEMID, index),//workItemId
+                getColumnDateValue(dataSetResult, COLUMN_TASK_SLA_DUE_DATE, index),//slaCompliance
```
jstastny-cz: I believe the comments for this line and the one below are swapped.

nmirasch: @jstastny-cz yes, updated!

PR #1465, opened 2019-03-21
[JBPM-8364] AsyncMode True causes error with Multi Instance Sub Process

elguardian:
unwrapping async node instances when foreach loop is created.

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/SubProcessHandler.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/SubProcessHandler.java
@@ -28,11 +28,11 @@ import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.workflow.core.Node;
 import org.jbpm.workflow.core.NodeContainer;
 import org.jbpm.workflow.core.impl.NodeImpl;
-import org.jbpm.workflow.core.node.ActionNode;
 import org.jbpm.workflow.core.node.CompositeContextNode;
 import org.jbpm.workflow.core.node.EventSubProcessNode;
 import org.jbpm.workflow.core.node.ForEachNode;
 import org.jbpm.workflow.core.node.StartNode;
+import org.jbpm.workflow.instance.NodeInstance;
 import org.kie.api.definition.process.Connection;
 import org.w3c.dom.Element;
 import org.xml.sax.Attributes;
@@ -70,7 +70,8 @@ public class SubProcessHandler extends AbstractNodeHandler {
             final ExtensibleXmlParser parser) throws SAXException {
 		final Element element = parser.endElementBuilder();
 		Node node = (Node) parser.getCurrent();	
-		
+
+
 		// determine type of event definition, so the correct type of node can be generated
 		boolean found = false;		
 		org.w3c.dom.Node xmlNode = element.getFirstChild();
@@ -80,6 +81,8 @@ public class SubProcessHandler extends AbstractNodeHandler {
 			    Boolean isAsync = Boolean.parseBoolean((String)node.getMetaData().get("customAsync"));
 				// create new timerNode
 				ForEachNode forEachNode = new ForEachNode();
+                // indirect to the real node if it is an async event node
+                forEachNode.getCompositeNode().getMetaData().put(NodeInstance.METADATA_WRAP_ASYNC_NODE, false);
 				forEachNode.setId(node.getId());
 				forEachNode.setName(node.getName());
 				forEachNode.setSequential(Boolean.parseBoolean(((Element) xmlNode).getAttribute("isSequential")));
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
@@ -24,11 +24,11 @@ import org.jbpm.workflow.core.impl.NodeImpl;
 import org.jbpm.workflow.core.node.AsyncEventNode;
 import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
 import org.kie.api.definition.process.Node;
+import org.kie.api.executor.CommandContext;
+import org.kie.api.executor.ExecutorService;
 import org.kie.api.runtime.manager.RuntimeManager;
 import org.kie.api.runtime.process.EventListener;
 import org.kie.api.runtime.process.NodeInstance;
-import org.kie.api.executor.CommandContext;
-import org.kie.api.executor.ExecutorService;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -91,13 +91,21 @@ public class AsyncEventNodeInstance extends EventNodeInstance {
     }
 
     @Override
+    protected NodeInstance getFrom() {
+        AsyncEventNode node = (AsyncEventNode) getNode();
+        org.jbpm.workflow.instance.NodeInstance instance = ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).getNodeInstance(node.getPreviousNode());
+        ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).removeNodeInstance(instance);
+        return instance;
+    }
+
+    @Override
     public void triggerCompleted() {
         getProcessInstance().removeEventListener(getEventType(), getEventListener(), true);
         ((org.jbpm.workflow.instance.NodeInstanceContainer)getNodeInstanceContainer()).setCurrentLevel(getLevel());
         ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).removeNodeInstance(this);
         
         NodeInstance instance = ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).getNodeInstance(getNode());
-        
+
         triggerNodeInstance((org.jbpm.workflow.instance.NodeInstance) instance, NodeImpl.CONNECTION_DEFAULT_TYPE);
     }
 
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
@@ -16,8 +16,6 @@
 
 package org.jbpm.workflow.instance.node;
 
-import static org.jbpm.workflow.instance.impl.DummyEventListener.EMPTY_EVENT_LISTENER;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -47,6 +45,8 @@ import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
 import org.kie.api.definition.process.Connection;
 import org.kie.api.definition.process.Node;
 import org.kie.api.definition.process.NodeContainer;
+
+import static org.jbpm.workflow.instance.impl.DummyEventListener.EMPTY_EVENT_LISTENER;
 
 /**
  * Runtime counterpart of a composite node.
@@ -122,9 +122,10 @@ public class CompositeNodeInstance extends StateBasedNodeInstance implements Nod
 	        List<Connection> connections = nodeAndType.getNode().getIncomingConnections(nodeAndType.getType());
 	        for (Iterator<Connection> iterator = connections.iterator(); iterator.hasNext(); ) {
 	            Connection connection = iterator.next();
+
 	            if ((connection.getFrom() instanceof CompositeNode.CompositeNodeStart) &&
-	            		(from == null ||
-	    				((CompositeNode.CompositeNodeStart) connection.getFrom()).getInNode().getId() == from.getNodeId())) {
+                    (from == null ||
+                     ((CompositeNode.CompositeNodeStart) connection.getFrom()).getInNode().getId() == from.getNodeId())) {
 	                NodeInstance nodeInstance = getNodeInstance(connection.getFrom());
 	                ((org.jbpm.workflow.instance.NodeInstance) nodeInstance).trigger(null, nodeAndType.getType());
 	                return;
@@ -243,6 +244,7 @@ public class CompositeNodeInstance extends StateBasedNodeInstance implements Nod
         return null;
     }
 
+    @Override
     public NodeInstance getNodeInstance(final Node node) {
         // TODO do this cleaner for start / end of composite?
         if (node instanceof CompositeNode.CompositeNodeStart) {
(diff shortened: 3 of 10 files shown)


elguardian:
@mswiderski give me feedback. looks a bit hacky for my taste but I didn't see how to unwrap the async node in any other way.

https://issues.jboss.org/browse/RHPAM-1968

pefernan:
@elguardian I think you're using a wrong JIRA number on the PR title, it should be  JBPM-8364

mswiderski:
would you mind adding a test case? that way I could give it a quick try to see how it behaves

elguardian:
@mswiderski added test case

elguardian:
this is ok to test

elguardian:
this is ok to test

elguardian:
retest this please

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -262,7 +271,7 @@ public NodeInstance getNodeInstance(final Node node) {
         // async continuation handling
         if (node instanceof AsyncEventNode) {
             actualNode = ((AsyncEventNode) node).getActualNode();
-        } else if (useAsync(node)) {
+        } else if (useAsync(node) && wrapAsync) {
```
mswiderski: why do we need this wrapAsync?

elguardian: that one is tricky
https://github.com/kiegroup/jbpm/pull/1465/files#diff-88f436b26f11134f1919b7fe8c0f4b38R162

if you try to get the node instance from an async node you will get an async node because of the env entry AsyncMode = true... I needed a way to avoid that.

elguardian: @mswiderski  ^^

mswiderski: so why not putting this logic into useAsync? So if it is already async node it should not wrap it or did I misunderstand it?

elguardian: that is fine by me. you are just moving some logic around.

elguardian:
done @mswiderski moved the argument to useAsync

mswiderski:
@elguardian I didn't mean to move the argument from one method to the other but to include the logic to figure out if the wrap should happen or not in that method already.

elguardian:
@mswiderski it is context dependendat (depends on the caller) so it is not possible to put that in the logic of that method. getNodeInstance is widely used.
The problem is in here
https://github.com/kiegroup/jbpm/pull/1465/files#diff-88f436b26f11134f1919b7fe8c0f4b38R162

you need to get there the node instance and not the async node instance so you need to convey to the node instance that you don't want to wrap. unless there is another way to convey info to that call (env vars can be discarded as probably will cause concurrency problems) I don't see any other way.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/ForEachNodeInstance.java
```diff
@@ -153,6 +155,12 @@ public void internalTrigger(org.kie.api.runtime.process.NodeInstance fromm, Stri
             		String variableName = getForEachNode().getVariableName();
             		NodeInstance nodeInstance = (NodeInstance)
             		((NodeInstanceContainer) getNodeInstanceContainer()).getNodeInstance(getForEachSplitNode().getTo().getTo());
+
+            		// indirect to the real node if it is an async event node
+                    if (nodeInstance.getNode() instanceof AsyncEventNode && getNodeInstanceContainer() instanceof CompositeNodeInstance) {
+                        Node node = ((AsyncEventNode) nodeInstance.getNode()).getActualNode();
+                        nodeInstance = (NodeInstance) ((CompositeNodeInstance) getNodeInstanceContainer()).getNodeInstance(node, false);
```
mswiderski: maybe an alternative approach would be to mark such nodes (entry node in foreach) as not async - like setting metadata on a node which would instruct to ignore asyncnode so it won't be wrapped. That would allow us to include that logic into the useAsync method, wdyt?

elguardian: @mswiderski  I cannot check the consequences of that. If we are talking about mutable instance another call in other place of the system could cause collateral problems. You know the code better, if those objects won't be used anywhere else then I don't have a problem implementing it that way.

mswiderski: what you're doing here is essentially disabling async node, so better to already mark such node once (at parsing time) so you don't need to do that casting and unwrapping many times. Nodes are immutable, meaning they don't change per process instance and always stay the same as they are just the definition. So if you have a foreach on a node or subprocess it will always be used in that context

elguardian: Added like this.
https://github.com/kiegroup/jbpm/pull/1465/files#diff-9c56661f380ef8af8dabbd9268c7ea1aR437

elguardian:
Jenkins retest this please

elguardian:
retest this please

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
```diff
@@ -37,7 +37,9 @@
  *
  */
 public class AsyncEventNodeInstance extends EventNodeInstance {
-    
+
+    public static final String METADATA_WRAP = "wrap";
```
mswiderski: this should be moved to node instead of node instance, and would actually keep it on generic node level rather than async

and maybe give it a bit more meaningful name

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -122,9 +122,13 @@ public void internalTrigger(final org.kie.api.runtime.process.NodeInstance from,
...
 	            Connection connection = iterator.next();
+                long fromNodeId = from != null ? from.getNodeId() : -1;
+                if (from != null && from.getNode() instanceof AsyncEventNode) {
+                    fromNodeId = ((AsyncEventNode) from.getNode()).getPreviousNode(type).getId();
+                }
 	            if ((connection.getFrom() instanceof CompositeNode.CompositeNodeStart) &&
-	            		(from == null ||
-	    				((CompositeNode.CompositeNodeStart) connection.getFrom()).getInNode().getId() == from.getNodeId())) {
+                    (from == null ||
+                     ((CompositeNode.CompositeNodeStart) connection.getFrom()).getInNode().getId() == fromNodeId)) {
```
mswiderski: if the exclude based on metadata will work, then I think all this code change won't be needed

elguardian: I cannot remove this part. the from is performed before unwrapping so this step is needed. The unwrap is not needed anymore.

elguardian:
ok to test

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
mswiderski: how about finding this logic in getFrom so it actually inspects if the from node is async and if so will use its actualNode, would that work?

elguardian: @mswiderski 
I am not sure how this should work. In this particular case when you are operating with async = false you get

ActionNode (connection.getFrom.getInNode) -> ActionNode (from node) -> ForEach (next)

with async = true you get
ActionNode (connection.getFrom.getInNode) -> AsyncNode (from node, ActualNode = forEach)

so when you are getting the from.getNodeId you are getting the foreach node and not the action node id (the previous node from the actual node) causing a mismatch and not able to go through the proper control flow. changin any call in  asyncnodeinstance (getNodeId, getNode) is not working because it causes collateral problems (unsolvable from my point of view)

changing getFrom connection won't work either as getFrom node does not hold the wrap information.

The call getFrom is working in both cases as it is receiving the proper node. If you get de NodeInstance (an async node) the workflow won't advance at that stage properly. that is the reason we need to avoid wrapping again in a AsyncNodeInstance (we would be triggering several times the async node instead of the actual node -> for each)

I don't see a way to work this same way or sink the logic anywhere in the code

mswiderski: In that case I can only say I trust your investigation and apparently you're right and this can't be moved anyway. So I think we are good to go with it. @MarianMacik can you have a look as well?

elguardian: @MarianMacik it would be great if this makes to the cut off this friday. Let's see if we can make it.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/SubProcessHandler.java
```diff
@@ -80,6 +81,8 @@ public Object end(final String uri, final String localName,
 			    Boolean isAsync = Boolean.parseBoolean((String)node.getMetaData().get("customAsync"));
 				// create new timerNode
 				ForEachNode forEachNode = new ForEachNode();
+                // indirect to the real node if it is an async event node
+                forEachNode.getCompositeNode().getMetaData().put(NodeInstance.METADATA_WRAP_ASYNC_NODE, true);
```
MarianMacik: I struggle with this explanation - if it is an asynce event node, mark wrap async node true - but in the end when you call useAsync method, it will do the negation, which means it won't wrap it. Moreover, the condition to wrap/set this property only if it is an async event node is quite misleading since it will be always done for every multiInstanceLoopCharacteristics node - just by reading the code.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
MarianMacik: So here we are matching connection.getFrom.getInNode (the node which is before CompositeStartNode) with the from node which is argument of internalTrigger method, right? If there is a match, we can trigger and continue... In case of sync it works ok since the from is a node which is before foreach node/composite node. But in case of async the from node is already the AsyncNode which wraps the foreach/composite node we are trying to trigger? In that case I think that we should properly pass the real from node (so the same node as with sync - the one before the foreach node) into the method - if it is possible. @mswiderski isn't an issue in that part - that we are passing the incorrect from node which is not the from node but rather a wrapper of the current node?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
```diff
@@ -429,6 +434,9 @@ public void setCurrentLevel(int currentLevel) {
     }
 
     protected boolean useAsync(final Node node) {
+        if (node.getMetaData().containsKey(NodeInstance.METADATA_WRAP_ASYNC_NODE)) {
+            return !((Boolean) node.getMetaData().get(NodeInstance.METADATA_WRAP_ASYNC_NODE));
+        }
```
MarianMacik: As I already pointed out - if this property is true - wrap async node - we are actually not wrapping anything.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
elguardian: @MarianMacik  Just to point out that with an async execution you are changing the bpmn adding a new node instance in between ActionNode and ForEachNode ... AsynEventNode

MarianMacik: well from what you wrote and from the code I understood that engine wraps it, not inserts it, doesn't it?

elguardian: Depends how you look at it...  the usual stack trace 
StartNode -> Action Node -> ForEach -> Action Node.... in this case... so the  from that is pass to the for each is from action node
This changes from async
Start Node -> AsyncNode (Action Node) ->  AsyncNode (For Each) -> AsyncNode (Action Node)

from node is the async node... not where it does come from. That is the whole problem. If you try to chain like normal flow you will get an infinite loop because asyncnode instance.
The algorithm of execution does not work the same way.

elguardian:
@MarianMacik @mswiderski 
ok. from the new perspective and for fixing the problem I needed to change the argument this when triggering the node execution from an AsyncEventNodeInstance so it is not coming from itself.

The unwrapping is still needed.

elguardian:
Jenkins retest this please

elguardian:
Jenkins retest this please

elguardian:
@MarianMacik any progress with this  ?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
MarianMacik: I understand the whole code except for this part. To me this will always return false in case of multi instance subprocess, which means that there will be no async node wrapped and subsequently no async node instance created (in the getNodeInstance() method of this class), so basically async mode won't be available for this type of subprocess node. Is that right or am I missing something? If it is not difficult to explain, why was this change needed?

MarianMacik: OK, probably got it. If it wasn't there, then if previous node was async, then it would wrap it as async event node and create again async event node instance (but we want the real node instance). Since the container is of type CompositeNodeInstance, we only need to handle it there. But still, afaik we use the from node instance to only compare its nodeId on lines 127-128 which should already point to the underlying actual node (since AsyncEventNode getId() delegates to the actual node and this is then preserved in async node instance as well).

elguardian:
@MarianMacik @mswiderski any progress ? we should do something about this, this week.

MarianMacik:
@elguardian I asked you about the unwrapping here https://github.com/kiegroup/jbpm/pull/1465#discussion_r283460677

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/CompositeNodeInstance.java
elguardian: Not really. The problem was triggering
https://github.com/kiegroup/jbpm/pull/1465/files#diff-0d58c1b1aa778882879a180d7b1f90bdR402

now the trigger has the right node because of the behaviour in AsyncEventNodeInstance is different
https://github.com/kiegroup/jbpm/pull/1465/files#diff-1f9184540bbaf5380b28bb54bcb5fbaeR94

otherwise from node would be the same node triggering.

wrapping problem is a different problem.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/SubProcessHandler.java
```diff
@@ -80,6 +81,8 @@ public Object end(final String uri, final String localName,
 			    Boolean isAsync = Boolean.parseBoolean((String)node.getMetaData().get("customAsync"));
 				// create new timerNode
 				ForEachNode forEachNode = new ForEachNode();
+                // indirect to the real node if it is an async event node
+                forEachNode.getCompositeNode().getMetaData().put(NodeInstance.METADATA_WRAP_ASYNC_NODE, false);
```
MarianMacik: I would use here rather setMetaData method.

PR #1470, opened 2019-04-01
[DO NOT MERGE YET] [RHBPMS-5231] Same job processed by 2 servers at the same time

tkobayas:
In order to avoid race condition written in RHBPMS-5231

- Check RequestInfo's status again
- De-prioritize PullExectuor's pick up by subtracting "now" (by default = 0 (no effect))

@mswiderski wrote:
> I’d say it’s good to go and as far as I can see it should not have negative side effects so might be good to be committed to 6.5.x as well. Though I leave that to Martin to decide as I don’t know what is planned for that branch moving forward

So please merge it if @martinweiler says go.


tkobayas:
As discussed with @mswiderski and @martinweiler , 6.5.x will not consume fixes from one-off patches. Closing this PR.

PR #1471, opened 2019-04-02
[JBPM-8305] getProcessInstance(readOnly=true) in workflow paused on task activity in subprocess causes extra run of first WorkItem after subprocess node.

elguardian:
disconnect again the process instance when it is in readOnly mode

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
@@ -174,6 +174,7 @@ public class JPAProcessInstanceManager
                 ((ProcessInstanceImpl) processInstance).reconnect();
                 if (readOnly) {
                     internalRemoveProcessInstance(processInstance);
+                    ((ProcessInstanceImpl) processInstance).disconnect();
                 }
             }
             return processInstance;
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/regression/subprocess/SubprocessDoublesTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/regression/subprocess/SubprocessDoublesTest.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.test.regression.subprocess;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+import org.jbpm.ruleflow.instance.RuleFlowProcessInstance;
+import org.jbpm.test.JbpmTestCase;
+import org.jbpm.workflow.instance.node.ActionNodeInstance;
+import org.jbpm.workflow.instance.node.SubProcessNodeInstance;
+import org.junit.Test;
+import org.kie.api.event.process.DefaultProcessEventListener;
+import org.kie.api.event.process.ProcessNodeLeftEvent;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.manager.RuntimeEngine;
+import org.kie.api.runtime.process.WorkItem;
+import org.kie.api.runtime.process.WorkItemHandler;
+import org.kie.api.runtime.process.WorkItemManager;
+
+import static org.junit.Assert.assertEquals;
+
+public class SubprocessDoublesTest extends JbpmTestCase {
+
+    public SubprocessDoublesTest() {
+        super(true, true);
+    }
+
+    class ExternalTaskWorkItemHandlerTest implements WorkItemHandler {
+
+        public void abortWorkItem(WorkItem workItem, WorkItemManager workItemManager) {
+
+        }
+
+        public void executeWorkItem(WorkItem workItem, WorkItemManager workItemManager) {
+            _lastWorkItemId = workItem.getId();
+        }
+
+        long _lastWorkItemId = 0;
+
+        public long getLastWorkItemId() {
+            return _lastWorkItemId;
+        }
+    }
+
+    @Test
+    public void testSubprocessDoubles() {
+        createRuntimeManager("org/jbpm/test/functional/subprocess/TestSubWorkflow.bpmn2",
+                             "org/jbpm/test/functional/subprocess/TestWorkflow.bpmn2");
+
+        RuntimeEngine runtimeEngine = getRuntimeEngine();
+        KieSession ksession = runtimeEngine.getKieSession();
+        WorkItemManager itemManager = ksession.getWorkItemManager();
+
+        ExternalTaskWorkItemHandlerTest externalTask = new ExternalTaskWorkItemHandlerTest();
+        itemManager.registerWorkItemHandler("GenericTask", externalTask);
+
+        RuleFlowProcessInstance processInstance = (RuleFlowProcessInstance) ksession
+                                                                                    .startProcess("TestProject.TestWorkflow");
+        final AtomicInteger counter = new AtomicInteger(0);
+        ksession.addEventListener(new DefaultProcessEventListener() {
+
+            @Override
+            public void afterNodeLeft(ProcessNodeLeftEvent event) {
+                if (event.getNodeInstance() instanceof ActionNodeInstance) {
+                    counter.incrementAndGet();
+                }
+            }
+        });
+
+        assertProcessInstanceActive(processInstance.getId());
+        assertNodeActive(processInstance.getId(), ksession, "subprocess");
+
+        SubProcessNodeInstance subprocessNode = (SubProcessNodeInstance) processInstance.getNodeInstances().toArray()[0];
+        RuleFlowProcessInstance subprocessInstance = (RuleFlowProcessInstance) ksession
+                                                                                       .getProcessInstance(subprocessNode.getProcessInstanceId());
+
+        // This code simaulates "GET process/instance" rest api call.
+        // Every call of getProcessInstance(id, true) causes one extra-call of firs WorkItem after subprocess.
+        ksession.getProcessInstance(processInstance.getId(), true);
+
+        assertNodeActive(subprocessInstance.getId(), ksession, "Task1");
+
+        // This code simulates "POST workitem/{workItemId}/complete".
+        itemManager.completeWorkItem(externalTask.getLastWorkItemId(), null);
+        assertProcessInstanceCompleted(processInstance.getId());
+        assertEquals(counter.get(), 1);
+    }
+}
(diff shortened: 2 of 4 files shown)


elguardian:
this ok to test.

fe-c:
Why [JBPM-8363] and not [JBPM-8305] as issue id?

elguardian:
@fe-c my bad.

elguardian:
this ok to test.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/StageActivationConditionTest.java
```diff
@@ -92,8 +92,8 @@ public void testAutoCompleteCheckActivationData() {
                                 @Override
                                 public Boolean execute(Context context) {
                                     KieSession ksession = ((RegistryContext) context).lookup( KieSession.class );
-                                    
-                                    WorkflowProcessInstance pi = (WorkflowProcessInstance) ksession.getProcessInstance(processInstanceId, true);
+
```
mswiderski: Why is this change needed?

elguardian: getVariable of node instance requires read/write process. if you get this read only the test will fail.

mswiderski: why getVariable requires read/write process?

elguardian: @mswiderski 
I cannot tell you why some operations were designed to work with read only or read / write. That is not documented anywhere.
The reason is when you disconnect the process this condition is fullyfilled
https://github.com/kiegroup/jbpm/blob/7.20.0.Final/jbpm-flow/src/main/java/org/jbpm/process/instance/impl/ProcessInstanceImpl.java#L97

causing failure in here
https://github.com/kiegroup/jbpm/pull/1471/files/e13dc20a8a8be0513b9aab4a1553d52703b31b3f#diff-6c5262fecaf86cc99369126786aad13cR97

mswiderski: what I am asking is why this is needed? It wasn't needed before and I don't see a point in writing anything to process instance info when just reading it's variables. So could you point me what is actually making to fail?

elguardian: the problem in the jira is that the listener were registed for a read only process instance. This was because while in some parts of the code the process instance reconnected then after being used was disconnected again.

That was happening here
https://github.com/kiegroup/jbpm/blob/7.19.0.Final/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java#L174

The collateral effect is that some operations that were being relying on the process instance were connected now are failing. Disconnecting the process instance causes the state variables to be null therefore you cannot navigate to the getProcess (this happens internally during getVariable)

stack trace for the getting variable
~~~
at org.jbpm.process.instance.impl.ProcessInstanceImpl.getProcess(ProcessInstanceImpl.java:97)
	at org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl.getWorkflowProcess(WorkflowProcessInstanceImpl.java:292)
	at org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl.getNodeContainer(WorkflowProcessInstanceImpl.java:124)
	at org.jbpm.workflow.instance.impl.NodeInstanceImpl.getNode(NodeInstanceImpl.java:134)
	at org.jbpm.workflow.instance.impl.NodeInstanceImpl.resolveContext(NodeInstanceImpl.java:434)
	at org.jbpm.workflow.instance.impl.NodeInstanceImpl.resolveContextInstance(NodeInstanceImpl.java:441)
	at org.jbpm.workflow.instance.impl.NodeInstanceImpl.getVariable(NodeInstanceImpl.java:487)
~~~

mswiderski: did you track down what is calling disconnect during the execution? It should only be invoked upon completion ... usually transaction completion.

elguardian: @mswiderski giving your comment the this makes me think that probably this fix is incorrect. I was following the logic based on this

https://github.com/kiegroup/jbpm/blob/7.20.0.Final/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java#L194

and copying in the in to the JPAProcessInstanceManager causing the disconnection. The disconnection causes removing the listener at several stages including the one giving problem.

mswiderski: @elguardian In general what I don't see a need is for reading operations we should not make modifications to the underlying entity - process instance info. hat's why I want to understand why we need to do it now. The main reason for it is that it will increase probability for optimistic lock exceptions as even getting process variables will update the version.
With that said, mixing readonly and write operations on process instance should be work as expected - as if I recall correctly readonly does the disconnect as you pointed out but also removes from the internal map of loaded instances and thus should be reloaded when write operation wants to access it. Maybe that would be the area to investigate more?

elguardian:
done @mswiderski created a method for setting the process instance as readonly

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -174,6 +174,7 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
                 ((ProcessInstanceImpl) processInstance).reconnect();
                 if (readOnly) {
                     internalRemoveProcessInstance(processInstance);
+                    ((ProcessInstanceImpl) processInstance).setReadOnly();
```
MarianMacik: Doesn't the `setReadOnly` method execute exactly the same method as `internalRemoveProcessInstance(processInstance);`? In `setReadOnly` method there is `...getProcessInstanceManager().internalRemoveProcessInstance(this);` The only addition I see is setting the `process` field to null in the `setReadOnly` method.

elguardian: what is your point ?

MarianMacik: Sorry, maybe the comment was confusing.

I am saying that `setReadOnly()` method calls:

```
public void setReadOnly() {
        ((InternalProcessRuntime) kruntime.getProcessRuntime()).getProcessInstanceManager().internalRemoveProcessInstance(this);
        process = null;
}
```

which is (specifically `internalRemoveProcessInstance(this)`) already called one line above in the code I commented:

```
if (readOnly) {
           internalRemoveProcessInstance(processInstance); // this is duplicated imho
          ((ProcessInstanceImpl) processInstance).setReadOnly();
```

And since the implementation of `internalRemoveProcessInstance()` is:

https://github.com/kiegroup/jbpm/blob/de4094d0023fe11e6188fa86fea85f85c0cddfc3/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java#L203-L205

Then I think the only real addition made by the new `setReadOnly` method is that it does `process = null;`. So why not only call the new `setReadOnly()` method in that if block?

But maybe there is something more to it which I didn't notice.

elguardian: my bad... I made a mistake with the latest push and the some changes were missing ..

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/SubprocessDoublesTest.java
```diff
@@ -0,0 +1,102 @@
...
+
+        assertProcessInstanceActive(processInstance.getId());
+        assertNodeActive(processInstance.getId(), ksession, "subprocess");
+
+        SubProcessNodeInstance subprocessNode = (SubProcessNodeInstance) processInstance.getNodeInstances().toArray()[0];
+        RuleFlowProcessInstance subprocessInstance = (RuleFlowProcessInstance) ksession
+                                                                                       .getProcessInstance(subprocessNode.getProcessInstanceId());
+
+        // This code simaulates "GET process/instance" rest api call.
+        // Every call of getProcessInstance(id, true) causes one extra-call of firs WorkItem after subprocess.
```
mswiderski: I don't understand this comment, what is an extra-call of first work item?

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -174,6 +174,9 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
                 ((ProcessInstanceImpl) processInstance).reconnect();
                 if (readOnly) {
```
mswiderski: I am starting to think if we need to make all this changes ... maybe we could simply not reconnect the process instance - not registering any listeners.... as this is just for reading the state of the process instance and should never react to any singals...wdyt?

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/subprocess/SubprocessDoublesTest.java
elguardian: it is referred that the the process invokes twices the listener once you get the read only process because it registers twice the same listener

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
MarianMacik: +1 from me, actually I think we discussed this approach but I wasn't sure if it would break something else or not. Then this would mean to just add else branch for `if (readOnly) {` with the code `((ProcessInstanceImpl) processInstance).reconnect();` I think.

elguardian:
ok to test

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
elguardian: it does not work.

MarianMacik:
jenkins retest this

elguardian:
jenkins retest this

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
elguardian: @MarianMacik it is not enough to reconnect... the listeners can be added anywhere so you have no control where they are added. not all the logic is centralized so it is rather difficult to unravel the whole thing.

elguardian:
jenkins retest this

PR #1476, opened 2019-04-07
[JBPM-8380] Remove JBPMHelper

karlnicholas:
[JBPM-8380] Remove JBPMHelper
Remove org.jbpm.test.JBPMHelper and org.jbpm.test.JbpmJUnitTestCase.
Add org/jbpm/test/util/SetupExamplesDatasource to create simple H2 memory Datasource for jbpm-examples.
Move some setup methods from JBPMHelper to org/jbpm/test/ProcessMain.
Move processStateName enumeration from JBPMHelper to JbpmJUnitBaseTestCase.


kie-ci4:

MarianMacik:
This is already being fixed by #1449. This Jira is only about removal of JBPMHelper from the code base. It should follow after #1449 is merged.

karlnicholas:
I didn't see PersistenceUtil so I will remove this PR and do a little more work to use PersistenceUtil. It would be nice to get this and JBPM-8097 merged so that the examples are working and the code is cleaner.

MarianMacik:
@karlnicholas I think that after #1449 is merged, examples should run fine. Then we only need to basically delete the class since it is probably not used at other places. At least that was my understanding.

karlnicholas:
@MarianMacik Yes, the work required to remove JBPMHelper was mostly the examples but there were some static fields that were still referenced in one or two places. There is a test case in the src/test/java/org.jbpm.test.ProcessMain that uses several routines in the JBPMHelper class for setup. I moved those routines into that class and I moved the static fields into JbpmJUnitBaseTestCase. 

So, in short, #1449 is about getting the examples to work and #1476 is about doing a little extra work to get remove of the old classes completely. It does make sense to merge #1476 first since otherwise the work on the examples is either duplicated or done differently neither of which is good.

PR #1477, opened 2019-04-08
[JBPM-8380]

karlnicholas:
Started with #1449 and did additional work to remove JBPMHelper.java and JBPMUnitTestCase.
Moved static routines from JBPMHelper to ProcessMain where they were exclusively used.
Moved a couple static fields from JBPMHelper to JbpmJunitBaseTestBase.java.


kie-ci4:

elguardian:
@MarianMacik @mswiderski this is based on my previous commit.

@karlnicholas plz keep in mind that there are other places where this is used
https://github.com/kiegroup/droolsjbpm-tools/blob/401c43085cb1c00df11c1bf325a7e2c39a61d865/drools-eclipse/org.jbpm.eclipse/src/main/resources/org/jbpm/eclipse/wizard/project/ProcessMain-advanced.java.v5.template

Also claim the Jira when you are working on it. otherwise will cause a misscordination among contributors.

Thanks for the contribution !

elguardian:
this is ok to test

jbpm-test/src/test/java/org/jbpm/test/ProcessMain.java
```diff
@@ -48,12 +67,103 @@ private static KieBase readKnowledgeBase() throws Exception {
...
+	    String datasourceName = properties.getProperty("datasourceName", "jdbc/jbpm-ds");
+		PersistenceUtil.setupPoolingDataSource(properties, datasourceName);
+	    String humanTaskEnabled = properties.getProperty("taskservice.enabled", "false");
+	    if ("true".equals(humanTaskEnabled)) {
+	        startTaskService();
+	    }
+	}
+
+	public
+	static void cleanupSingletonSessionId() {
```
MarianMacik: This should be on one line.

jbpm-test/src/test/java/org/jbpm/test/ProcessMain.java
```diff
@@ -48,12 +67,103 @@ private static KieBase readKnowledgeBase() throws Exception {
...
+	                return name.endsWith("-jbpmSessionId.ser");
+	            }
+	        });
+	        for (String file : jbpmSerFiles) {	            
+	            new File(tempDir, file).delete();
+	        }
+	    }
+	}
+
+	public static TaskService startTaskService() {
```
MarianMacik: Every method except `main` and `startUp` can be declared private here. By doing this we will be sure that it is only used in the `ProcessMain` class.

elguardian:
@karlnicholas use the same naming convention as this one (same branch name)

karlnicholas:
@elguardian Okay, I looked at the droolsjbpm-tools and create PR https://github.com/kiegroup/droolsjbpm-tools/pull/93.
One of the plugin code samples in the droolsjbpm-tools project is an exact match of https://github.com/kiegroup/jbpm/blob/master/jbpm-test/src/test/java/org/jbpm/test/ProcessMain.java and both of them require JBPMHelper to function properly. This is because JBPMHelper handles two JPA instantiations which these codes require. Previously I had just moved that code to the above ProcessMain but now I see it is required in two separate locations.

ProcessMain (and the plugin snippet) is not currently functioning properly without the changes to https://github.com/kiegroup/jbpm/blob/master/jbpm-test/src/test/filtered-resources/META-INF/persistence.xml made by this PR. This is because the schema for persistence unit "org.jbpm.services.task" has been changed. I copied an updated version from somewhere else that works. 

This means that the ProcessMain code was allowed to fall into disrepair since it is not tested during the build processes. I would recommend a new Jira be opened to convert ProcessMain to a properly functioning JUnit test case.

elguardian:
@MarianMacik ^^

mswiderski:
ok to test

karlnicholas:
@mswiderski - Did I break something? Is there anyway to get to the test logs?

karlnicholas:
Also please issue ok to test on [Remove JBPMHelper #93](https://github.com/kiegroup/droolsjbpm-tools/pull/93)

mswiderski:
@MarianMacik @elguardian guys, could you please review and approve so it can be merged?

MarianMacik:
@karlnicholas So if I understood correctly you didn't remove JBPMHelper because it manages 2 datasources? Is this what you mean by JPA installations? I can see that JBPMHelper is instantiating the jdbc/jbpm-ds datasource here:

https://github.com/kiegroup/jbpm/pull/1477/files#diff-2ac0d6e56076eabf2e312f84f53a4663R54

But I cannot see where it is instantiating the second one for task service which is later used in this method:

https://github.com/kiegroup/jbpm/pull/1477/files#diff-2ac0d6e56076eabf2e312f84f53a4663R62

I don't understand why we cannot simply get rid of this class and move the relevant code to the let's say AbstractMain class or just create a new ExamplesUtil method or whatever. And I am wondering how it works with that second datasource since it is not created anywhere :) Then  `Persistence.createEntityManagerFactory(properties.getProperty("taskservice.datasource.name", "org.jbpm.services.task"), map);` should simply fail.

jbpm-examples/src/main/resources/datasource.properties
```diff
@@ -0,0 +1,23 @@
+# for creating a datasource
+driverClassName=org.h2.Driver
+className=org.h2.jdbcx.JdbcDataSource
+user=sa
+password=
+url=jdbc:h2:tcp://localhost/~/jbpm-db
+datasourceName=jdbc/jbpm-ds
```
MarianMacik: You can leave only these properties here, others should not be used by PersistenceUtil class. They are only used by current jBPMHelper class.

karlnicholas: These two, task service are used by ProcessMain which uses JBPMHelper. See additional comments below.

taskservice.enabled=true
taskservice.datasource.name=org.jbpm.services.task

karlnicholas:
@MarianMacik ProcessMain is the issue. It's in src/test/java/org/jbpm/test. This code is an exact copy of the code in the EclipsePlugin in the droolsjbpm-tools (somewhere, see PR). This is the "two" places that JBPMHelper is used. This project and the other droolsjbpm-tools project. 

In the droolsjbpm-tools project the code seems to be part of a template for an eclipse plugin. So, I assume after looking at it briefly that a user in eclipse can ask for the template and the class will be injected a project somewhere. Sort of a quickstart. 

I don't know why the class (ProcessMain and also ProcessKModuleMain) are here in this project as well. I assume so that it can be run and tested. However the two classes are not tested as part of the build.

So, if you remove JBPMHelper the eclipse plugin will have template code that fails. 

Having template code that fails is probably no big deal since it's not working at the moment anyway.  This PR will fix that, but at the cost of leaving JBPMHelper in the project.


Regards the second datasource mentioned. The persistence.xml file used by JBPMHelper contains several datasources. Once JPA reads it all the datasources are available to the running application. ProcessMain demonstrates using a TaskService so it not only connects to org.jbpm.services.task

MarianMacik:
OK, I see that it those two are persistence units reusing one datasource. So the property taskservice.datasource.name is quite misleading.

Never mind, regarding the `jbpm-examples/src/main/resources/datasource.properties ` file, I still don't see how it is used by JBPMHelper since JBPMHelper `getProperties()` method only reads `JBPM.properties` file and not `datasource.properties` one. So I still think those other properties can be omitted since they are taken from `JBPM.properties` file.

Regarding deletion of JBPMHelper. I understand that JBPMHelper is used in Eclipse plugin etc. But if you replace JBPMHelper with a new ExamplesUtil as I mentioned and you just leave there what is actually used, I think we would be safe. The issue with leaving JBPMHelper is that it has been marked as deprecated so I think we have to come up with a new class and only leave there functionality which we really need.

I now realized that JBPMHelper says:

```
Since version 6.0 this class is deprecated. Instead <code>RuntimeManager</code> should be used directly.
See documentation on how to use <code>RuntimeManager</code>
```

So I think we need to just rewrite ProcessMain as ProcessKModuleMain to use runtime manager and engine and obtain ksession from there. Then we can get rid of the `loadStatefulKnowledgeSession` method etc.
So we have 2 possibilities I think - replace JBPMHelper with let's say ExamplesUtil and retain there only functionality which is needed - which means migrate to RuntimeManager way of setup, or delete JBPMHelper and move desired code to the ProcessMain/ProcessKModuleMain classes - but both of them will use runtime manager. Can you try that?

karlnicholas:
Okay, it wasn't so bad after looking a little closer to put the code completely in the ProcessMain and ProcessKModuleMain in jbpm-test and the plugin template as well. I just did away with reading any kind of configuration file and hardcoded driver parameters for H2. I think it is fine since it is just example code anyway. Really, it seems much cleaner anyway since you can easily see what's going on.

jbpm-test/src/test/java/org/jbpm/test/ProcessMain.java
```diff
@@ -48,12 +64,44 @@ private static KieBase readKnowledgeBase() throws Exception {
...
+		    Map<String, String> map = new HashMap<String, String>();
+		    map.put("hibernate.dialect", "org.hibernate.dialect.H2Dialect");
+	        EntityManagerFactory emf = Persistence.createEntityManagerFactory(properties.getProperty("persistence.persistenceunit.name", "org.jbpm.persistence.jpa"), map);            	            
+	        builder = RuntimeEnvironmentBuilder.Factory.get()
+                .newDefaultBuilder()
+                .entityManagerFactory(emf)                
+                .addEnvironmentEntry(EnvironmentName.TRANSACTION_MANAGER, com.arjuna.ats.jta.TransactionManager.transactionManager());
+    	} else {
+            builder = RuntimeEnvironmentBuilder.Factory.get()
+                .newDefaultInMemoryBuilder();
```
MarianMacik: I see that in droolsjbpm-tools you did
```
DeploymentDescriptor descriptor = 
		new DeploymentDescriptorManager().getDefaultDescriptor().getBuilder().auditMode(AuditMode.NONE).get();	
            builder.addEnvironmentEntry("KieDeploymentDescriptor", descriptor);
```

Which is triggered if `usePersistence == false`, but this is always true. So I reckon it is for users to manually change in the code. But anyway it probably needs to be added here as well.

karlnicholas: Changes made. I opened a new Jira ticket because it seems unseemly to have to set AuditMode to NONE for a memory Runtime. Removed datasource.properties because the default used by PersistenceUtil is sufficient.

MarianMacik: I think that the reason is you can have only auditing turned on - for example to store events from engine, but engine itself won't use persistence for persisting process instances etc. But maybe it is really not intended.

MarianMacik:
looks good now, will try to run full downstream build just to check if nothing is broken

MarianMacik:

jbpm-test/src/test/java/org/jbpm/test/ProcessMain.java
karlnicholas: Yea, the DeploymentDescriptor has AuditMode.JPA by default and the code `RuntimeEnvironmentBuilder.Factory.get() .newDefaultInMemoryBuilder();` doesn't try to change it. I think it would make the most sense to have that above routine or whatever is core to creating an in memory runtime environment set AuditMode.NONE but I didn't want to do it as part of this Jira ticket.

karlnicholas:
Also don't forget jenkins execute full downstream build

MarianMacik:
Yes, I know about that one too. For that we don't have downstream build, just the repo build.
Here, in FDB there were Just 2 unrelated tests, will trigger it once more and if it is OK, we can merge.

MarianMacik:
OK, I think we are good to go here. @elguardian can you approve?

karlnicholas:
Are this build errors a problem with this PR?

mareknovotny:
2 from 3 failed tests were seen before so they are not related, but I am not sure about `org.kie.wb.selenium.ui.ProjectLibraryIntegrationTest.importAndBuildProjectFromStockRepository`

karlnicholas:
> 2 from 3 failed tests were seen before so they are not related, but I am not sure about org.kie.wb.selenium.ui.ProjectLibraryIntegrationTest.importAndBuildProjectFromStockRepository

I looked: This is something from the Workbench and seems to be importing/testing a project named OptaCloud which seems to be about OptaPlanner. Shouldn't have anything to do with these changes.

MarianMacik:
I am 100% sure issues are not related. Can you rebase so we will merge this?

karlnicholas:
@MarianMacik  Okay, I merged the latest changes from the master branch. Since there were conflicts I needed to create a new commit to support the merge changes. Should be good to go.

MarianMacik:
jenkins retest this

MarianMacik:

MarianMacik:
jenkins retest this

MarianMacik:

MarianMacik:
Looks good after full downstream. @mswiderski any comments?

PR #1478, opened 2019-04-16
add WARN for a combination of Singleton strategy and EJB Timer Scheduler

tkobayas:
Discussed with @mswiderski 

I wanted to avoid to add dependency 'jbpm-services-ejb-timer' so it may look a bit verbose. Can this be written more concise?

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/SingletonRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/SingletonRuntimeManager.java
@@ -25,7 +25,9 @@ import java.io.ObjectOutputStream;
 import org.drools.core.common.InternalKnowledgeRuntime;
 import org.drools.core.runtime.process.InternalProcessRuntime;
 import org.drools.persistence.api.TransactionManager;
+import org.jbpm.process.core.timer.GlobalSchedulerService;
 import org.jbpm.process.instance.ProcessRuntimeImpl;
+import org.jbpm.runtime.manager.api.SchedulerProvider;
 import org.jbpm.services.task.impl.TaskContentRegistry;
 import org.kie.api.command.ExecutableCommand;
 import org.kie.api.runtime.KieSession;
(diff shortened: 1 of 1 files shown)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -125,6 +125,18 @@ public AbstractRuntimeManager(RuntimeEnvironment environment, String identifier)
         }
         this.executionErrorManager = new ExecutionErrorManagerImpl(storage);
         ((SimpleRuntimeEnvironment)environment).getEnvironmentTemplate().set(EnvironmentName.EXEC_ERROR_MANAGER, executionErrorManager);
+        if ((this instanceof SingletonRuntimeManager) && (environment instanceof SchedulerProvider)) {
```
mswiderski: wouldn't it be better to place it in SingletonRuntimeManager? then we don't have to cast it any more

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -125,6 +125,18 @@ public AbstractRuntimeManager(RuntimeEnvironment environment, String identifier)
...
+            GlobalSchedulerService schedulerService = ((SchedulerProvider) environment).getSchedulerService();
+            Class<?> clazz;
+            try {
+                clazz = Class.forName("org.jbpm.services.ejb.timer.EjbSchedulerService");
```
MarianMacik: Cannot this also happen with the Quartz scheduler?

mswiderski: no, as quartz does not control transaction so it's not an issue

tkobayas:
@mswiderski Thank you for the suggestion. I moved it to SingletonRuntimeManager.

PR #1479, opened 2019-04-17
[JBPM-8430] RETRY handlingStrategy should handle exception multiple times while handling exception via subprocess

elguardian:
change a bit the work item handler

executed work item handler -> error handling -> subprocess fork -> strategy -> retry loop

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
@@ -92,6 +92,8 @@ public class WorkItemNodeInstance extends StateBasedNodeInstance implements Even
     
     private long exceptionHandlingProcessInstanceId = -1;
 
+    private int triggerCount = 0;
+
     protected WorkItemNode getWorkItemNode() {
         return (WorkItemNode) getNode();
     }
@@ -127,34 +129,44 @@ public class WorkItemNodeInstance extends StateBasedNodeInstance implements Even
         if (getNodeInstanceContainer().getNodeInstance(getId()) == null) {
             return;
         }
-        // TODO this should be included for ruleflow only, not for BPEL
-        //        if (!Node.CONNECTION_DEFAULT_TYPE.equals(type)) {
-        //            throw new IllegalArgumentException(
-        //                "A WorkItemNode only accepts default incoming connections!");
-        //        }
+
         WorkItemNode workItemNode = getWorkItemNode();
         createWorkItem(workItemNode);
         if (workItemNode.isWaitForCompletion()) {
             addWorkItemListener();
         }
+
+        WorkItemManager workItemManager = (WorkItemManager) ((ProcessInstance) getProcessInstance()).getKnowledgeRuntime().getWorkItemManager();
         String deploymentId = (String) getProcessInstance().getKnowledgeRuntime().getEnvironment().get(EnvironmentName.DEPLOYMENT_ID);
         workItem.setDeploymentId(deploymentId);
         workItem.setNodeInstanceId(this.getId());
         workItem.setNodeId(getNodeId());
+
+        processWorkItemHandler(() -> workItemManager.internalExecuteWorkItem((org.drools.core.process.instance.WorkItem) workItem));
+
+        if (!workItemNode.isWaitForCompletion()) {
+            triggerCompleted();
+        }
+        this.workItemId = workItem.getId();
+    }
+
+    private void processWorkItemHandler(Runnable handler) {
         if (isInversionOfControl()) {
             ((ProcessInstance) getProcessInstance()).getKnowledgeRuntime()
                                                     .update(((ProcessInstance) getProcessInstance()).getKnowledgeRuntime().getFactHandle(this), this);
         } else {
             try {
-                ((WorkItemManager) ((ProcessInstance) getProcessInstance())
-                                                                           .getKnowledgeRuntime().getWorkItemManager()).internalExecuteWorkItem(
-                                                                                                                                                (org.drools.core.process.instance.WorkItem) workItem);
+                handler.run();
             } catch (WorkItemHandlerNotFoundException wihnfe) {
                 getProcessInstance().setState(ProcessInstance.STATE_ABORTED);
                 throw wihnfe;
             } catch (ProcessWorkItemHandlerException handlerException) {
-                this.workItemId = workItem.getId();
-                handleWorkItemHandlerException(handlerException, workItem);
+                if (triggerCount++ < handlerException.getRetries() + 1) {
+                    this.workItemId = workItem.getId();
+                    handleWorkItemHandlerException(handlerException, workItem);
+                } else {
+                    throw handlerException;
+                }
             } catch (Exception e) {
                 String exceptionName = e.getClass().getName();
                 ExceptionScopeInstance exceptionScopeInstance = (ExceptionScopeInstance) resolveContextInstance(ExceptionScope.EXCEPTION_SCOPE, exceptionName);
@@ -166,10 +178,6 @@ public class WorkItemNodeInstance extends StateBasedNodeInstance implements Even
                 exceptionScopeInstance.handleException(exceptionName, e);
             }
         }
-        if (!workItemNode.isWaitForCompletion()) {
-            triggerCompleted();
-        }
-        this.workItemId = workItem.getId();
     }
 
     protected WorkItem createWorkItem(WorkItemNode workItemNode) {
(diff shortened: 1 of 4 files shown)


elguardian:
this is ok to test.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
```diff
@@ -127,28 +127,34 @@ public void internalTrigger(final NodeInstance from, String type) {
...
+
+        processWorkItemHandler(() -> workItemManager.internalExecuteWorkItem((org.drools.core.process.instance.WorkItem) workItem));
+
+        if (!workItemNode.isWaitForCompletion()) {
+            triggerCompleted();
+        }
+        this.workItemId = workItem.getId();
+    }
+
+    private void processWorkItemHandler(Runnable handler) {
```
mswiderski: why do we have runnable here?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
```diff
@@ -127,28 +127,34 @@ public void internalTrigger(final NodeInstance from, String type) {
...
         if (isInversionOfControl()) {
             ((ProcessInstance) getProcessInstance()).getKnowledgeRuntime()
                                                     .update(((ProcessInstance) getProcessInstance()).getKnowledgeRuntime().getFactHandle(this), this);
         } else {
             try {
-                ((WorkItemManager) ((ProcessInstance) getProcessInstance())
-                                                                           .getKnowledgeRuntime().getWorkItemManager()).internalExecuteWorkItem(
-                                                                                                                                                (org.drools.core.process.instance.WorkItem) workItem);
+                handler.run();
```
mswiderski: is this spinning a new thread?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
```diff
@@ -587,17 +588,11 @@ private void handleWorkItemHandlerException(ProcessWorkItemHandlerException hand
         ((ProcessInstanceImpl) processInstance).setParentProcessInstanceId(getProcessInstance().getId());
         ((ProcessInstanceImpl) processInstance).setSignalCompletion(true);
 
+        addExceptionProcessListener();
         kruntime.startProcessInstance(processInstance.getId());
-        if (processInstance.getState() == ProcessInstance.STATE_COMPLETED
-                || processInstance.getState() == ProcessInstance.STATE_ABORTED) {
-            exceptionHandlingCompleted(processInstance, handlerException);
```
mswiderski: why skipping check of the process instance state?

elguardian: avoid recursion if the retry tries too hard.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
elguardian: nope... just convinient interface.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
mswiderski: can you elaborate more on this case?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
mswiderski: this got me really confused... any other convenient interface we could use that does not bring threading into the picture ... or is this just me...

elguardian: https://docs.oracle.com/javase/8/docs/api/java/util/function/package-summary.html

Not as far as I know. Runnable is marked as Functional interface. I can create a new interface just for this case.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/WorkItemNodeInstance.java
elguardian: this was during testing. I know this needs to be tuned.
One of the problems I see here is that the Jira is asking for retry but there is no "stop condition". During testing I found that if you set this you will get a stack overflow instead of an infinite loop.
That is one of the requirements that needs to be clarified. Either we let process control the retries or we set some sort of limit (not really sure if possible, because we need to count)

elguardian: @mswiderski Added the test case so you can see what I mean.

MarianMacik: I have been thinking about this and I think that registering the listener and then starting a subprocess instance is worse approach than just decision based on the subprocess instance outcome (i.e. if it is completed/aborted, continue, otherwise register a listener). I think this change just changes the "amount" of information we keep on stack while recursion is in progress. Obiously, I am now talking with scenario when the subprocess instance is finished in one go, otherwise the change made here has no effect since the wait state is reached and listener attached.

Before:
If the subprocess instance is already finished, we can continue with retrying. This means we are not anymore in `kruntime.startProcessInstance()` call, this is already done and we can retry. Then in case of exception, we have to recursively catch exception, call `handleWorkItemHandlerException()` method again, which will run the process instance and again decide what is its outcome and continue further. So with each iteration `kruntime.startProcessInstance(processInstance.getId());` call is finished, we are not recursing through this.

So the stack is something like `handleWorkItemHandlerException() ... -> ... exceptionHandlingCompleted() ... -> ... handleWorkItemHandlerException() ... -> ... exceptionHandlingCompleted() ...`

After, i.e. attaching the listener way in all use cases:
Imagine the subprocess instance is completed in one go, that would mean that it sends the completed event to all nodes, including the "calling" WorkitemNodeInstance, which will trigger `exceptionHandlingCompleted()`, then next retry and another exception needed to be caught and handled in `handleWorkItemHandlerException()` - but this is still the part of the "parent" `kruntime.startProcessInstance(processInstance.getId());` call, i.e. the call of starting of the first subprocess instance is de facto not finished yet. So in this case we keep recursing in `kruntime.startProcessInstance(processInstance.getId());`, which then calls `exceptionHandlingCompleted()`, which then calls `handleWorkItemHandlerException()`.

So in this case the stack is something like `handleWorkItemHandlerException() ... -> ... kruntime.startProcessInstance() ...-> ... exceptionHandlingCompleted() ... -> ... handleWorkItemHandlerException() ... -> ... kruntime.startProcessInstance() ... -> ... exceptionHandlingCompleted() ...`

In the second example we are recursing into startProcessInstance which I think is not needed and it may be confusing since the workflow of this mechanism is to finish subprocess instance first and then continue based on the handling strategy, which the first and the current implementation reflects better.

And regarding the recursion limit - I would go the way of setting this to 3 by default as a field of the `AbstractLogOrThrowWorkItemHandler` and possibly let this be configurable through another constructor parameter of the specific WIH. This way it will be configurable per WIH. Then you can pass the limit in exception - like the `handlingProcessId` and `handlingStrategy` is right now and the current count can be saved as a field in `WorkItemNodeInstance` which should be serialized, so it won't reset after reload.

Maybe it is a bad idea, but I hope I am not completely wrong on this approach. @mswiderski wdyt?

mswiderski: I would certainly give it a try as it does sound correct to me

elguardian:
now  depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/381

jbpm-workitems/jbpm-workitems-core/src/main/java/org/jbpm/process/workitem/core/AbstractLogOrThrowWorkItemHandler.java
```diff
@@ -47,7 +49,7 @@ protected void handleException(Throwable cause) {
     protected void handleException(Throwable cause,
                                    Map<String, Object> handlerInfoMap) {
         if (handlingProcessId != null && handlingStrategy != null) {
-            throw new ProcessWorkItemHandlerException(handlingProcessId, handlingStrategy, cause);
+            throw new ProcessWorkItemHandlerException(handlingProcessId, handlingStrategy, cause, retries);
```
MarianMacik: I see you added other constructors for `ProcessWorkItemHandlerException` as well, which have a default number of retries. This can cause issues in case number of retries is changed in retries field in AbstractLogOrThrowWorkItemHandler, but then the version of Exception constructor without retries is used. This will get out of sync, for example user then expects to set it for 5 retries, but then he forgets to pass it to the Exception constructor. Therefore, I think we should have retries parameter in all constructors of that exception; this will take it mandatory.

MarianMacik: OK, forget what I said here, then it would be confusing in case we use other handling strategies...

elguardian:
@MarianMacik @mswiderski anything stopping this from merge ?

mswiderski:
not from my end, @MarianMacik how about you?

MarianMacik:
I have added my opinion in https://github.com/kiegroup/jbpm/pull/1479#discussion_r283244597 but there hasn't been any reply.

elguardian:
done @MarianMacik

PR #1487, opened 2019-04-26
[6.5.x] [RHBPMS-5247] Control which classes are registered for serialization at kjar level

elguardian:
port http://git.app.eng.bos.redhat.com/git/kiegroup/jbpm.git/commit/?h=RHBPMS-5237&id=869f89a0c8666e89d817c4880d300c992de463aa

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/KModuleDeploymentService.java
@@ -16,8 +16,6 @@
 
 package org.jbpm.kie.services.impl;
 
-import static org.kie.scanner.MavenRepository.getMavenRepository;
-
 import java.io.StringReader;
 import java.io.UnsupportedEncodingException;
 import java.lang.annotation.Annotation;
@@ -82,6 +80,8 @@ import org.reflections.util.ConfigurationBuilder;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.InputSource;
+
+import static org.kie.scanner.MavenRepository.getMavenRepository;
 
 
 public class KModuleDeploymentService extends AbstractDeploymentService {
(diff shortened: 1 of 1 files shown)


MarianMacik:
BTW this change is not backward-compatible. Do we really want this in 6.5.x stream? I have just noticed that in DeploymentDescriptor the default is `false`:

https://github.com/kiegroup/jbpm/blob/2b9fe30ace384ed81bbc30242c7d3cc64794e351/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/deploy/DeploymentDescriptorImpl.java#L111

So we probably will need to alter the default there as well, not mentioning the JavaDoc:

https://github.com/kiegroup/droolsjbpm-knowledge/blob/a8ea70a510a44d9a50189cfbbab0d275be75fe47/kie-internal/src/main/java/org/kie/internal/runtime/conf/DeploymentDescriptor.java#L154-L162

@mswiderski isn't it the case that all deployment units have at least default deployment descriptor?

elguardian:
I will see the failing tests today
@MarianMacik this is already in a one-off in 6.4.z (6.5.x branch).
In theory in SET procedure a one-off should have a jira provided in the sustaining stream it was merged and become blocker for the next CP (no regressions policy). So if there is one-off approved that means it should be automatically devel-ack by SET and provided a forward port. otherwise that one-off should not exists.
@mareknovotny thoughts ?

mareknovotny:
@martinweiler please check previous comment from https://github.com/kiegroup/jbpm/pull/1487#issuecomment-487638665 as it seems this shoulld not be included in micro due backward compatible change.

martinweiler:
@MarianMacik - true, this fix changes the behavior from 6.4.11. This was discussed [here](http://post-office.corp.redhat.com/archives/jboss-support-droolsjbpm/2018-August/msg00131.html) and @mswiderski suggested this code change.

@elguardian true that we should have a JIRA for the branch first before we create a one-off, but this was not done here as the plan had been not to release any CPs after 6.4.11.

MarianMacik:
I'm OK with that, then we should just emphasize this change in behaviour in the release notes so users are not surprised with the change.

elguardian:
Jenkins retest this please.

mareknovotny:
We need a decision if we can merge it due the change of the behavior or NOT.

krisv:
@mswiderski can you confirm and if so merge?

MarianMacik:
also what about the https://github.com/kiegroup/jbpm/pull/1487#issuecomment-487638665 - shouldn't we alter JavaDoc and default value for DeploymentDescriptor class as well?

mswiderski:
from my point of view this is not really breaking any compatibility as it is more to guard the deployment itself from attempting to deploy classes that won't really work for remote invocation. So +1 from to go ahead and merge.

mareknovotny:
Jenkins retest this please.

PR #1488, opened 2019-05-03
JBPM-8459 - multiInstanceLoopCharacteristics cause NPE if output coll…

tsurdilo:
…ection is not specified


tsurdilo:
@MarianMacik hi, would you mind taking a look so can merge after? thanks!

PR #1493, opened 2019-05-16
JBPM-8485 - Add support ISO8601 expressions for user task notifications

tsurdilo:


tsurdilo:
deadline expiresAt supports now the already existing durations like "2s", full ISO date string and the ISO period format like "PT2S" for example.

MarianMacik:
Review provided, just minor thing and one idea...

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/DeadlinesBaseTest.java
```diff
@@ -49,322 +47,385 @@
 import org.kie.internal.task.api.model.InternalTask;
 import org.kie.internal.task.api.model.InternalTaskData;
 
-
+import static org.assertj.core.api.Assertions.*;
```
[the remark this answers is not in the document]
tsurdilo: Reverted formatting

jbpm-flow/src/main/java/org/jbpm/process/core/timer/DateTimeUtils.java
```diff
@@ -73,7 +74,24 @@ public static long parseDateAsDuration(String dateTimeStr) {
...
+                Duration duration = Duration.between(OffsetDateTime.now(), dateTime);
+
+                return (new Date(System.currentTimeMillis() +  duration.toMillis()));
+            }
+        } catch(Exception e) {
+            return (new Date(System.currentTimeMillis() + TimeUtils.parseTimeString(durationStr)));
+        }
+
+    }
+
```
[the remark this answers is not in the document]
tsurdilo: Those methods are currently used by some other core classes. I would feel hesitant to change them as it would involve adding a TON more of tests for both ISO date and period format for things that this Jira does not ask for, and it might cause issues that are not related. We could do this at some later point if we want to support iso date/period format for everything but I just don't feel comfortable doing it for this jira. wdyt?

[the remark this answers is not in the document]
MarianMacik: Ah, you are right, this would implicitly enable support for things we don't want yet. :)

mswiderski:
Jenkins retest this

elguardian:
@mswiderski @tsurdilo 
This is not correct from my point of view. If you specify PT2S ... the reminder will be sent every "Expires At" value until the Human Task is submitted. which is against the ISO format. Also this is not fullfillying the requirment.
I don't see any change in the logic of periodic repetition in here.

https://issues.jboss.org/browse/BAPL-1303 
Just parsing ISO string is not enough in this case.

tsurdilo:
@elguardian @mswiderski updated for periodic repeat.

jbpm-human-task/jbpm-human-task-workitems/src/test/java/org/jbpm/services/task/wih/util/HumanTaskHandlerHelperTest.java
```diff
@@ -137,130 +304,986 @@ public void testSetDeadlinesNotStartedReassignTwoTimes() {
...
-	
+
 	@Test
-	public void testSetDeadlinesNotCompletedReassignTwoTimes() {
-		
+	public void testSetDeadlinesNotStartedReassignTwoTimesWithISOExpirationTimePeriodFormatWithRepeatCount() {
 		WorkItem workItem = new WorkItemImpl();
-		workItem.setParameter("NotCompletedReassign", "[users:john]@[4h,6h]");
-		
+		workItem.setParameter("NotStartedReassign", "[users:john]@[R3/PT4H,R/PT6H]");
```
elguardian: This is a strange one... you are setting one limited repetition and the other one unbounded repetition... does not look strange to you ?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/commands/ExecuteDeadlinesCommand.java
```diff
@@ -171,9 +171,13 @@ public Void execute(Context context) {
 	        }
 	        
 	        deadline.setEscalated(true);
-	        persistenceContext.updateDeadline(deadline);
-	        persistenceContext.updateTask(task);
-            DeadlineSchedulerHelper.rescheduleDeadlinesForTask((InternalTask) task, ctx, type);
+			persistenceContext.updateDeadline(deadline);
+			persistenceContext.updateTask(task);
+
+	        if(deadline.getRepeatCount() != 0) {
```
elguardian: where is the logic regarding the number of repetition ? I mean you set the variable in 
https://github.com/kiegroup/jbpm/pull/1493/files#diff-1a2981938587c542131221bf215dea1cR215

but don't see where you are decrementing till reach zero

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +203,43 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+				} else {
+					// just "R/" is accepted
+					deadLine.setRepeatCount(-1);
+				}
+
+				if (DateTimeUtils.isPeriod((tempDateTimeStr))) {
+					deadLine.setDate(new Date(System.currentTimeMillis() + Duration.parse(tempDateTimeStr).toMillis()));
+				} else {
+					deadLine.setDate(new Date(System.currentTimeMillis() + DateTimeUtils.parseDateAsDuration(tempDateTimeStr)));
+				}
```
MarianMacik: Here you are omitting startDate which is represented by repetableParts[1], i.e. `R/<startDate>/<Duration>` or `R/<startDate>/<endDate>`. In the first if you expect startDate to be now and nothing else. But you can specify startDate and then period, which will result in the deadline date.

In the second if I think you are right here, since if we have end time as a date - we can directly use it.

tsurdilo: there is a check for repetable above ..should be

tsurdilo:
updated to add decrementing but still working on it as maciej gave me a good suggestion how to make this better.

tsurdilo:
Jenkins retest please

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/commands/ExecuteDeadlinesCommand.java
tsurdilo: @elguardian need some advice here - rescheduling always re-parses all task deadlines from scratch, in which case the repeat count even if decremented will always go back to for example 3 in "R3/...". Since you wrote the rescheduling, is is possible to get it to with using existing task deadlines rather than reparsing each time ?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
MarianMacik: Yes, but it still doesn't cover case when you have ```R/<startDate>/<Duration>``` In that case the deadline is at startDate+duration time. This code currently covers only a case when startDate is now().

tsurdilo: start date has always been "now" i didnt change that. will ask in jira if this is needed or not.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/commands/ExecuteDeadlinesCommand.java
tsurdilo: nvm :) instead of re-parsing each time we now generate all the dealines up-front for repeat count expires

tsurdilo:
@mswiderski thanks for the advice for setting up all deadlines up-front. with this rescheduling is not needed. Updated with changes. I think this should be good now except the question @MarianMacik had regarding R/<startDate>/<Duration>.

The code before always assumed that startDate is "now", I changed that to when generating all durations up-front the fist one will be from "now" and all consecutive the start date will be the previously calculated duration date. Let me know if you think this is ok or we should also support specific start date for this jira. Thanks!!

tsurdilo:
Jenkins retest please

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +220,48 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+					return new Date(fromMilli + Duration.parse(tempDateTimeStr).toMillis());
+				} else {
+					return new Date(fromMilli+ DateTimeUtils.parseDateAsDuration(tempDateTimeStr));
+				}
+			} else {
+				if (DateTimeUtils.isPeriod((durationStr))) {
+					return new Date(fromMilli + Duration.parse(durationStr).toMillis());
+				} else {
+					return new Date(fromMilli + DateTimeUtils.parseDateAsDuration(durationStr));
+				}
```
MarianMacik: Just a note that with timers we also support notation 3s###4s like delay###period, but not sure we want this here.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
MarianMacik: Yes, I know, but with ISO8601 format it is possible to define it, we already do this with timers here:
https://github.com/kiegroup/jbpm/blob/853baec3573aecf9888658a00a3d7a6b86d56700/jbpm-flow/src/main/java/org/jbpm/process/core/timer/DateTimeUtils.java#L93

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +220,48 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+		try {
+
+			long fromMilli;
+			if(fromDate != null) {
+				fromMilli = fromDate.getTime();
+			} else {
+				fromMilli = System.currentTimeMillis();
+			}
+			if (DateTimeUtils.isRepeatable(durationStr)) {
+				String[] repetableParts = DateTimeUtils.parseISORepeatable(durationStr);
```
MarianMacik: typo, should be repeatableParts

tsurdilo: fixed

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -104,33 +105,51 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
-	                escalation.setNotifications(parseNotifications(actionComponent, businessAdministrators));
-	                
-	                deadlines.add(taskDeadline);
+	            	if(businessCalendar != null) {
+						deadlines.add(getNewDeadline(expiresAt, businessCalendar.calculateBusinessTimeAsDate(expiresAt.trim()), actionComponent, businessAdministrators));
+					} else {
+						if (DateTimeUtils.isRepeatable(expiresAt.trim())) {
+							int repeatLimit = getDeadlineRepeatLimit(expiresAt.trim());
+							if(repeatLimit > 0) {
+								Date tmpDate = null;
```
mswiderski: I guess you need to set the tmpDate based on the optional start date from repeatable fom iso format

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
mswiderski: No, we don’t want that here, it’s legacy format that should no longer be used

tsurdilo:
@mswiderski yes the api pr is not needed. i remove it.

tsurdilo:
@mswiderski updated to use the optional start date for repeatable -- thanks!

tsurdilo:
test fail not related to changes ( and runs fine on local...)

tsurdilo:
Jenkins retest please

mswiderski:
@tsurdilo looks very good to me, I think we have it covered. @MarianMacik @elguardian what's your option? Is it ready to be merged?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +217,43 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+
+			if (DateTimeUtils.isRepeatable(durationStr)) {
+				String[] repeatableParts = DateTimeUtils.parseISORepeatable(durationStr);
+				String tempTimeDelay = repeatableParts[1];
+				String tempDateTimeStr = repeatableParts[2];
+
+				if (DateTimeUtils.isPeriod((tempDateTimeStr))) {
+					return new Date(DateTimeUtils.parseDateTime(tempTimeDelay) + Duration.parse(tempDateTimeStr).toMillis());
+				} else {
+					return new Date(DateTimeUtils.parseDateTime(tempTimeDelay)+ DateTimeUtils.parseDateAsDuration(tempDateTimeStr));
```
MarianMacik: This line won't work when there is `R/<duration>/<endTime>` which is also a valid use case (parseDateTime cannot parse duration). But since in case when the third part is date you don't care about the second part (whether it is duration or startTime), you can just do:
You can make it more precise by doing `OffsetDateTime.parse(tempDateTimeStr, DateTimeFormatter.ISO_DATE_TIME);` and then calling `parsedOffsetDateTime.toInstant().toEpochMilli()` which you can use in Date constructor. Right now we are parsing the middle part and then adding duration to it, which is not needed at all.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +217,43 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+				}
+			} else {
+				if (DateTimeUtils.isPeriod((durationStr))) {
+					return new Date(System.currentTimeMillis() + Duration.parse(durationStr).toMillis());
+				} else {
+					return new Date(System.currentTimeMillis() + DateTimeUtils.parseDateAsDuration(durationStr));
```
MarianMacik: Same can go here - we can directly convert date into instant, then into millis and then into date.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -101,36 +102,51 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+					} else {
+						if (DateTimeUtils.isRepeatable(expiresAt.trim())) {
+							int repeatLimit = getDeadlineRepeatLimit(expiresAt.trim());
+							if(repeatLimit > 0) {
+								for(int i=0; i < repeatLimit; i++) {
+									Date durationDate = getDeadlineDurationDate(expiresAt.trim());
+									deadlines.add(getNewDeadline(expiresAt, durationDate, actionComponent, businessAdministrators));
+								}
+							} else {
+								deadlines.add(getNewDeadline(expiresAt, getDeadlineDurationDate(expiresAt.trim()), actionComponent, businessAdministrators));
```
MarianMacik: In this case will it be repeated without limit? Do we check anywhere that if the deadline expires if we should create a new one if there is no limit for number of repetitions?

jbpm-human-task/jbpm-human-task-workitems/src/test/java/org/jbpm/services/task/wih/util/HumanTaskHandlerHelperTest.java
```diff
@@ -63,7 +65,86 @@ public void testSetDeadlinesNotStartedReassign() {
 		long expirationTime = deadlines.getStartDeadlines().get(0).getDate().getTime() - System.currentTimeMillis();
 		
 		assertEquals(4, roundExpirationTime(expirationTime));
-		
+	}
+
+	@Test
```
MarianMacik: I think we are missing test cases for:
R/<startDate>/<duration>
R/<duration>/<endDate>
R/<startDate>/<endDate>

I have only found dates and durations separately.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
tsurdilo: agree but then we would also need to add the try/catch already provided by DateTimeUtils.parseDateAsDuration. i rather reuse the existing methods i think.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
tsurdilo: agreed. there is a method DateTimeUtils.parseDateTime which does that (and works for us in case of repetable). ill just use it

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
tsurdilo: in case of just R / it should be refreshed when reschedule is called.

MarianMacik: OK and will this reschedule stop with a definite number of repetitions? Let's say I have R3/PT1H - every hour, up to 3 times - you generate 3 deadlines at once.  And after the first one is executed, reschedule happens again - it will again add 3 deadlines on top of 2 remainings?

(I have also noticed you deleted tmpDate field which means it will generate 3 identical deadlines now in case of R3)

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
MarianMacik: OK, true.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
MarianMacik: First IF handles `R/startDate/duration`
And the second IF can handle both `R/startDate/endDate` and `R/duration/endDate` since with deadlines you are only interested in `endDate`. So I would just parse the `tempDateTimeStr` as ISO Date since you "don't care" about the middle part at all.

jbpm-human-task/jbpm-human-task-workitems/src/test/java/org/jbpm/services/task/wih/util/HumanTaskHandlerHelperTest.java
tsurdilo: added

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
tsurdilo: reschedule does not keep any currently existing ones. it refreshes completely

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
tsurdilo: i updated the code just now there are checks for both and added 2 tests

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
MarianMacik: OK, so this will be then rescheduled infinitely? Each time there will be 3 deadlines created, but after the first is triggered, another 3 are created again?

tsurdilo: yeah but in that case of hard-expire you should use explicit start/end dates

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +217,51 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+					if(DateTimeUtils.isPeriod(tempTimeDelay)) {
+						return new Date(System.currentTimeMillis() + Duration.parse(tempTimeDelay).toMillis() + Duration.parse(tempDateTimeStr).toMillis());
+					} else {
+						return new Date(DateTimeUtils.parseDateTime(tempTimeDelay) + Duration.parse(tempDateTimeStr).toMillis());
+					}
+				} else {
+					if(DateTimeUtils.isPeriod(tempTimeDelay)) {
+						return new Date(System.currentTimeMillis() + Duration.parse(tempTimeDelay).toMillis() + DateTimeUtils.parseDateAsDuration(tempDateTimeStr));
+					} else {
+						return new Date(DateTimeUtils.parseDateTime(tempTimeDelay) + DateTimeUtils.parseDateAsDuration(tempDateTimeStr));
```
MarianMacik: I think we misunderstood each other here :)
Right now you also handle `R/duration/duration` which I think doesn't make sense and it is not a part of ISO8601. You can have only 4 types according to [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Time_intervals):
1. `R/startDate/duration`
1. `R/duration/endDate`
1. `R/startDate/endDate`
1. `R/duration with context information` - this is parsed by `DateTimeUtils.parseISORepeatable(durationStr);` as if it was `R/now()/duration`

So 1. and 4. was handled correctly before in the first if - if `tempDateTimeStr` is Period, we take startDate + duration as a deadline.
The 2. and 3. (`tempDateTimeStr` is not Period) can be merged into one else branch since here we are only interested in deadline - so when the interval ends, so we can just take `tempDateTimeStr` and convert it into Date.

At least this is how I understand ISO8601 and our implementation of it in timers. @mswiderski wdyt?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
MarianMacik: ok, understand. But in that case we can throw away subsequent calls - i.e. 2nd and 3rd deadline will never be executed. They will be always rewritten by the first deadline.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
elguardian: @MarianMacik correcting:
1 R/start/end 
2 R/start/duration
3 R/duration/end
4 R/duration

MarianMacik: Did I miss anything? I think I have enumerated every possibility... I think all of them make sense. In the first one there is end date as well so you can use it for deadline... Start may be specified just for users to have better context, ideally we could incorporate both start and end deadlines in one notation, but currently this is not the requirement...

tsurdilo: updated to remove the R / duration / duration option

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
tsurdilo: not sure i understand. for example

R2/2019-05-27T13:00:00Z/P1Y2M

creates 2 distinct non recurring deadlines now:

2019-05-27T13:00:00Z/P1Y2M
2020-07-01T15:30:00Z/P1Y2M

MarianMacik: well it looks in this code:

```
if(repeatLimit > 0) {
    for(int i=0; i < repeatLimit; i++) {
        Date durationDate = getDeadlineDurationDate(expiresAt.trim());
        deadlines.add(getNewDeadline(expiresAt, durationDate, actionComponent, businessAdministrators));
    }
```
that you are reusing the same durationDate over and over again. Every time it will create the same deadline.

tsurdilo: Opps yes, I didnt check in some changes. Updated and look at added tests in HumanTaskHandlerHelperTest

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +216,72 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+
+	}
+
+	protected static long repeatDuration(String durationStr, int repeatCount) {
+		long duration = Duration.parse(durationStr).toMillis();
+		for(int i = 0; i < repeatCount; i++) {
+			duration += duration;
+		}
+		return duration;
+	}
```
MarianMacik: Here when `repeatCount` is 0, we will add duration in `new Date(DateTimeUtils.parseDateTime(tempTimeDelay) + repeatDuration(tempDateTimeStr, repeatCount));` anyway. But as I wrote in the email, first trigger should occur at the beginning of the interval. Subsequent triggers should be then shifted by a duration.

So I think we need here just `return Duration.parse(durationStr).toMillis()*repeatCount`. If it is 0, it won't add anything. E.g. with R3 we will have one trigger at the startDate, then startDate+duration, then startDate+2*duration.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +216,72 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+					if(DateTimeUtils.isPeriod(tempTimeDelay)) {
+						// e.g R/duration/end
+						long firstRet = DateTimeUtils.parseDateTime(tempDateTimeStr) -  Duration.parse(tempTimeDelay).toMillis();
+						retDate =  new Date(firstRet + repeatDuration(tempTimeDelay, repeatCount));
+					} else {
+						// e.g R/start/end
+						// duration is end - start
+						long duration = DateTimeUtils.parseDateTime(tempDateTimeStr) - DateTimeUtils.parseDateTime(tempTimeDelay);
+						retDate =  new Date(DateTimeUtils.parseDateTime(tempTimeDelay) + repeatDuration(duration, repeatCount));
+					}
```
MarianMacik: This can be simplified to:

```
if (DateTimeUtils.isPeriod((tempDateTimeStr))) {
...
} else if(DateTimeUtils.isPeriod(tempTimeDelay)) {
...
} else {
...
}
```

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/HumanTaskHandlerHelper.java
```diff
@@ -201,6 +216,72 @@ public static Deadlines setDeadlines(Map<String, Object> parameters, List<Organi
...
+	protected static Date getDeadlineDurationDate(String durationStr, int repeatCount) {
+		// handles iso repetable and both period and date notation
+		try {
+
+			if (DateTimeUtils.isRepeatable(durationStr)) {
+				String[] repeatableParts = DateTimeUtils.parseISORepeatable(durationStr);
+				String tempTimeDelay = repeatableParts[1];
+				String tempDateTimeStr = repeatableParts[2];
+
+				Date retDate;
```
MarianMacik: Just curious - what does "ret" mean here?

tsurdilo:
@MarianMacik updated let me know if there is anything else

tsurdilo:
@mswiderski thanks for the help! looks good. @MarianMacik wdyt? i think we are good now wdyt?

MarianMacik:
@tsurdilo we had a discussion with @mswiderski that we should to what was agreed in the email, mainly set 1st deadline in case of R/start/period to start and not start+period.

So now it looks like this:

So now we handle timers in general (both timers and deadlines) like this: 
1. R/duration - we act as if it was R/now + duration/duration - then
first trigger is 1*duration from now, period is duration, so second trigger is another duration away
**This is what I fixed here: https://github.com/kiegroup/jbpm/pull/1493/commits/d4219e4c6cf41dd61e65ea44fae4723b921b425b#diff-90dd5fca712fe6e3e93e6ae06b85b659R87**
Since when I spoke with @mswiderski he agreed that this is a bug in timers in general - R/PT2S should fire first time after 2 seconds, not immediately.

2. R/startDate/duration - first trigger is at startDate, period is
duration

3. R/duration/endDate - first trigger is at endDate-duration, period is
duration

4. R/startDate/endDate - first trigger is at startDate, period is
endDate-startDate

I have also enhanced test coverage in `HumanTaskHandlerHelperTest` by asserting on all created deadlines in case they are bounded repeatable ones.

tsurdilo:
@MarianMacik looked at your updates - look greats. thanks

tsurdilo:
Jenkins retest please

tsurdilo:
Jenkins retest please

tsurdilo:
@MarianMacik @mswiderski so are we good to merge this now ??

MarianMacik:
From my side yes.

mswiderski:
@tsurdilo @MarianMacik could one of you file a docs jira with description of what this brings and what it removes (the reschedule for period based deadlines)?

great work all! thanks

tsurdilo:
@mswiderski -- https://issues.jboss.org/browse/JBPM-8567. will merge this one then. thanks all

PR #1494, opened 2019-05-16
Allow extensibility in the process builder/validator 

SimonRinguette:
These 2 changes allows us to have extensibility options to keep node information with the error message for builder/validation.

Our tool highlights the shape in error after a publication and those simple modifications let us hook into the framework and retrieve the info we need in the fronted.

The default behavior of the engine is not changed at all.

ProcessBuilderImpl: Icreated a method to transfer the error into ParseEerror so that we can overload it to provide our own class that extends ParserError and keep the node information

RuleFlowProcessValidator: The message formating hardcoding is moved to the new class that extends the previous class. The non formatted message is also accessible as well as the node.


SimonRinguette:
@mswiderski let me know if you don't like this

mswiderski:
Ok to test

mswiderski:
@SimonRinguette looks good, let me try it out tomorrow

mswiderski:
thanks @SimonRinguette

PR #1501, opened 2019-06-17
JBPM-8485 - Add support ISO8601 expressions for user task notificatio…

MarianMacik:
…ns - reschedule only if deadlines not empty

@martinweiler @tsurdilo Please have a look. It fixes the added test, which is otherwise failing.

The issue was that when the first deadline of R3/PT2S was triggered, then it again parsed this String etc. and since we don't reschedule R3/PT2S - only R/ is rescheduled, no deadlines were find, but the attempt to scheduling was made. This attempt will iterate over all deadlines and schedule ones which are not escalated, which means the 2nd and 3rd are scheduled again, leaving 2 timers for 2nd and 3rd deadlines. When 2nd deadlines timers are executed (now there is 2 of them), each of them will create another timer for 3rd deadline. So for 3rd execution, there will be 4 timers.

https://github.com/kiegroup/jbpm/blob/cd57cd17925f37e9abe9188e5bff44126464c619/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/util/DeadlineSchedulerHelper.java#L84-L92

This fix simply won't reschedule if no deadlines are set by `HumanTaskHandlerHelper#setDeadlines`. This should work since we currently support either more bounded deadlines (setDeadlines will always return no deadlines) or just one unbounded (setDeadlines will always return only the deadline being executed).


tsurdilo:
looks good to me. thanks

PR #1504, opened 2019-06-18
JBPM-8485 - backwards compat for simple format

tsurdilo:


tsurdilo:
@martinweiler hi, would it be possible to apply this pr locally and test your process with it (from the email thread). Thanks!

tsurdilo:
removing pr - maciej said we should not do the backwards compat.

PR #1505, opened 2019-06-21
JBPM-8515: Error displayed if you try to Trigger task inside of sub-p…

nmirasch:
…rocess

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
@@ -116,4 +116,22 @@ public class WorkflowProcessImpl extends ProcessImpl implements WorkflowProcess,
             }
         }
     }
+
+    protected Node getContainerNode(Node currentNode, org.jbpm.workflow.core.NodeContainer nodeContainer, long nodeId) {
+        for (Node node : nodeContainer.getNodes()) {
+            if (nodeId == node.getId()) {
+                return currentNode;
+            } else {
+                if (node instanceof org.jbpm.workflow.core.NodeContainer) {
+                    return getContainerNode(node, (org.jbpm.workflow.core.NodeContainer) node, nodeId);
+                }
+            }
+        }
+        return null;
+    }
+
+    public Node getParentNode(long nodeId) {
+        return getContainerNode(null, nodeContainer, nodeId);
+    }
+
 }
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
@@ -19,6 +19,7 @@ package org.jbpm.kie.services.impl.admin.commands;
 import org.jbpm.ruleflow.instance.RuleFlowProcessInstance;
 import org.jbpm.services.api.NodeNotFoundException;
 import org.jbpm.services.api.ProcessInstanceNotFoundException;
+import org.jbpm.workflow.instance.NodeInstanceContainer;
 import org.kie.api.command.ExecutableCommand;
 import org.kie.api.definition.process.Node;
 import org.kie.api.runtime.Context;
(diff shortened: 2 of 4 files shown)


nmirasch:
@mswiderski @MarianMacik  I didn't find a better way to get the parent node (the container) and I've created the recursive function getParentNode at WorkflowProcess. If you know a better way to get that node would be very welcome.
Could you take a look on this? Thanks!

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
```diff
@@ -43,28 +44,40 @@ public TriggerNodeCommand(long processInstanceId, long nodeId) {
...
+        try {
+            node = wfp.getRuleFlowProcess().getNode(nodeId);
+        } catch (Exception e) {
+
+        }
+
+        if (node == null) { //search container node
+            node = wfp.getRuleFlowProcess().getNodesRecursively().stream().filter(ni -> ni.getId() == nodeId).findFirst().orElse(null);
+            if (node == null) {
+                throw new NodeNotFoundException("Node instance with id " + nodeId + " not found");
```
mswiderski: message is bit misleading as it is node and not node instance that is not found

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
```diff
@@ -43,28 +44,40 @@ public TriggerNodeCommand(long processInstanceId, long nodeId) {
...
-        Node node = wfp.getRuleFlowProcess().getNodesRecursively().stream().filter(ni -> ni.getId() == nodeId).findFirst().orElse(null);
-        if (node == null) {
-            throw new NodeNotFoundException("Node instance with id " + nodeId + " not found");
+
+        Node node = null;
+        try {
+            node = wfp.getRuleFlowProcess().getNode(nodeId);
+        } catch (Exception e) {
+
+        }
```
mswiderski: why do we have empty catch here?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
```diff
@@ -43,28 +44,40 @@ public TriggerNodeCommand(long processInstanceId, long nodeId) {
...
+
+        if (node == null) { //search container node
+            node = wfp.getRuleFlowProcess().getNodesRecursively().stream().filter(ni -> ni.getId() == nodeId).findFirst().orElse(null);
+            if (node == null) {
+                throw new NodeNotFoundException("Node instance with id " + nodeId + " not found");
+            }
+            CompositeNodeInstance nodeInstanceContainerNode =
+                    ((CompositeNodeInstance) wfp.getNodeInstance(wfp.getRuleFlowProcess().getParentNode(nodeId)));
```
mswiderski: we sure it's always CompositeNodeInstance? or shall we make it NodeContainer as the variable name suggests?

jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
```diff
@@ -116,4 +116,30 @@ protected void processNodeContainer(org.jbpm.workflow.core.NodeContainer nodeCon
             }
         }
     }
+
+    protected void getParentNodes(Node currentNode, org.jbpm.workflow.core.NodeContainer nodeContainer, List<Node> nodes, long nodeId) {
```
mswiderski: why this is not returning found node and instead populates the list? Asking as it seems this just stops when node is found so wouldn't it be better to return found node so the list won't be needed at all?

jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
```diff
@@ -116,4 +116,30 @@ protected void processNodeContainer(org.jbpm.workflow.core.NodeContainer nodeCon
...
+        }
+    }
+
+    public Node getParentNode(long nodeId) {
+        List<Node> nodes = new ArrayList<>();
+
+        getParentNodes(null, nodeContainer, nodes, nodeId);
+
+        if (nodes.size() > 0) {
+            return nodes.get(0);
```
mswiderski: how do we know it's the right node? :)

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
nmirasch: @mswiderski agree :S, updated

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
nmirasch: @mswiderski  It was empty because even the node is not found at the same level and an exception is thrown, the search of it have to continue deeper. Now I added a log message there.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
nmirasch: @mswiderski  Nop, you are absolutely right I changed that to a generic NodeInstanceContainer and it works in the same way. Better now, before it was to tied to the examples I tried, Thanks!

jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
nmirasch: Yes I changed the function a bit to stop when it find a node that accomplish the condition and return that node and the list is not needed at all :)

jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
nmirasch: Just picking the first ? :) Anyway, now the function return the first node that contain a node with the passed nodeId. I hope that would be the good one..

nmirasch:
@mswiderski Thanks for reviewing. I updated the code, let's see if it looks better now. Could you re-review?

jbpm-flow/src/main/java/org/jbpm/workflow/core/impl/WorkflowProcessImpl.java
```diff
@@ -116,4 +116,24 @@ protected void processNodeContainer(org.jbpm.workflow.core.NodeContainer nodeCon
             }
         }
     }
+
+    protected Node getContainerNode(Node currentNode, org.jbpm.workflow.core.NodeContainer nodeContainer, long nodeId) {
+        for (Node node : nodeContainer.getNodes()) {
+            if (nodeId == node.getId()) {
+                if (currentNode != null) {
+                    return currentNode;
```
MarianMacik: Maybe just one thing when I looked at this the second time. Is this check really needed? If the `currentNode` is `null`, then we can immediately return `null` because given that nodeId is unique, it won't match any of these conditions next time, so it will eventually reach `return null;`.

nmirasch:
@MarianMacik  Absolutely agree, updated. Thanks!

MarianMacik:
Looks good now. It's good to merge.

nmirasch:
Jenkins retest this please

nmirasch:
Jenkins retest this please

nmirasch:
rebased!

nmirasch:
Jenkins retest this please

nmirasch:
Jenkins retest this please

nmirasch:
rebased.

cristianonicolai:
failures look unrelated to changes.

PR #1509, opened 2019-07-03
RHPAM-2198 - NullPointerException in org.drools.core.marshalling.impl…

mswiderski:
….ProtobufOutputMarshaller.writeTimers in a multi-node setup

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -34,6 +34,7 @@ import javax.ejb.Startup;
 import javax.ejb.Timeout;
 import javax.ejb.Timer;
 import javax.ejb.TimerConfig;
+import javax.ejb.TimerService;
 
 import org.drools.core.time.JobHandle;
 import org.drools.core.time.impl.TimerJobInstance;
@@ -54,7 +55,7 @@ public class EJBTimerScheduler {
 	private ConcurrentMap<String, TimerJobInstance> localCache = new ConcurrentHashMap<String, TimerJobInstance>();
 	
 	@Resource
-	private javax.ejb.TimerService timerService;
+	protected javax.ejb.TimerService timerService;
 	
 	@PostConstruct
 	public void setup() {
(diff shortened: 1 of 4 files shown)


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -153,10 +154,11 @@ public TimerJobInstance getTimerByName(String jobName) {
...
     				
     				EjbGlobalJobHandle handle = (EjbGlobalJobHandle) job.getTimerJobInstance().getJobHandle();
-    				localCache.putIfAbsent(jobName, handle.getTimerJobInstance());
-    				if (handle.getUuid().equals(jobName)) {
-    					logger.debug("Job  {} does match timer and is going to be returned", jobName);
+    				
+    				if (handle.getUuid().equals(jobName)) {    					
     					found = handle.getTimerJobInstance();
+    					localCache.putIfAbsent(jobName, found);
+    					logger.debug("Job {} does match timer and is going to be returned {}", jobName, found);
```
MarianMacik: Since we have found it and put it in the cache, we can directly return it and avoid iterating over other timers, can't we?

mswiderski: good point, updated

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -153,10 +154,13 @@ public TimerJobInstance getTimerByName(String jobName) {
     				EjbTimerJob job = (EjbTimerJob) info;
     				
     				EjbGlobalJobHandle handle = (EjbGlobalJobHandle) job.getTimerJobInstance().getJobHandle();
-    				localCache.putIfAbsent(jobName, handle.getTimerJobInstance());
-    				if (handle.getUuid().equals(jobName)) {
-    					logger.debug("Job  {} does match timer and is going to be returned", jobName);
+    				
+    				if (handle.getUuid().equals(jobName)) {    					
     					found = handle.getTimerJobInstance();
+    					localCache.putIfAbsent(jobName, found);
```
gmunozfe: I was wondering about this scenario:
if the cache already contains a key with the same jobName, currently is skipping the insertion (because it's considered to be the same value associated to that key), but if it's not the case -not sure if the scenario is possible-, a warning log may help to detect it.

```
TimerJobInstance tjb = localCache.putIfAbsent(jobName, handle.getTimerJobInstance());
if (!tjb.equals(handle.getTimerJobInstance()))
    logger.warn("Job {} already contained value {} and won't be override with {}", jobName, tjb, handle.getTimerJobInstance());
```

mswiderski: this should not be a case as each timer must be uniquely identified

gmunozfe: thanks, then looks good to me

MarianMacik:
7.18.x branch is no longer maintained

PR #1516, opened 2019-07-18
JBPM-8655: Task operations are slower after a join was introduced in …

nmirasch:
…"getTaskInstanceById" query

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
@@ -811,6 +811,17 @@ public class RuntimeDataServiceImpl implements RuntimeDataService, DeploymentEve
         return null;
 	}
 
+    @Override
+    public UserTaskInstanceDesc getTaskById(Long taskId, boolean withSLA){
+        UserTaskInstanceDesc taskInstanceDesc = getTaskById(taskId);
+        if(withSLA && (taskInstanceDesc != null && taskInstanceDesc.getWorkItemId() != null)) {
+            NodeInstanceDesc nodeInstanceDesc = getNodeInstanceForWorkItem(taskInstanceDesc.getWorkItemId());
+            taskInstanceDesc.setSlaCompliance(nodeInstanceDesc.getSlaCompliance());
+            taskInstanceDesc.setSlaDueDate(nodeInstanceDesc.getSlaDueDate());
+        }
+        return taskInstanceDesc;
+    }
+
 	@Override
 	public List<TaskSummary> getTasksAssignedAsBusinessAdministrator(String userId, QueryFilter filter) {
 	
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/mapper/UserTaskInstanceQueryMapper.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/mapper/UserTaskInstanceQueryMapper.java
@@ -78,9 +78,7 @@ public class UserTaskInstanceQueryMapper extends AbstractQueryMapper<UserTaskIns
                 getColumnLongValue(dataSetResult, COLUMN_TASK_PROCESSINSTANCEID, index),//processInstanceId,
                 getColumnDateValue(dataSetResult, COLUMN_CREATEDON, index),//createdOn,
                 getColumnDateValue(dataSetResult, COLUMN_DUEDATE, index),//dueDate
-                getColumnLongValue(dataSetResult, COLUMN_WORKITEMID, index),//workItemId
-                getColumnDateValue(dataSetResult, COLUMN_TASK_SLA_DUE_DATE, index),//slaDueDate
-                getColumnIntValue(dataSetResult, COLUMN_TASK_SLA_COMPLIANCE, index)//slaCompliance
+                getColumnLongValue(dataSetResult, COLUMN_WORKITEMID, index)//workItemId
                 );
         return userTask;
     }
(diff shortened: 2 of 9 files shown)


cristianonicolai:
jenkins please retest this

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -811,6 +811,25 @@ public UserTaskInstanceDesc getTaskById(Long taskId) {
         return null;
 	}
 
+    @Override
+    public UserTaskInstanceDesc getTaskByIdWithSLA(Long taskId) {
+        Map<String, Object> params = new HashMap<String, Object>();
+        params.put("taskId", taskId);
+        params.put("maxResults", 1);
+        List<UserTaskInstanceDesc> tasks = commandService.execute(
+                new QueryNameCommand<List<UserTaskInstanceDesc>>("getTaskInstanceById", params));
```
MarianMacik: This looks like an exact copy of `getTaskById`. Can we perhaps reuse that method in this new one?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/mapper/UserTaskInstanceQueryMapper.java
```diff
@@ -78,9 +78,7 @@ protected UserTaskInstanceDesc buildInstance(DataSet dataSetResult, int index) {
                 getColumnLongValue(dataSetResult, COLUMN_TASK_PROCESSINSTANCEID, index),//processInstanceId,
                 getColumnDateValue(dataSetResult, COLUMN_CREATEDON, index),//createdOn,
                 getColumnDateValue(dataSetResult, COLUMN_DUEDATE, index),//dueDate
-                getColumnLongValue(dataSetResult, COLUMN_WORKITEMID, index),//workItemId
-                getColumnDateValue(dataSetResult, COLUMN_TASK_SLA_DUE_DATE, index),//slaDueDate
-                getColumnIntValue(dataSetResult, COLUMN_TASK_SLA_COMPLIANCE, index)//slaCompliance
+                getColumnLongValue(dataSetResult, COLUMN_WORKITEMID, index)//workItemId
```
MarianMacik: Was this needed even before this change? This is only used for custom queries and since this mapper is aimed to be used only with "standard" UserTaskInstanceDesc fields which come from AuditTaskImpl I guess, it is OK. What if a user wants to obtain also SLA using custom queries? Then they would have to create their own mapper? I am fine with that, just asking.

nmirasch: @MarianMacik the purpose of letting this workitemId is becasuse we need it to generate de second query in case the SLA data is requested.

MarianMacik: Yeah, I didn't question the workitemId, but the other SLA fields. So I guess when somebody uses this in a custom query, they can get SLA data in a separate query based on workitemId or even create a mapper like it was before, including the SLA fields, right?

nmirasch: @MarianMacik  maybe I did it wrongly, but here I try to have at least the 'standard' mapping used by the getTaskById, but you are absolutely right they have to rely on their own mapper of the custom query. Would you suggest let this as it was then? I can remove also the workitemId. Makes sense?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
nmirasch: @MarianMacik yes, I updated the PR reusing getTaskById. Thanks!

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -811,6 +811,17 @@ public UserTaskInstanceDesc getTaskById(Long taskId) {
         return null;
 	}
 
+    @Override
+    public UserTaskInstanceDesc getTaskByIdWithSLA(Long taskId) {
+        UserTaskInstanceDesc taskInstanceDesc = getTaskById(taskId);
```
MarianMacik: I believe this can return `null` so the next line can fail. So we need a null check first.

nmirasch: @MarianMacik  ups, sorry too fast, Thanks.
updated.

nmirasch:
jenkins please retest this

nmirasch:
Jenkins retest this please

cristianonicolai:
jenkins please retest this

cristianonicolai:
jenkins please retest this

cristianonicolai:
@mswiderski can you take a quick look on this, thanks

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/RuntimeDataServiceImpl.java
```diff
@@ -811,6 +811,17 @@ public UserTaskInstanceDesc getTaskById(Long taskId) {
         return null;
 	}
 
+    @Override
+    public UserTaskInstanceDesc getTaskByIdWithSLA(Long taskId) {
```
cristianonicolai: @nmirasch perhaps to the same here as in the kie-server, same method name with boolean withSLA? That way you could reuse the method again with a default of false.

nmirasch: @cristianonicolai updated, Thanks!

cristianonicolai:
jenkins please retest this

PR #1520, opened 2019-08-02
[JBPM-8688] Timers do not recover after database disconnection

elguardian:
providing recovery of timers. executed till it is succeed

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -17,6 +17,8 @@
 package org.jbpm.services.ejb.timer;
 
 import java.io.Serializable;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoUnit;
 import java.util.Date;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentHashMap;
@@ -34,7 +36,10 @@ import javax.ejb.Startup;
 import javax.ejb.Timeout;
 import javax.ejb.Timer;
 import javax.ejb.TimerConfig;
-import javax.ejb.TimerService;
+import javax.ejb.TransactionManagement;
+import javax.ejb.TransactionManagementType;
+import javax.transaction.Status;
+import javax.transaction.UserTransaction;
 
 import org.drools.core.time.JobHandle;
 import org.drools.core.time.impl.TimerJobInstance;
@@ -45,10 +50,22 @@ import org.slf4j.LoggerFactory;
 @Singleton
 @Startup
 @ConcurrencyManagement(ConcurrencyManagementType.CONTAINER)
+@TransactionManagement(TransactionManagementType.BEAN)
 @Lock(LockType.READ)
 public class EJBTimerScheduler {
 
 	private static final Logger logger = LoggerFactory.getLogger(EJBTimerScheduler.class);
+
+    private enum TimerExceptionPolicy {
+        RETRY,
+        PLATFORM
+    };
+
+    private static final Long TIMER_RETRY_INTERVAL = Long.parseLong(System.getProperty("org.kie.jbpm.timer.retry.interval", "5000"));
+
+    private static final Integer TIMER_RETRY_LIMIT = Integer.parseInt(System.getProperty("org.kie.jbpm.timer.retry.limit", "3"));
+
+    private static final TimerExceptionPolicy TIMER_RETRY_POLICY = Enum.valueOf(TimerExceptionPolicy.class, System.getProperty("org.kie.jbpm.timer.retry.policy", "PLATFORM"));
 
 	private static final Integer OVERDUE_WAIT_TIME = Integer.parseInt(System.getProperty("org.jbpm.overdue.timer.wait", "20000"));
 
@@ -59,6 +76,10 @@ public class EJBTimerScheduler {
 	@Resource
 	protected javax.ejb.TimerService timerService;
 
+    @Resource
+    protected UserTransaction utx;
+
+
 	@PostConstruct
 	public void setup() {
 	    // disable auto init of timers since ejb timer service supports persistence of timers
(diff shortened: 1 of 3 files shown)


elguardian:
@mswiderski can you have a look if this goes in the right direction ?

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -59,46 +78,116 @@
 	@Resource
 	protected javax.ejb.TimerService timerService;
 
+    @Resource
+    protected javax.ejb.SessionContext sessionContext;
+
+    @Resource
+    protected TransactionSynchronizationRegistry tsr;
```
mswiderski: this seems to be not used

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -59,46 +78,116 @@
 	@Resource
 	protected javax.ejb.TimerService timerService;
 
+    @Resource
+    protected javax.ejb.SessionContext sessionContext;
```
mswiderski: maybe directly inject UserTransaction instead of session context...

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -45,11 +51,24 @@
...
+@TransactionManagement(TransactionManagementType.BEAN)
 @Lock(LockType.READ)
 public class EJBTimerScheduler {
 
 	private static final Logger logger = LoggerFactory.getLogger(EJBTimerScheduler.class);
 
+    static private enum TimerExceptionPolicy {
+        RETRY,
+        IGNORE,
+        PLATFORM
```
mswiderski: do we really need to have the policy here? what would be the need for IGNORE or PLATFORM? And what PLATFORM means ...

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -59,46 +78,116 @@
...
+                    transaction(operation, ejbTimerJob.getTimerJobInstance());
+                } catch (Exception e1) {
+                    logger.error("Failed to executed timer recovery {}", e1.getMessage(), e1);
+                }
+                break;
+            case IGNORE:
+                logger.warn("Execution of time failed. The timer won't be retried {}", ejbTimerJob.getTimerJobInstance());
+                break;
+            case PLATFORM:
+                logger.warn("Execution of time failed. Application server policy applied {}", ejbTimerJob.getTimerJobInstance());
```
mswiderski: does this differ between app servers? we use WAS and WLS with EJB timer service as far as I know.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: PLATFORM is dependant. The spec says that it will retry at least one, but that is app to app server to decide. in wildfly case it will retry once right after failure.
IGNORE is just it fails, don't try anymore.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: it might be... I don't know the detail of those app servers. 
for instance in WAS you can configure the numbers of retries
https://www.ibm.com/support/knowledgecenter/en/SSEQTP_9.0.5/com.ibm.websphere.base.doc/ae/uejb_timerservice.html

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
mswiderski: I'd remove the IGNORE one and default to PLATFORM as you did already

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
mswiderski: does EAP has this configurable as well?

elguardian: nop.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: done

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: done

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: remove IGNORE

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
mswiderski: I still can’t find any use of the tsr... maybe it should be removed if it is not used?

mswiderski:
@MarianMacik could you please have a look as well?

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -45,11 +51,23 @@
 @Singleton
 @Startup
 @ConcurrencyManagement(ConcurrencyManagementType.CONTAINER)
+@TransactionManagement(TransactionManagementType.BEAN)
 @Lock(LockType.READ)
 public class EJBTimerScheduler {
 
 	private static final Logger logger = LoggerFactory.getLogger(EJBTimerScheduler.class);
 
+    static private enum TimerExceptionPolicy {
```
MarianMacik: static is redundant here :)

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
MarianMacik: I can confirm this is not used at all.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -59,46 +77,120 @@
...
+                logger.warn("Execution of time failed. Application server policy applied {}", ejbTimerJob.getTimerJobInstance());
+                throw new RuntimeException(e);
+        }
+    }
+
+    @FunctionalInterface
+    private interface Transaction<I> {
+
+        void doWork(I item) throws Exception;
+    }
```
MarianMacik: Do we need another functional interface? What about using a simple Consumer here?

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -59,46 +77,120 @@
...
+            throw e;
+        }
+    }
+
+    private void recoverTimerJobInstance(EjbTimerJob ejbTimerJob, Exception e) {
+        // if we have next date fired means that it would have been reescheduled already by DefaultTimerJobInstance
+        if (ejbTimerJob.getTimerJobInstance().getTrigger().hasNextFireTime() != null) {
+            logger.warn("Execution of time failed Interval Trigger failed {}", ejbTimerJob.getTimerJobInstance());
+            return;
+        }
```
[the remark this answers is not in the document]
elguardian: This need some thought. I didn't allow that because overlapping. It gets a bit messy with for instance timer start events and I don't think add any value on these cases.
Also it would be very complicated to end user to understand what is going on.  It is not easy to follow through logging what is failing and what is not. 
Also the danger of increasing dramatically the number of triggers. if there is a network glitch like the one we are trying to solve here and we have interval triggers it will end up in multiple retrys
1) 1 interval trigger
2) retry #1 1 interval trigger
2) 2 interval trigger
3) retry #2 1 interval trigger 
4) 3 interval trigger
4) retry #3 1 interval trigger
4) retry #1 2 interval trigger

in the 5 step would be already a large amount of timers in the db.
and so on... this would grow exponentially.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: AFAIF consumer does not allow to throw an exception. I didn't see any functional interface allowing that.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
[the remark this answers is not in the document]
MarianMacik: Well, will it really look like you wrote? I think that if the first trigger is not triggered because of an exception and let's say we implemented the retry also for reoccurring timers, then it won't fire `2 interval trigger` because the `1 interval trigger` did not succeed, which means it will first try to trigger the first instance and only after that succeeds it will move on to the second one. But I understand your point that we cannot save everything from application itself :) Maybe a little paragraph in documentation about how this behaves in such situation would be helpful for users.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
MarianMacik: That's true, didn't notice that, sorry.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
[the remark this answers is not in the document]
elguardian: @MarianMacik  interval triggers by definition happens at fixed time so there is policy regarding delaying the next trigger because previous one has failed. At least I didn't see it that way.
your are presuming that way of work. let see that in case of interval trigger it won't retry the timer. I don't see any usefulness to that.

MarianMacik:
@mswiderski is this ok to merge?

elguardian:
@MarianMacik this was approved. any progress

PR #1521, opened 2019-08-05
Fix WebServiceWorkitemHandler with JDK 11

MarianMacik:
@gmunozfe There was an issue on JDK11 in jbpm-container tests. An upgrade of cxf-rt-databinding-jaxb no longer depends on jaxb-impl (see https://github.com/apache/cxf/commit/54b71171e89bb1be7d169ac91bb626c9c1ca293b#diff-3ab950a3633cee6bcef4ee7e98da8d33), so I had to add it in jbpm-container test suite. Moreover, we don't need jaxb-core there since cxf-rt-databinding-jaxb no longer depends on it and tests are passing (at least locally).

### DIFF
--- a/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/pom.xml
+++ b/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/pom.xml
@@ -350,6 +350,13 @@
       <scope>test</scope>
     </dependency>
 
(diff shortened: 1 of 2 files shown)


MarianMacik:

MarianMacik:
jenkins retest this

MarianMacik:
jenkins retest this

PR #1524, opened 2019-08-07
BAQE-1049 Measure cross-module test coverage

rsynek:
Depends on kiegroup/droolsjbpm-build-bootstrap#1024


rsynek:
The test failure does not seem to be connected with this change.

PR #1525, opened 2019-08-08
[JBPM-8694] ConditionalEvent isn't triggered when process id contains hyphen

elguardian:
removing the last 2 hyphens so you get the right conditional event.

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
@@ -36,7 +36,6 @@ import org.drools.core.phreak.PropagationEntry;
 import org.drools.core.time.TimeUtils;
 import org.drools.core.time.TimerService;
 import org.drools.core.time.impl.CommandServiceTimerJobFactoryManager;
-import org.drools.core.time.impl.CronExpression;
 import org.drools.core.time.impl.ThreadSafeTrackableTimeJobFactoryManager;
 import org.jbpm.process.core.event.EventFilter;
 import org.jbpm.process.core.event.EventTransformer;
@@ -74,6 +73,9 @@ import org.kie.internal.runtime.StatefulKnowledgeSession;
 import org.kie.internal.runtime.manager.context.CaseContext;
 import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
 import org.kie.internal.utils.CompositeClassLoader;
+
+import static org.drools.core.time.TimeUtils.parseTimeString;
+import static org.kie.soup.commons.cron.CronExpression.isValidExpression;
 
 public class ProcessRuntimeImpl implements InternalProcessRuntime {
 	
@@ -425,13 +427,11 @@ public class ProcessRuntimeImpl implements InternalProcessRuntime {
                     // new activations of the rule associate with a state node
                     // signal process instances of that state node
                     String ruleName = event.getMatch().getRule().getName();
-                    if ( ruleName.startsWith( "RuleFlowStateNode-" )) {
-                        int index = ruleName.indexOf( '-',
-                                                      18 );
-                        index = ruleName.indexOf( '-',
-                                                  index + 1 );
-                        String eventType = ruleName.substring( 0,
-                                                               index );
+                    String rulePrefix = "RuleFlowStateNode-";
+                    if (ruleName.startsWith(rulePrefix)) {
+                        int index = ruleName.lastIndexOf('-');
+                        index = ruleName.lastIndexOf('-', index - 1);
+                        String eventType = ruleName.substring(0, index);
 
                         kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(eventType, event));
                     } else if (ruleName.startsWith( "RuleFlowStateEventSubProcess-" ) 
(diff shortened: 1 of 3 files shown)


elguardian:
this is ok to test.

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -425,13 +425,13 @@ public void matchCreated(MatchCreatedEvent event) {
...
-                                                  index + 1 );
-                        String eventType = ruleName.substring( 0,
-                                                               index );
+                    String rulePrefix = "RuleFlowStateNode-";
+                    if (ruleName.startsWith(rulePrefix)) {
+                        int index = ruleName.indexOf('-', rulePrefix.length() + 1);
+                        int numberOfHyphens = (int) ruleName.chars().filter(ch -> ch == '-').count();
+                        // remove the last 2
+                        index = ordinalIndexOf(ruleName, '-', numberOfHyphens - 1);
+                        String eventType = ruleName.substring(0, index);
```
mswiderski: could you please explain what's the intension here?

and is this backward compatible ...

elguardian: @mswiderski  yep. Problem is that you don't know how many hyphens has the process id. But you know how many there are at the end. So you need to remove the last two. and then you get the eventType. It should be backward compatible. 
rule + processId+ uniqueId +....

mswiderski: as long as this is backward compatible then I am fine with it

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -425,13 +425,13 @@ public void matchCreated(MatchCreatedEvent event) {
...
-                    if ( ruleName.startsWith( "RuleFlowStateNode-" )) {
-                        int index = ruleName.indexOf( '-',
-                                                      18 );
-                        index = ruleName.indexOf( '-',
-                                                  index + 1 );
-                        String eventType = ruleName.substring( 0,
-                                                               index );
+                    String rulePrefix = "RuleFlowStateNode-";
+                    if (ruleName.startsWith(rulePrefix)) {
+                        int index = ruleName.indexOf('-', rulePrefix.length() + 1);
```
MarianMacik: This is not used at all since this is overwritten on line 433.

Can you also please further describe why you have to remove last 2 hyphens? Thanks.

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -425,13 +425,12 @@ public void matchCreated(MatchCreatedEvent event) {
...
-                        int index = ruleName.indexOf( '-',
-                                                      18 );
-                        index = ruleName.indexOf( '-',
-                                                  index + 1 );
-                        String eventType = ruleName.substring( 0,
-                                                               index );
+                    String rulePrefix = "RuleFlowStateNode-";
+                    if (ruleName.startsWith(rulePrefix)) {
+                        int numberOfHyphens = (int) ruleName.chars().filter(ch -> ch == '-').count();
+                        // remove the last 2
```
MarianMacik: If we want to remove last 2 hyphens, why not to use just `String#lastIndexOf` method twice or in a loop? This can be much quicker in case there are many hyphens in the beginning.

And btw why do you have to remove last 2 hyphens exactly? Can you please explain? Thanks.

elguardian: will change that.

Regarding the last 2 hyphens is more complicated. The rule name is a composition of (only in case of state node, check ProcessBuilderImpl where the rule is created.

RuleFlowStateNode-Intermediate-Catch-Event-2 -3 -DROOLS_DEFAULT
RuleFlowStateNode-" + process.getId() + "-" + state.getUniqueId() + "-" + key.getNodeId() + "-" + key.getToType() 

The activatioy type is  rule + processID + node uniqueId (this is general for every rule)

MarianMacik: Thanks, now it looks good!

MarianMacik:
@mswiderski can you approve as well?

elguardian:
@mswiderski 
AsyncContinuationSupportTest is fixed in antoher PR.
the other two are unrelated. @MarianMacik I don't know if they are recurrent, if they are if there is a jira I can have a look

MarianMacik:
jenkins retest this

elguardian:
Jenkins retest this please.

elguardian:
jenkins retest this

elguardian:
jenkins retest this

elguardian:
@MarianMacik have a look. did the changes requested but for some reason the build is staled.

MarianMacik:
I have reopened the PR, we will see if Jenkins picks it up.

elguardian:
jenkins retest this

elguardian:
@MarianMacik test failures are unrelated. Something regarding DMN ?

MarianMacik:
jenkins retest this

PR #1528, opened 2019-08-20
[JBPM-8644] AsyncMode doesn't wait on Inclusive converging Gateway

elguardian:
setting the proper level when we are moving backward in the flow
so it is not interpreted incorrectly by the iteration level.

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
@@ -93,8 +93,10 @@ public class AsyncEventNodeInstance extends EventNodeInstance {
     @Override
     protected NodeInstance getFrom() {
         AsyncEventNode node = (AsyncEventNode) getNode();
+        int level = getLevel();
         org.jbpm.workflow.instance.NodeInstance instance = ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).getNodeInstance(node.getPreviousNode());
         ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).removeNodeInstance(instance);
+        ((org.jbpm.workflow.instance.impl.NodeInstanceImpl) instance).setLevel(level);
         return instance;
     }
 
--- a/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/AsyncContinuationSupportTest.java
+++ b/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/AsyncContinuationSupportTest.java
@@ -1208,7 +1208,7 @@ public class AsyncContinuationSupportTest extends AbstractExecutorBaseTest {
     @Test(timeout = 10000)
     public void testAsyncModeWithInclusiveGateway() throws Exception {
         // JBPM-7414 // there are two paths for end process , therefore it is executed twice
-        final NodeLeftCountDownProcessEventListener countDownListener = new NodeLeftCountDownProcessEventListener("EndProcess", 2);
+        final NodeLeftCountDownProcessEventListener countDownListener = new NodeLeftCountDownProcessEventListener("EndProcess", 1);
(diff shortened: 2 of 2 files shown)


jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/AsyncContinuationSupportTest.java
```diff
@@ -1208,7 +1208,7 @@ public void testAsyncModeWithSignalEventSubProcess() throws Exception {
     @Test(timeout = 10000)
     public void testAsyncModeWithInclusiveGateway() throws Exception {
         // JBPM-7414 // there are two paths for end process , therefore it is executed twice
-        final NodeLeftCountDownProcessEventListener countDownListener = new NodeLeftCountDownProcessEventListener("EndProcess", 2);
+        final NodeLeftCountDownProcessEventListener countDownListener = new NodeLeftCountDownProcessEventListener("EndProcess", 1);
```
elguardian: changing this was incorrect. see this 
https://github.com/kiegroup/jbpm/pull/1270/files

elguardian:
@gmunozfe have a look

gmunozfe:
looks good to me: now, test succeeds also in this scenario:
Script Task-1
Executing work item WorkItem 1 [name=Rest, state=0, processInstanceId=1, parameters{Method=GET, TaskName=Rest, Url=https://www.google.com}]
Script Task-2
Script Task-3

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
```diff
@@ -93,8 +93,10 @@ public Node getNode() {
     @Override
     protected NodeInstance getFrom() {
         AsyncEventNode node = (AsyncEventNode) getNode();
+        int level = getLevel();
         org.jbpm.workflow.instance.NodeInstance instance = ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).getNodeInstance(node.getPreviousNode());
         ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).removeNodeInstance(instance);
+        ((org.jbpm.workflow.instance.impl.NodeInstanceImpl) instance).setLevel(level);
```
MarianMacik: Does it have to be split in 2 commands for some reason?

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/impl/wih/AsyncContinuationSupportTest.java
MarianMacik: @elguardian hmm, looks obvious now :) Thanks for having a look at it.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
elguardian: I don't understand what you mean :/ can you be more specific ?. There is no way to change removeInstance and not affect other parts of the code.

MarianMacik: Sorry, GitHub cut out one line, so the context is not full :)
I think you can do just
`((org.jbpm.workflow.instance.impl.NodeInstanceImpl) instance).setLevel(getLevel());`
I am not saying it's wrong, it is just a matter of preference, but I was curious if there is any side-effect because of which it has to be split into 2 separate statements.

elguardian: when you create the instance you change the level already. so it does not mean the same AFAIK

MarianMacik:
jenkins retest this

MarianMacik:
@gmunozfe can you have a look at this as well?

gmunozfe:
My feedback is a little bit above: it looks good to me.
Marian's proposal also works (I've tested it), but I think readability is worse

MarianMacik:
@gmunozfe Thanks, merging.

PR #1530, opened 2019-08-26
Stabilization and fixes on some wrong traces

gmunozfe:
In windows 2K16 instances, two tests were failing due to execution is so fast that milliseconds precision is not enough: 

- added  "<=" in the failing Assert for _testCustomIncrementingRequestRetrySpecialValues_
- changed PrintOutCommand to DelayedPrintOutCommand for _testPrioritizedJobsExecutionInvalidProrities_

Fixing also some warnings at priority bounds check, as the order of binding parameters was wrong.

@MarianMacik please review

### DIFF
--- a/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/BasicExecutorBaseTest.java
+++ b/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/BasicExecutorBaseTest.java
@@ -461,7 +461,7 @@ public abstract class BasicExecutorBaseTest {
         // Future execution is planned to be started 2 minutes and 20 seconds after last fail.
         // Time difference vary because of test thread sleeping for 10 seconds.
         diff = allRequests.get(0).getTime().getTime() - Calendar.getInstance().getTimeInMillis();
-        assertTrue(diff < 140000);
+        assertTrue(diff <= 140000);
         assertTrue(diff > 130000);
 
         executorService.clearAllRequests();
@@ -659,7 +659,7 @@ public abstract class BasicExecutorBaseTest {
         assertNotNull(executedLow);
         assertEquals("low priority", executedLow.getKey());
         
-        assertTrue(executedLow.getTime().getTime() > executedHigh.getTime().getTime());
+        assertTrue(executedLow.getTime().getTime() >= executedHigh.getTime().getTime());
     }
     
     @Test(timeout=10000)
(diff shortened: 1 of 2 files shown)


jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/BasicExecutorBaseTest.java
```diff
@@ -631,16 +631,18 @@ public void testPrioritizedJobsExecutionInvalidProrities() throws InterruptedExc
         CommandContext ctxCMD = new CommandContext();
         ctxCMD.setData("businessKey", "low priority");
         ctxCMD.setData("priority", -1);
+        ctxCMD.setData("delay", 10L);
         
         Date futureDate = new Date(System.currentTimeMillis() + EXTRA_TIME);
 
-        executorService.scheduleRequest("org.jbpm.executor.commands.PrintOutCommand", futureDate, ctxCMD);
+        executorService.scheduleRequest("org.jbpm.executor.commands.DelayedPrintOutCommand", futureDate, ctxCMD);
```
MarianMacik: This will just prolong the execution of the command by including the `Thread.sleep()` in it. I think we discussed change on the line 664 that we need `>=` there, didn't we?

gmunozfe: Yes, we discussed "greater and equals" as a possible solution; my point is that if the command lasts several milliseconds (with the delay we can control that), then, we may have a strict greater. Also I've noticed that there were no tests for DelayPrintOutCommand. But, if you prefer, we can just keep the change as commented previously.

MarianMacik: Well, if i understand it correctly, the duration of the command doesn't influence the `time` field of the `RequestInfo` object. This field is updated only when the request is obtained from DB to be run. See here:
https://github.com/kiegroup/jbpm/blob/05a632fc1c11608a89dbce5f63532af04b38fb97/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/jpa/ExecutorQueryServiceImpl.java#L283-L291

It is not updated afterwards, at least I haven't found any occurrence of it. But I may be missing something, of course.

gmunozfe: You're right; equals comparison is therefore needed anyway. Keep it simple with just that change. Thanks!

MarianMacik:
Jenkins retest this

PR #1535, opened 2019-09-18
[JBPM-8781] API to view all of sub-cases belong to a parent case

elguardian:
Adding a query a collection of sub cases from parentCaseId

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseRuntimeDataService.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseRuntimeDataService.java
@@ -184,7 +184,22 @@ public interface CaseRuntimeDataService {
      *
      */
     Collection<CaseInstance> getCaseInstances(QueryContext queryContext);
-   
+
+    /**
+     * Returns all first level children cases given an parent id case.
+     * @param parentCaseId parent case
+     * @param statuses list of statuses that subcases should be in to match
+     * @param queryContext control parameters for the result e.g. sorting, paging
+     */
+    Collection<CaseInstance> getSubCaseInstancesByParentCaseId(String parentCaseId, List<CaseStatus> statuses, QueryContext queryContext);
+
+    /**
+     * Returns all subtree cases given an parent id case.
+     * @param parentCaseId parent case
+     * @param statuses list of statuses that subcases should be in to match
+     */
+    Collection<CaseInstance> getAllDescendantSubCaseInstancesByParentCaseId(String parentCaseId, List<CaseStatus> statuses);
+
     /**
      * Returns all available active case instances that match given statuses
      * @param statuses list of statuses that case should be in to match
--- a/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/model/instance/CaseInstance.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/model/instance/CaseInstance.java
@@ -105,4 +105,10 @@ public interface CaseInstance {
      * Returns up to date SLA compliance level for case instance
      */
     Integer getSlaCompliance();
+
+    /**
+     * Returns the parent id case if any
+     * @return the parent case id
+     */
+    String getParentCaseId();
 }
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
@@ -16,11 +16,6 @@
 
 package org.jbpm.casemgmt.impl;
 
-import static java.util.stream.Collectors.toList;
-import static java.util.stream.Collectors.toMap;
-import static org.jbpm.kie.services.impl.CommonUtils.getAuthenticatedUserRoles;
-import static org.kie.internal.query.QueryParameterIdentifiers.FILTER;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -88,6 +83,11 @@ import org.kie.internal.KieInternalServices;
 import org.kie.internal.identity.IdentityProvider;
 import org.kie.internal.process.CorrelationKey;
 import org.kie.internal.process.CorrelationKeyFactory;
+
+import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
+import static org.jbpm.kie.services.impl.CommonUtils.getAuthenticatedUserRoles;
+import static org.kie.internal.query.QueryParameterIdentifiers.FILTER;
 
 
 public class CaseRuntimeDataServiceImpl implements CaseRuntimeDataService, DeploymentEventListener {
@@ -437,6 +437,33 @@ public class CaseRuntimeDataServiceImpl implements CaseRuntimeDataService, Deplo
         return processInstances;
     }
 
+    @Override
+    public Collection<CaseInstance> getSubCaseInstancesByParentCaseId(String parentCaseId, List<CaseStatus> statuses, QueryContext queryContext) {
+        return getSubCases(Arrays.asList(parentCaseId), statuses, queryContext);
+    }
+
+    @Override
+    public Collection<CaseInstance> getAllDescendantSubCaseInstancesByParentCaseId(String parentCaseId, List<CaseStatus> statuses) {
+        // this will iterate foreach level. It will do as much queries as levels in the hierarchy
+        List<CaseInstance> cases = new ArrayList<>();
+        List<String> ids = Arrays.asList(parentCaseId);
+        do {
+            Collection<CaseInstance> processInstances = getSubCases(ids, statuses, null);
+            cases.addAll(processInstances);
+            ids = processInstances.stream().map(CaseInstance::getCaseId).collect(Collectors.toList());
+        } while (!ids.isEmpty()); // empty means there are not more leves to go
+
+        return cases;
+    }
+
+    private Collection<CaseInstance> getSubCases(List<String> parentCaseId, List<CaseStatus> statuses, QueryContext queryContext) {
+        Map<String, Object> params = new HashMap<String, Object>();
+        params.put("parentCaseId", parentCaseId);
+        params.put("statuses", resolveCaseStatuses(statuses));
+        applyQueryContext(params, queryContext);
+        applyDeploymentFilter(params);
+        return commandService.execute(new QueryNameCommand<List<CaseInstance>>("getSubCaseInstancesByParentCaseIds", params));
+    }
 
     @Override
     public Collection<CaseInstance> getCaseInstancesByDeployment(String deploymentId, List<CaseStatus> statuses, QueryContext queryContext) {
@@ -885,5 +912,4 @@ public class CaseRuntimeDataServiceImpl implements CaseRuntimeDataService, Deplo
 
         return caseFileItems;
     }
-
 }
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
@@ -25,6 +25,7 @@ import java.util.Map.Entry;
 import org.drools.core.definitions.rule.impl.RuleImpl;
 import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
 import org.jbpm.casemgmt.impl.event.CaseEventSupport;
+import org.jbpm.casemgmt.impl.model.instance.CaseFileInstanceImpl;
 import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.services.api.ProcessService;
 import org.kie.api.KieServices;
@@ -118,6 +119,10 @@ public class StartCaseCommand extends CaseCommand<Void> {
         logger.debug("Starting process instance for case {} and case definition {}", caseId, caseDefinitionId);
         CorrelationKey correlationKey = correlationKeyFactory.newCorrelationKey(caseId);
         Map<String, Object> params = new HashMap<>();
+        // add process parent
+        if (caseFile instanceof CaseFileInstanceImpl && ((CaseFileInstanceImpl) caseFile).getParentInstanceId() != null && ((CaseFileInstanceImpl) caseFile).getParentInstanceId() >= 0) {
+            params.put("ParentInstanceId", ((CaseFileInstanceImpl) caseFile).getParentInstanceId());
+        }
         // set case id to allow it to use CaseContext when creating runtime engine
         params.put(EnvironmentName.CASE_ID, caseId);
         final Map<String, Object> caseData = caseFile.getData();
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
@@ -21,10 +21,12 @@
         log.processInstanceId, <!-- case top level process instance id -->
         log.outcome,
         log.slaDueDate,
-        log.slaCompliance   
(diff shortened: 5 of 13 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/CaseRuntimeDataService.java
```diff
@@ -184,7 +184,14 @@
      *
      */
     Collection<CaseInstance> getCaseInstances(QueryContext queryContext);
-   
+
+    /**
+     * Returns all first level children cases given an parent id case.
+     * @param parentCaseId parent case
+     * @param statuses 
+     * @param queryContext
```
gmunozfe: missing javadoc coments, as in other methods (clarify that statuses are for subcases):
 _@param statuses list of statuses that subcases should be in to match
   @param queryContext control parameters for the result e.g. sorting, paging_

jbpm-case-mgmt/jbpm-case-mgmt-api/src/main/java/org/jbpm/casemgmt/api/model/instance/CaseInstance.java
```diff
@@ -105,4 +105,10 @@
      * Returns up to date SLA compliance level for case instance
      */
     Integer getSlaCompliance();
+
+    /**
+     * Returns the parent id case if any
+     * @return
```
gmunozfe: Same as previous comment

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
```diff
@@ -34,6 +36,38 @@
...
+        log.slaCompliance,
+        parent.correlationKey <!-- case parent id -->
+      )
+      from
+        ProcessInstanceLog log, ProcessInstanceLog parent
+      where 
+        log.processType = 2
+        and log.status in (:statuses)
+        and log.parentProcessInstanceId = parent.processInstanceId
+        and parent.correlationKey like :parentCaseId
```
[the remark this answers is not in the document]
elguardian: have a look at getActiveNodesForCase.... just following that convention.

elguardian:
Jenkins retest this please.
@gmunozfe I think i didn't miss any comment

gmunozfe:
@elguardian thanks! you covered all.
As SonarCloud is complaining about code coverage, you can also add this line to the test:

_assertEquals(SUB_CASE_ID, subcases.iterator().next().getParentCaseId());_

elguardian:
this is ok to test.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/SubCaseServiceImplTest.java
```diff
@@ -485,4 +486,52 @@ public void testStartCaseWithSubCaseAbortProcessInstanceOfSubCase() {
...
+
+            // we check the parent
+            List<CaseStatus> allStatus = Arrays.asList(CaseStatus.values());
+            Collection<CaseInstance> subcases = caseRuntimeDataService.getSubCaseInstancesByParentCaseId(SUB_CASE_ID, allStatus, new QueryContext());
+            assertNotNull(subcases);
+            assertEquals(1, subcases.size());
+            List<String> childId = subcases.stream().map(CaseInstance::getCaseId).collect(Collectors.toList());
+            assertEquals(1, childId.size()); // only one child
+            assertTrue(childId.contains(HR_CASE_ID));
+            assertEquals(SUB_CASE_ID, subcases.iterator().next().getParentCaseId());
```
gmunozfe: SonarCloud still complaining about coverage under 80% :-( 
Include toString in some test case.

```suggestion
            CaseInstance sci = subcases.iterator().next();
            assertEquals(SUB_CASE_ID, sci.getParentCaseId());
            assertTrue(sci.toString().contains("parentCaseId="+SUB_CASE_ID));
```

elguardian:
@MarianMacik @gmunozfe is there any progress with this ?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
```diff
@@ -437,6 +437,36 @@ public CaseDefinition getCase(String deploymentId, String caseDefinitionId) {
...
+        params.put("statuses", resolveCaseStatuses(statuses));
+        applyQueryContext(params, queryContext);
+        applyDeploymentFilter(params);
+        List<CaseInstance> processInstances = commandService.execute(new QueryNameCommand<List<CaseInstance>>("getSubCaseInstancesByParentCaseIds", params));
+
+        return processInstances;
+    }
+
+    @Override
+    public Collection<CaseInstance> getAllDescendantSubCaseInstancesByParentCaseId(String parentCaseId, List<CaseStatus> statuses, QueryContext queryContext) {
```
gmunozfe: Here the QueryContext is applied per level:
Suppose limit is X, then if each level returns X-1 results, the API will return N levels * (X-1) results, exceeding the global limit (X). 
Same issue happens with the offset: if first level subcases number is lower than the offset, then it won't iterate over the next level (though there're remaining results).
Perhaps for this API, it's better an implementation without queryContext, querying as many times as needed per level and consolidate them all to be returned in a single call.
What do you think?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/SubCaseServiceImplTest.java
```diff
@@ -485,4 +486,60 @@ public void testStartCaseWithSubCaseAbortProcessInstanceOfSubCase() {
...
+
+            Collection<CaseInstance> allSubcases = caseRuntimeDataService.getAllDescendantSubCaseInstancesByParentCaseId(SUB_CASE_ID, allStatus, new QueryContext());
```
gmunozfe: This test covers just one level (same result as the other API).
To add more levels, it's needed to add the custom task (subcase) inside UserTaskCase.bpmn, and then invoke it as many times as nested cases we want to have:
`caseService.triggerAdHocFragment("HR-000000000..", "New Nested Sub Case", null);`

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
elguardian: done

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
```diff
@@ -437,6 +437,36 @@ public CaseDefinition getCase(String deploymentId, String caseDefinitionId) {
...
+    @Override
+    public Collection<CaseInstance> getSubCaseInstancesByParentCaseId(String parentCaseId, List<CaseStatus> statuses, QueryContext queryContext) {
+        Map<String, Object> params = new HashMap<String, Object>();
+        params.put("parentCaseId", Arrays.asList(parentCaseId));
+        params.put("statuses", resolveCaseStatuses(statuses));
+        applyQueryContext(params, queryContext);
+        applyDeploymentFilter(params);
+        List<CaseInstance> processInstances = commandService.execute(new QueryNameCommand<List<CaseInstance>>("getSubCaseInstancesByParentCaseIds", params));
+
+        return processInstances;
```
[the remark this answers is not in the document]
elguardian: done

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/SubCaseServiceImplTest.java
elguardian: done...

elguardian:
@gmunozfe check the last changes...
@mswiderski approve if you think it is ok

elguardian:
this is ok to test.

gmunozfe:
@elguardian now it looks fine to me

elguardian:
@MarianMacik don't know who is responsbile for merging now when @mswiderski is not available :?

MarianMacik:
@elguardian I can merge it as well. @gmunozfe are we ok with this, can you approve it?

gmunozfe:
@MarianMacik I'm ok with this but have no permissions to approve it

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
```diff
@@ -21,10 +21,12 @@
         log.processInstanceId, <!-- case top level process instance id -->
         log.outcome,
         log.slaDueDate,
-        log.slaCompliance   
+        log.slaCompliance,
+        parent.correlationKey
       )
       from
         ProcessInstanceLog log
+      left join ProcessInstanceLog parent on log.parentProcessInstanceId = parent.processInstanceId
```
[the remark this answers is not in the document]
elguardian: @MarianMacik  id and fk should be indexed, so it should be fine.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/cases/UserTaskCase.bpmn2
```diff
@@ -5,7 +5,7 @@
   <bpmn2:process id="UserTaskCase" tns:adHoc="true" name="Simple Case with User Tasks" isExecutable="true" processType="Private" tns:version="1.0" >
     <bpmn2:extensionElements>
       <tns:metaData name="customDescription">
-        <tns:metaValue>Case #{s}</tns:metaValue>
+        <tns:metaValue>Case </tns:metaValue>
```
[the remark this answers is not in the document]
elguardian: having a stacktrace because of the missing parameter.

elguardian:
I did not go that further with the analisys there I presumed there was a good reason to discard parent-child link. To be honest it seems a bit off because we are losing info. But breaking that logic now would be problematic. We can discuss in a new jira. For now the current scope should be sufficient for the features should be enough.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/cases/UserTaskCase.bpmn2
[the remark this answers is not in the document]
MarianMacik: Ah, you mean error log? I was wondering why the tests are passing, but the process instance/case instance variable is still available and defined as s and in fact no test checks the custom description so it works :)

MarianMacik:
@elguardian I think the reason was as I said in the last paragraph. Makes sense to leave it as it is now.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
```diff
@@ -118,6 +119,10 @@ public void matchCreated(MatchCreatedEvent event) {
         logger.debug("Starting process instance for case {} and case definition {}", caseId, caseDefinitionId);
         CorrelationKey correlationKey = correlationKeyFactory.newCorrelationKey(caseId);
         Map<String, Object> params = new HashMap<>();
+        // add process parent
+        if (caseFile instanceof CaseFileInstanceImpl && ((CaseFileInstanceImpl) caseFile).getParentInstanceId() != null && ((CaseFileInstanceImpl) caseFile).getParentInstanceId() >= 0) {
+            params.put("ParentProcessCaseId", ((CaseFileInstanceImpl) caseFile).getParentInstanceId());
```
mswiderski: this key is very weird named - so what it is actually? is it parent process id or parent case id or... you got me confused here

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
mswiderski: has someone tested this on large ProcessInstanceLog table? asking if it has significant impact on performance or maybe there will be a need for some indexes...

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
```diff
@@ -598,10 +646,12 @@
         log.processInstanceId, <!-- case top level process instance id -->
         log.outcome,
         log.slaDueDate,
-        log.slaCompliance
+        log.slaCompliance,
+        parent.correlationKey
       )
       from
         ProcessInstanceLog log
+      left join ProcessInstanceLog parent on log.parentProcessInstanceId = parent.processInstanceId
```
mswiderski: I am bit concerned with adding this extra join to all queries, again... we need to assess the impact on performance

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
elguardian: @mswiderski ParentInstanceId

mswiderski: so would make sense to update it so it is clear what it's meant to be

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
elguardian: EXPLAIN SELECT * FROM ProcessInstanceLog a LEFT JOIN ProcessInstanceLog b on a.id = b.parentProcessInstanceId 
throws this this data IDX_PInstLog_parentPInstId will be used

The only thing that come to my mind is to duplicate the call or something.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
elguardian: above

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
mswiderski: so with this we assume it will be performant enough?

@MarianMacik do we have already some test cases with loaded db tables?

elguardian: @mswiderski @MarianMacik 
just tried a small snippet here to load a mysql database 
https://github.com/kiegroup/jbpm/pull/1562

2800 (1400 + 1 child per case) cases about to 3 seconds with or without the join. I let the code in there (SubCaseServiceImplLoadTest)

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
elguardian: done

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/resources/META-INF/CaseMgmtorm.xml
mswiderski: ok, so we don't have much difference then we should be fine, thanks @elguardian

elguardian:
can anyone merge @kiegroup/gatekeepers ?

krisv:
@elguardian can you open the PR for 7.30.x as well, and push this forward so this gets merged before EOD Thu?

elguardian:
done @krisv

PR #1536, opened 2019-09-20
Stabilize tests due to fast exec at win2k16 servers

gmunozfe:
Some tests need a little tuning due to fast execution at windows 2016 servers:

-deleteErrorLogsByDate, it's removing two logs (instead of expecting one) because dates are the same for both. Added this scenario to the test.

-ServiceTaskHandlerTests (testShorten[Full]InterfaceNameBPMN2[Workitems]) are expecting 1 as result (input date lower) but obtaining 0 (same date). Updated to pass a previous date (day before) instead of current date.

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ExecutorLogCleanTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ExecutorLogCleanTest.java
@@ -120,11 +120,22 @@ public class ExecutorLogCleanTest extends JbpmAsyncJobTestCase {
                 .date(errorList.get(0).getTime())
                 .build()
                 .execute();
-        Assertions.assertThat(resultCount).isEqualTo(1);
+        
+        //normal scenario, i.e, errorList time is different for both
+        int deleted = 1;
+        int remaining = 1;
+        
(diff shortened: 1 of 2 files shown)


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ExecutorLogCleanTest.java
```diff
@@ -120,11 +120,22 @@ public void deleteErrorLogsByDate() throws Exception {
...
                 .build()
                 .execute();
-        Assertions.assertThat(resultCount).isEqualTo(1);
+        
+        //normal scenario, i.e, errorList time is different for both
+        int deleted = 1;
+        int remaining = 1;
+        
+        //if both have the same time
+        if (errorList.get(0).getTime().compareTo(errorList.get(1).getTime())==0) {
```
MarianMacik: Why not to use equals here?

gmunozfe: Updated to equals (easier to read)

MarianMacik:
jenkins retest this

PR #1545, opened 2019-10-03
[JBPM-8698] Cannot trigger activities inside asynchronous ad-hoc subprocess

elguardian:
adding support for async ad-hoc subprocess

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
@@ -256,13 +256,17 @@ public abstract class WorkflowProcessInstanceImpl extends ProcessInstanceImpl
 		return result;
 	}
 
-	@Override
-	public NodeInstance getNodeInstance(final Node node) {
+    @Override
+    public NodeInstance getNodeInstance(final Node node) {
+        return getNodeInstance(node, true);
+    }
+
+    public NodeInstance getNodeInstance(final Node node, boolean wrap) {
 	    Node actualNode = node;
 	    // async continuation handling
-	    if (node instanceof AsyncEventNode) {
+        if (node instanceof AsyncEventNode) {
             actualNode = ((AsyncEventNode) node).getActualNode();
-        } else if (useAsync(node)) {
+        } else if (wrap && useAsync(node)) {
             actualNode = new AsyncEventNode(node);
         }
 
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
@@ -43,8 +43,8 @@ public class TriggerNodeCommand implements ExecutableCommand<Void>, ProcessInsta
     }
 
     public Void execute(Context context) {
-    	KieSession kieSession = ((RegistryContext) context).lookup( KieSession.class );
-    	logger.debug("About to trigger (create) node instance for node {} in process instance {}", nodeId, processInstanceId);
+        KieSession kieSession = ((RegistryContext) context).lookup(KieSession.class);
+        logger.debug("About to trigger (create) node instance for node {} in process instance {}", nodeId, processInstanceId);
         RuleFlowProcessInstance wfp = (RuleFlowProcessInstance) kieSession.getProcessInstance(processInstanceId, false);
         if (wfp == null) {
             throw new ProcessInstanceNotFoundException("Process instance with id " + processInstanceId + " not found");
@@ -62,10 +62,13 @@ public class TriggerNodeCommand implements ExecutableCommand<Void>, ProcessInsta
             if (node == null) {
                 throw new NodeNotFoundException("Node with id " + nodeId + " not found");
             }
-            NodeInstanceContainer nodeInstanceContainerNode =
-                    ((NodeInstanceContainer) wfp.getNodeInstance(wfp.getRuleFlowProcess().getParentNode(nodeId)));
+
+            Node parentNode = wfp.getRuleFlowProcess().getParentNode(nodeId);
+            // we need to unwrap the instance otherwise we are not getting the right nodeInstance
+            NodeInstanceContainer parentContainer = (NodeInstanceContainer) wfp.getNodeInstance(parentNode, false);
+
             logger.debug("Triggering node {} on process instance {}", node, wfp);
-            nodeInstanceContainerNode.getNodeInstance(node).trigger(null, org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
+            parentContainer.getNodeInstance(node).trigger(null, org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
             logger.debug("Node {} successfully triggered", node);
         } else {
             logger.debug("Triggering node {} on process instance {}", node, wfp);
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/async/AsyncAdHocSubprocessTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/async/AsyncAdHocSubprocessTest.java
@@ -1,0 +1,96 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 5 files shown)


elguardian:
this is ok to test.

elguardian:
this is ok to test.

elguardian:
this is ok to test.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
```diff
@@ -62,8 +65,16 @@ public Void execute(Context context) {
...
-            NodeInstanceContainer nodeInstanceContainerNode =
-                    ((NodeInstanceContainer) wfp.getNodeInstance(wfp.getRuleFlowProcess().getParentNode(nodeId)));
+
+            Node parentNode = wfp.getRuleFlowProcess().getParentNode(nodeId);
+
+            // we need to unwrap the instance otherwise we are not getting the right nodeInstance
+            Map<String, Object> parentNodeMetaData = parentNode.getMetaData();
+            Boolean oldWrapValue = (Boolean) parentNodeMetaData.get(NodeInstance.METADATA_WRAP_ASYNC_NODE);
+            parentNodeMetaData.put(NodeInstance.METADATA_WRAP_ASYNC_NODE, false);
+            NodeInstanceContainer nodeInstanceContainerNode = (NodeInstanceContainer) wfp.getNodeInstance(parentNode.getId());
```
gmunozfe: There's a _NullPointerException_ in test _**ProcessInstanceAdminServiceImplTest#testTriggerSubProcessNodeInstance**_ due to this line returns null.

In that test, _parentNode.getId()_ corresponds to _node 2_:

_id: 2,  name= Hello_, being its container a _RuleFlowProcess_ (Ad-Hoc Sub-Process).

However, when retrieving the NodeInstanceContainer, as it's done in this line, with 

> wfp.getNodeInstance(parentNode.getId())

it's not found, because the node _"Hello"_ has:
_id=1, nodeId=2_

Therefore, it should match _nodeId_, instead of _id_.

Other possibility I've tested and worked out is using a filter like following suggestion replacing _getNodeInstance_ invocation:

```suggestion
           NodeInstanceContainer nodeInstanceContainerNode = (NodeInstanceContainer) wfp.getNodeInstances().stream().filter(ni -> ni.getNodeId() == parentNode.getId()).findFirst().orElse(null);
```

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
```diff
@@ -62,8 +65,16 @@ public Void execute(Context context) {
...
+            parentNodeMetaData.put(NodeInstance.METADATA_WRAP_ASYNC_NODE, oldWrapValue);
+
             logger.debug("Triggering node {} on process instance {}", node, wfp);
             nodeInstanceContainerNode.getNodeInstance(node).trigger(null, org.jbpm.workflow.core.Node.CONNECTION_DEFAULT_TYPE);
```
gmunozfe: It should be checked that nodeInstanceContainerNode is not null.

elguardian:
@gmunozfe I am just testing here solutions. I would take forever to test this in local.
this is ok to test.

gmunozfe:
sure @elguardian, no problem. 

> wfp.getNodeInstance(parentNode);

 fixed also that test :)

elguardian:
@gmunozfe seems uberfire is broken.

elguardian:
this is ok to test.

elguardian:
Jenkins restest this please.

elguardian:
@gmunozfe now you can review this.

elguardian:
this is ok to test

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
```diff
@@ -74,11 +73,7 @@ public void internalTrigger(NodeInstance from, String type) {
         	agenda.getRuleFlowGroup(ruleFlowGroup).setAutoDeactivate(false);
         	agenda.activateRuleFlowGroup(ruleFlowGroup, getProcessInstance().getId(), getUniqueId());
     	}
-//    	if (getDynamicNode().isAutoComplete() && getNodeInstances(false).isEmpty()) {
-//    		triggerCompleted(NodeImpl.CONNECTION_DEFAULT_TYPE);
-//    	}
```
mswiderski: why this is commented out?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/AsyncEventNodeInstance.java
```diff
@@ -105,12 +106,13 @@ public void triggerCompleted() {
         getProcessInstance().removeEventListener(getEventType(), getEventListener(), true);
         ((org.jbpm.workflow.instance.NodeInstanceContainer)getNodeInstanceContainer()).setCurrentLevel(getLevel());
         ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).removeNodeInstance(this);
-        
-        NodeInstance instance = ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer()).getNodeInstance(getNode());
+        NodeInstanceContainer container = ((org.jbpm.workflow.instance.NodeInstanceContainer) getNodeInstanceContainer());
+        NodeInstance instance = container.getNodeInstance(getNode());
```
mswiderski: what's the reason for this extra variable `container`?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/TriggerNodeCommand.java
```diff
@@ -62,10 +65,18 @@ public Void execute(Context context) {
...
                 throw new NodeNotFoundException("Node with id " + nodeId + " not found");
             }
-            NodeInstanceContainer nodeInstanceContainerNode =
-                    ((NodeInstanceContainer) wfp.getNodeInstance(wfp.getRuleFlowProcess().getParentNode(nodeId)));
+
+            Node parentNode = wfp.getRuleFlowProcess().getParentNode(nodeId);
+            // we need to unwrap the instance otherwise we are not getting the right nodeInstance
+            Map<String, Object> parentNodeMetaData = parentNode.getMetaData();
+            Boolean oldWrapValue = (Boolean) parentNodeMetaData.get(NodeInstance.METADATA_WRAP_ASYNC_NODE);
+            parentNodeMetaData.put(NodeInstance.METADATA_WRAP_ASYNC_NODE, false);
```
mswiderski: this is not thread safe operation as the node is shared across threads running in parallel. So changing it here and another thread for another process instance can change it back before the node instance is retrieved. I'd create new method in workflowprocessinstance that can to the explicit unwrap.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
elguardian: no idea.... Just removed it when I saw commented code.

elguardian:
done @mswiderski

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicNodeInstance.java
mswiderski: interesting I noticed it was commented out but not that it was (rightfully removed), thanks!

elguardian:
Jenkins restest this please.

elguardian:
Jenkins restest this please.

elguardian:
Jenkins restest this please.

elguardian:
@kiegroup/gatekeepers can anyone merge ?

mareknovotny:
Can you approve this PR ? @gmunozfe

gmunozfe:
This is ok to merge

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -256,13 +256,17 @@ public NodeInstance getFirstNodeInstance(final long nodeId) {
...
+    }
+
+    public NodeInstance getNodeInstance(final Node node, boolean unwrap) {
 	    Node actualNode = node;
 	    // async continuation handling
-	    if (node instanceof AsyncEventNode) {
+        if (node instanceof AsyncEventNode) {
             actualNode = ((AsyncEventNode) node).getActualNode();
-        } else if (useAsync(node)) {
+        } else if (unwrap && useAsync(node)) {
```
MarianMacik: @elguardian just a question - this should be called wrap then, shouldn't it? We are passing there false when we want it unwrapped.

elguardian: done

mareknovotny:
Jenkins retest this please.

elguardian:
@mareknovotny can you merge ?

PR #1547, opened 2019-10-14
RHPAM-2449 - Perform security (based on required roles) on process op…

mswiderski:
…erations only

depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/405

@MarianMacik @gmunozfe @elguardian could you please guys have a look? I tried to make a bit of progress on it but would appreciate some work on verifying it.

### DIFF
--- a/jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/ExternalTaskEventListener.java
+++ b/jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/ExternalTaskEventListener.java
@@ -26,6 +26,7 @@ import org.kie.api.task.TaskEvent;
 import org.kie.api.task.TaskLifeCycleEventListener;
 import org.kie.api.task.model.Status;
 import org.kie.api.task.model.Task;
+import org.kie.internal.runtime.manager.InternalRuntimeEngine;
 import org.kie.internal.runtime.manager.RuntimeManagerRegistry;
 import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
 import org.kie.internal.task.api.prediction.PredictionService;
@@ -49,7 +50,7 @@ public class ExternalTaskEventListener implements TaskLifeCycleEventListener {
         long processInstanceId = task.getTaskData().getProcessInstanceId();
         RuntimeManager manager = getManager(task);
         RuntimeEngine runtime = manager.getRuntimeEngine(ProcessInstanceIdContext.get(processInstanceId));
-        KieSession session = runtime.getKieSession();
+        KieSession session = ((InternalRuntimeEngine) runtime).internalGetKieSession();
 
         if (task.getTaskData().getStatus() == Status.Completed) {
             String userId = task.getTaskData().getActualOwner().getId();
@@ -105,7 +106,7 @@ public class ExternalTaskEventListener implements TaskLifeCycleEventListener {
             throw new RuntimeException("No RuntimeManager registered with identifier: " + task.getTaskData().getDeploymentId());
         }
         RuntimeEngine runtime = manager.getRuntimeEngine(ProcessInstanceIdContext.get(processInstanceId));
-        KieSession session = runtime.getKieSession();
+        KieSession session = ((InternalRuntimeEngine) runtime).internalGetKieSession();
         if (session != null) {
             logger.debug(">> I've recieved an event for a known session (" + task.getTaskData().getProcessSessionId()+")");
             processTaskState(task);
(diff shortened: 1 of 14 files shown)


gmunozfe:
hi @mswiderski,
There's a NPE in many tests, due to
https://github.com/mswiderski/jbpm/blob/0ebe8151743406630be113a8a4747e7aa55b386d/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java#L201
can return null, when ksession is not initialized.

Check my proposal to sort it out:
https://github.com/mswiderski/jbpm/compare/RHPAM-2449...gmunozfe:RHPAM-2449

Both methods will call initializer, but to avoid infinite-recursively calls to internalGetKieSession() from RuntimeManagers implementation, we may use getKieSessionId() to check that sessionId is not null.

With this approach, only 2 tests keep on failing, and need to be studied:
ERROR] Failures: 
[ERROR]   SecurityManagerTest.testCustomSecurityManager Expected exception: java.lang.SecurityException
[ERROR]   SecurityManagerTest.testDenyAllSecurityManager Expected exception: java.lang.SecurityException

If you agree, I can create a PR to your repo for you to merge.

mswiderski:
@gmunozfe looks like a valid approach, so go ahead with PR. Thanks

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskWithSecurityTest.java
```diff
@@ -0,0 +1,203 @@
...
+        
+        roles.clear();
+        identityProvider.setRoles(roles);
+        List<TaskSummary> tinstances = runtimeDataService.getTasksAssignedAsPotentialOwner(identityProvider.getName(), new QueryFilter());
+        assertNotNull(tinstances);
+        assertEquals(1, tinstances.size());
+        
+        Long taskID = tinstances.get(0).getId();
+        
+        userTaskService.completeAutoProgress(taskID, identityProvider.getName(), null);
```
gmunozfe: Shouldn't roles be added before this call? If not, a SecurityException is thrown. 
Just moving this line, 4 lines below makes it pass.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskWithSecurityTest.java
```diff
@@ -0,0 +1,203 @@
...
+    @Parameters(name="{0}")
+    public static Collection<Object[]> data() {
+        return Arrays.asList(new Object[][] {     
+                 { RuntimeStrategy.SINGLETON }, { RuntimeStrategy.PER_PROCESS_INSTANCE }, { RuntimeStrategy.PER_REQUEST }, { RuntimeStrategy.PER_CASE } 
+           });
+    }
+    
+    private List<DeploymentUnit> units = new ArrayList<DeploymentUnit>();
+    protected String correctUser = "testUser";
+    protected String wrongUser = "wrongUser";
```
gmunozfe: These users are not used in the test

jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/ExternalTaskEventListener.java
```diff
@@ -105,7 +106,7 @@ public void afterTaskCompletedEvent(TaskEvent event) {
             throw new RuntimeException("No RuntimeManager registered with identifier: " + task.getTaskData().getDeploymentId());
         }
         RuntimeEngine runtime = manager.getRuntimeEngine(ProcessInstanceIdContext.get(processInstanceId));
-        KieSession session = runtime.getKieSession();
+        KieSession session = ((InternalRuntimeEngine) runtime).internalGetKieSession();
```
MarianMacik: Just if I get it right - this swap of methods allows us to omit the call to `((AbstractRuntimeManager) manager).checkPermission();`?

gmunozfe:
@MarianMacik the point is that previous calls to _RuntimeEngineImpl.getKieSession()_ have switched to _internalGetKieSession()_ that just returned the ksession (even if null, which provoked a NPE).

My proposal was that both methods have the same implementation (calling _initializer.initKieSession_ when there's no session, so getKieSession internally will invoke _internalGetKieSession_) but _getKieSession_ also adds the security checks.

To avoid infinite recursive calls (_initKieSession_ checked that session was not null with _internalGetKieSession_), we can check it with the _getKieSessionId()_ that returns just the id (a long, that must be initialized to null).

MarianMacik:
@gmunozfe thanks, got it.

gmunozfe:
jenkins retest this

gmunozfe:
jenkins retest this

gmunozfe:
jenkins retest this

PR #1549, opened 2019-10-21
Don't use forkMode because it is deprecated

MarianMacik:
1. `forkMode` is deprecated
2. ~~We don't need it in 2 of 3 modules as it doesn't contain any Arquillian tests.~~

### DIFF
--- a/jbpm-runtime-manager/pom.xml
+++ b/jbpm-runtime-manager/pom.xml
@@ -255,7 +255,6 @@
       and to avoid clashing between them as both need to setup data source and bound it to JNDI -->
         <artifactId>maven-surefire-plugin</artifactId>
         <configuration>
-          <forkMode>always</forkMode>
           <!-- Properties for Byteman which allows it to run on IBM Java as well -->
           <argLine>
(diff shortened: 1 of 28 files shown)


MarianMacik:
@gmunozfe OK, I have debugged some tests and actually found some room for improvement. For example here https://github.com/kiegroup/jbpm/pull/1549/files#diff-974af927fd1e2163040f1e67ec3653adR167 the unit wasn't added which broke tests. It doesn't matter when we create a new JVM for each test, but I think it is better now that the test is properly cleaned up.

Another thing was that the deployment listener was registered, but not configured at all, see https://github.com/kiegroup/jbpm/pull/1549/files#diff-7181ebb799758e9645876cbcab511e48R134-R139, so tests in `DeploymentServiceWithSyncTest` were relying on timeouts as the countdown latch always timed out since the listener never triggered it. Now the tests execute twice as fast since the listener works correctly.

And I also renamed `CoundDownDeploymentListener` to `CountDownDeploymentListener`. This listener is only used for tests, so it is not a public API change.

PR #1550, opened 2019-10-24
JBPM-8837 Remove Service Task from the repository

bxf12315:
@mswiderski @tsurdilo 
I add gav in here, is it reasonable?


mswiderski:
frankly speaking I don't think we need it as described in the other comments in Jbpm-work-items repo

PR #1552, opened 2019-10-28
[JBPM-8866] GlobalTimerService.timerJobsPerSession leak with StartProcess

elguardian:


elguardian:
this is ok to test

elguardian:
@mswiderski Not really sure if this is the right fix. 
the StartProcessJobContext is registered with and old session id during session creation.
During the process start (timeout scheduler) it creates a new KieSession (GlobalTimerService.getRunner - different from the one the job contexts was created. When the timeout happens at the end it disposes the jobs per session, given that the session id is not the one which the process is being produced it fails to clean up.
it also happens that the StartProcessJobContext loses its sessionId during scheduling.
I think removing this sort of job context makes sense because it is not a job belonging to the process (it happens before it starts).

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
```diff
@@ -80,7 +81,9 @@ private TimerJobFactoryManager initJobFactoryManager() {
 
     @Override
     public JobHandle scheduleJob(Job job, JobContext ctx, Trigger trigger) {
-        if (ctx instanceof ProcessJobContext) {
+        // start process job context it is not job per session is triggers even before 
+        // the process
+        if (ctx instanceof ProcessJobContext && !(ctx instanceof StartProcessJobContext)) {
```
MarianMacik: This kind of makes sense to me that the job is fired before the process is started so I think there is no reason to hold this in memory probably? If it is a timer in the middle of the process instance, then it can be useful to "track" progress, but otherwise...

So what is really the point of saving this in memory is not clear to me right now. Maybe it is serialized somewhere?

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
```diff
@@ -80,7 +81,9 @@ private TimerJobFactoryManager initJobFactoryManager() {
 
     @Override
     public JobHandle scheduleJob(Job job, JobContext ctx, Trigger trigger) {
-        if (ctx instanceof ProcessJobContext) {
+        // start process job context it is not job per session is triggers even before 
+        // the process
```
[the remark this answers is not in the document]
elguardian: pls read my comment explaining this.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
elguardian: pls read my comment explaining this.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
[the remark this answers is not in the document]
MarianMacik: I meant that the comment in the code doesn't make sense to me :)

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
MarianMacik: I understand what you said in the comment. My questions was: "do we need to keep it in memory for normal ProcessJobContext as well? Because by looking at the code, it is only needed for "check duplicated purposes as far as I can tell. See: https://github.com/kiegroup/jbpm/blob/186692eb6c88942a82a754147d4c2eab54939ab9/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java#L86-L98

Just wanted to know what is its purpose. That's all.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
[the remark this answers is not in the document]
elguardian: @MarianMacik try to avoid some schedules twice (for instance, it happens with deadlines )

MarianMacik:
jenkins retest this

tkobayas:
@ederign  I noticed one problem.

Could you run GlobalQuartzDBTimerServiceTest under jbpm-test-coverage?

For example, just one test method GlobalQuartzDBTimerServiceTest.testTimerStartManagerClose() would be fine to detect the problem.

`mvn clean test -Pintegration-tests -Dtest=GlobalQuartzDBTimerServiceTest#testTimerStartManagerClose*`

It results in test errors due to test timeout. 

```
Triggered ...
Triggered ...
Triggered ...
2019-11-05 17:11:45,610 [TestScheduler_Worker-3] INFO  Job default-per-request.StartProcess-Minimal-1 threw a JobExecutionException: 
org.quartz.JobExecutionException: Exception when executing scheduled job
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:376)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
Caused by: java.lang.IllegalStateException: Runtime manager default-per-request is already closed
	at org.jbpm.runtime.manager.impl.PerRequestRuntimeManager.disposeRuntimeEngine(PerRequestRuntimeManager.java:134)
	at org.jbpm.process.core.timer.impl.GlobalTimerService$DisposableCommandService.dispose(GlobalTimerService.java:366)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:95)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:48)
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:357)
	... 2 common frames omitted
java.lang.RuntimeException: No scheduler found for default-per-request-timerServiceId
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:74)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:48)
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:357)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
...
[INFO] Results:
[INFO] 
[ERROR] Errors: 
[ERROR]   GlobalQuartzDBTimerServiceTest.testTimerStartManagerClose:166->Object.wait:-2 » TestTimedOut
[ERROR]   GlobalQuartzDBTimerServiceTest.testTimerStartManagerClose:131 » Runtime Except...
[INFO] 
[ERROR] Tests run: 3, Failures: 0, Errors: 2, Skipped: 0
```

Quartz keeps firing the job even after RuntimeManager.close(). I think it is because GlobalTimerService doesn't delete the jobs on RuntimeManager.close() because timerJobsPerSession doesn't contain the jobs.

https://github.com/kiegroup/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java#L155-L162

elguardian:
@tkobayas strange
https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/jbpm-pullrequests/190/testReport/org.jbpm.test.functional.timer/

tkobayas:
@ederign "Timeout or not" may depend on environment. Rather, a problem is that Quartz job records are not cleaned up. Could you run GlobalQuartzDBTimerServiceTest.testTimerStartManagerClose() on your side?

mareknovotny:
@tkobayas I guess you mistakenly referrences here different guy ;) @elguardian is on this PR not @ederign

MarianMacik:
@tkobayas It works for me on my local machine too. We can merge this.

tkobayas:
oh, I'm sorry, @ederign  ! Yeah, I meant @elguardian  here.

@elguardian  and @MarianMacik  , in order to clarify my point, I added a unit test. Please have a look at the unit test in https://github.com/kiegroup/jbpm/pull/1555

It checks if Quartz tables are cleaned up after RuntimeManager.close().

The unit test passes with current master. But if I apply this PR change (GlobalTimerService.java), the test fails.

Could you kindly look at this and share your thought?

I think this is important because if Quartz tables are not cleaned up, Start Timer would keep being triggered and fail with "No scheduler found for XXX-timerServiceId" (for example, imagine a case where one deployment is stopped in kie-server.war).

elguardian:
good catch @tkobayas

MarianMacik:
@tkobayas good catch. What worries me though is that you have experienced it with the current test suite but the PR check passed and locally it passed as well for me. So maybe it was just a coincidence?

elguardian:
@MarianMacik There wasn't any test prior to his test. (there is no check against the database in any of them AFAIK)

MarianMacik:
@elguardian My understanding is the same. That's why I am asking because he mentioned the `testTimerStartManagerClose` test has failed, which was a part of the build already.

mareknovotny:
@MarianMacik well the test  was not here until @elguardian added it here in https://github.com/kiegroup/jbpm/pull/1552/commits/32dffceb1984c2e32b5b87f8e4e7bed98c6fc855#diff-62ea259a6b9b5508f0e428e227702b90R414 the last update of PR. @tkobayas sent different PR just with that test ;) in #1555

MarianMacik:
@mareknovotny I know, but I am more concerned why the test he posted in the https://github.com/kiegroup/jbpm/pull/1552#issuecomment-549716856 failed only for him. In the PR check and locally it just works.

tkobayas:
Hi, @MarianMacik 

I tested testTimerStartManagerClose on my end again. Note that I use '*' at the end of the test method so it runs 3 times with parameters -- Singleton/PerRequest/PerProcessInstance.

`mvn clean test -Pintegration-tests -Dtest=GlobalQuartzDBTimerServiceTest#testTimerStartManagerClose*`

This is really random but I see that the test fails at least within 5 run (5 * 3 = 15 times). 

Also I noticed that it is not related to the PR fix (sorry!). Even without the change, I can reproduce the error. I'm not sure if the fix affects the probability of reproduction.

Could you give it a try again?

The error starts from this Exception:

```
2019-11-06 22:16:06,404 [TestScheduler_Worker-3] INFO  Job default-per-pinstance.StartProcess-Minimal-1 threw a JobExecutionException: 
org.quartz.JobExecutionException: Exception when executing scheduled job
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:376)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
Caused by: java.lang.IllegalStateException: Runtime manager default-per-pinstance is already closed
	at org.jbpm.runtime.manager.impl.PerProcessInstanceRuntimeManager.disposeRuntimeEngine(PerProcessInstanceRuntimeManager.java:222)
	at org.jbpm.process.core.timer.impl.GlobalTimerService$DisposableCommandService.dispose(GlobalTimerService.java:363)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:95)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:48)
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:357)
	... 2 common frames omitted
```

So I think it's a race condition between the working thread which tries to finish the job (= disposing RuntimeEngine) and the main thread which closes the RuntimeManager.

In other words, I think this test case is unstable. I wonder if anyone hit this issue before...?

MarianMacik:
@tkobayas OK, reproduced it once out of 10 runs (30 tests) :)

The issue is as you wrote that the test wants to close the RuntimeManager after the 3rd timer has fired, but the 4th timer has fired and started its job and closing happens only after it. In that case, `((AbstractRuntimeManager)manager).close(true);` will remove all scheduled jobs, but the one which is already running (not scheduled) won't be removed. So the timer job is run, but cannot use the runtime manager, which is now closed. So it tries again and again and the test times out, because `additionalCopy.shutdown();` will wait for running jobs to complete.

So we have 2 options:
* Prolong the intervals between timers from 1s to 2s. This will increase chances that the 4th timer will be in scheduled state this removed when close is called.
* Use `additionalCopy.forceShutdown();` so in case there is the 4th timer running, the exceptions will be thrown by the timer job, but the test won't time out and will shutdown the timer.

I think both actions will preserve what we are trying to test there - that after you close the RuntimeManager (undeploy the kjar) no additional timers are scheduled and the scheduled ones are removed.

Can you try if prolonging the timer duration works for you? We can start with that I guess.

Wdyt is the better approach?

tkobayas:
Thanks, @MarianMacik

I prolonged the timer duration from 1 sec to 2 sec but I still hit the error.

I think that the problematic worker thread is processing the 3rd timer (not 4th). You can confirm it by enabling DEBUG log for "org.quartz". I also print the thread name in process's Script Task.

```
2019-11-07 15:44:48,832 [TestScheduler_Worker-3] DEBUG Calling execute on job default-per-pinstance.StartProcess-Minimal-1
2019-11-07 15:44:48,837 [TestScheduler_QuartzSchedulerThread] DEBUG batch acquisition of 0 triggers
TestScheduler_Worker-3 : Triggered ...
2019-11-07 15:44:48,891 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' is desired by: TestScheduler_Worker-3
2019-11-07 15:44:48,891 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' is being obtained: TestScheduler_Worker-3
2019-11-07 15:44:48,891 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' given to: TestScheduler_Worker-3
2019-11-07 15:44:48,892 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' returned by: TestScheduler_Worker-3
2019-11-07 15:44:48,893 [TestScheduler_QuartzSchedulerThread] DEBUG batch acquisition of 0 triggers
2019-11-07 15:44:48,893 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' is desired by: TestScheduler_Worker-3
2019-11-07 15:44:48,893 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' is being obtained: TestScheduler_Worker-3
2019-11-07 15:44:48,893 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' given to: TestScheduler_Worker-3
2019-11-07 15:44:48,894 [TestScheduler_Worker-3] DEBUG Lock 'TRIGGER_ACCESS' returned by: TestScheduler_Worker-3
2019-11-07 15:44:48,895 [TestScheduler_QuartzSchedulerThread] DEBUG batch acquisition of 0 triggers
2019-11-07 15:44:48,899 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is desired by: Time-limited test
2019-11-07 15:44:48,899 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is being obtained: Time-limited test
2019-11-07 15:44:48,899 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' given to: Time-limited test
2019-11-07 15:44:48,900 [Time-limited test] DEBUG Deleting job: default-per-pinstance.StartProcess-Minimal-1
2019-11-07 15:44:48,900 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' returned by: Time-limited test
2019-11-07 15:44:48,900 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is desired by: Time-limited test
2019-11-07 15:44:48,900 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is being obtained: Time-limited test
2019-11-07 15:44:48,900 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' given to: Time-limited test
2019-11-07 15:44:48,902 [Time-limited test] DEBUG Deleting job: default-per-pinstance.StartProcess-Minimal-1
2019-11-07 15:44:48,902 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' returned by: Time-limited test
2019-11-07 15:44:48,903 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is desired by: Time-limited test
2019-11-07 15:44:48,903 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is being obtained: Time-limited test
2019-11-07 15:44:48,903 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' given to: Time-limited test
2019-11-07 15:44:48,903 [Time-limited test] DEBUG Deleting job: default-per-pinstance.StartProcess-Minimal-1
2019-11-07 15:44:48,903 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' returned by: Time-limited test
2019-11-07 15:44:48,904 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is desired by: Time-limited test
2019-11-07 15:44:48,904 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' is being obtained: Time-limited test
2019-11-07 15:44:48,904 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' given to: Time-limited test
2019-11-07 15:44:48,904 [Time-limited test] DEBUG Deleting job: default-per-pinstance.StartProcess-Minimal-1
2019-11-07 15:44:48,904 [Time-limited test] DEBUG Lock 'TRIGGER_ACCESS' returned by: Time-limited test
2019-11-07 15:44:48,905 [TestScheduler_QuartzSchedulerThread] DEBUG batch acquisition of 0 triggers
2019-11-07 15:44:48,906 [TestScheduler_QuartzSchedulerThread] DEBUG batch acquisition of 0 triggers
2019-11-07 15:44:49,828 [TestScheduler_QuartzSchedulerThread] DEBUG batch acquisition of 0 triggers
2019-11-07 15:44:49,914 [TestScheduler_Worker-3] INFO  Job default-per-pinstance.StartProcess-Minimal-1 threw a JobExecutionException: 
org.quartz.JobExecutionException: Exception when executing scheduled job
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:376)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
Caused by: java.lang.IllegalStateException: Runtime manager default-per-pinstance is already closed
	at org.jbpm.runtime.manager.impl.PerProcessInstanceRuntimeManager.disposeRuntimeEngine(PerProcessInstanceRuntimeManager.java:222)
	at org.jbpm.process.core.timer.impl.GlobalTimerService$DisposableCommandService.dispose(GlobalTimerService.java:363)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:95)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:48)
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:357)
	... 2 common frames omitted
```

The problem is that the test method calls RuntimeManager.close() immediately after countDownListener.waitTillCompleted() which is still inside the process execution. So the runtimeManger is closed before the worker thread finishes the job. 

```
        countDownListener.waitTillCompleted();
        manager.disposeRuntimeEngine(runtime);
        int atDispose = timerExporations.size();
        assertTrue(atDispose > 0);

        ((AbstractRuntimeManager)manager).close(true);
```

If I add Thread.sleep(500) right after countDownListener.waitTillCompleted(), the issue doesn't occur (tested 10 run).

```
        countDownListener.waitTillCompleted();
        Thread.sleep(500);
        manager.disposeRuntimeEngine(runtime);
        int atDispose = timerExporations.size();
        assertTrue(atDispose > 0);

        ((AbstractRuntimeManager)manager).close(true);
```

Thread.sleep() is not elegant but I haven't come up with other ideas so far.

WDYT?

elguardian:
Jenkins restest his please

MarianMacik:
@tkobayas 

I think we experience 2 different scenarios here:

1. As you wrote in the https://github.com/kiegroup/jbpm/pull/1552#issuecomment-549716856 the exception and the message is: 

```
2019-11-05 17:11:45,610 [TestScheduler_Worker-3] INFO  Job default-per-request.StartProcess-Minimal-1 threw a JobExecutionException: 
org.quartz.JobExecutionException: Exception when executing scheduled job
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:376)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
Caused by: java.lang.IllegalStateException: Runtime manager default-per-request is already closed
	at org.jbpm.runtime.manager.impl.PerRequestRuntimeManager.disposeRuntimeEngine(PerRequestRuntimeManager.java:134)
	at org.jbpm.process.core.timer.impl.GlobalTimerService$DisposableCommandService.dispose(GlobalTimerService.java:366)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:95)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:48)
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:357)
	... 2 common frames omitted
java.lang.RuntimeException: No scheduler found for default-per-request-timerServiceId
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:74)
	at org.jbpm.persistence.timer.GlobalJpaTimerJobInstance.call(GlobalJpaTimerJobInstance.java:48)
	at org.jbpm.process.core.timer.impl.QuartzSchedulerService$QuartzJob.execute(QuartzSchedulerService.java:357)
	at org.quartz.core.JobRunShell.run(JobRunShell.java:202)
	at org.quartz.simpl.SimpleThreadPool$WorkerThread.run(SimpleThreadPool.java:573)
```

The java.lang.RuntimeException exception is strange - this cannot happen with the third timer because the message 'No scheduler found for default-per-request-timerServiceId' is thrown before the process instance itself is started (and printing Triggering...) and because the test thread waits for 3rd process to be started, RuntimeManager must have been available at that time. But maybe it was just a copy&paste mistake from other run.

This is what happens to me most of the time, but with Worker 4 - which means when the 4th timer is being executed. I think the behaviour in this situation is correct - the RuntimeManager is closed (simulating the kjar being undeployed) and the currently running timer job cannot find it. It fails even after retries. Should a user redeploy the kjar again, the timers will be recreated, though I am not sure if they will overwrite the old (unfinished in case their jobs were running during the undeployment) ones from the previous deployment and new process instances will be started (their Quartz job name ends with timerId which is unique, so I guess they will remain in db). But still I guess no scheduled timers will be left in the db after redeploy because the ones from the previous deployment will be unscheduled after retry count has been reached. What happens if you manage to redeploy the kjar when the retries are executed is easy - the process will be started. So I guess this is expected when dealing with these scenarios.

2. The second scenario is more tricky. It happened to me few times and I have probably found the issue. When you look here: https://github.com/kiegroup/jbpm/blob/ee3941e40c00daebf87f5328a8bddfc8e9f53256/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/timer/GlobalJpaTimerJobInstance.java#L77-L100

There is `startTxIfNeeded` for starting the tx and `closeTansactionIfNeeded` for ending it. And when you look at `startTxIfNeeded` implementation: https://github.com/kiegroup/jbpm/blob/ee3941e40c00daebf87f5328a8bddfc8e9f53256/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/timer/GlobalJpaTimerJobInstance.java#L139-L158

With plain jBPM engine and Quartz, it always returns null. And because of that, the transaction spans just the operation inside the `runner.execute( command );` (line 81 in GlobalJpaTimerJobInstance), i.e. inside the engine. In that case the commit is done before the `GlobalJpaTimerJobInstance` is finished, so before the Runtime Engine disposal. Outside the tests the scenario still works because even if the disposal in the finally block fails, the engine has already started the process instance, it's just the enclosing job which cannot finish properly. But there is one thing which is wrong right now I think - inside that `runner.execute( command );` call the next timer trigger, if exists, is already scheduled. So it might happen that the next timer has already been scheduled but the current execution of the timer will fail because the disposal happens outside of the tx and those 2 methods (startTxIfNeeded, closeTansactionIfNeeded) are no-op.

@mswiderski can you please confirm or correct my findings? @tkobayas what do you think of it?

I don't understand why these lines are there:
https://github.com/kiegroup/jbpm/blob/ee3941e40c00daebf87f5328a8bddfc8e9f53256/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/timer/GlobalJpaTimerJobInstance.java#L145-L151

When we have EJB Timers, we are already inside the transaction so the methods (like tm.begin()) are no-op. If we are in the application server with CMT and a user has properly configured engine to use `ContainerManagedTransactionManager` then we are fine as well - we are again inside the transaction already. But if we use just plain engine with Quartz, this if will never be triggered. I tried to remove the if and call
```
TransactionManager tm = TransactionManagerFactory.get().newTransactionManager();
    			
    			if (tm.begin()) {    			
    				return tm;
    			}
```

always and the Quartz tests work. I think we are safe to call this every time because in case there is already a tx open, `tm.begin()` will be simply no-op and return false.

What do you think?

tkobayas:
Hi @MarianMacik

I think this issue is not related to the PR fix made by @elguardian . So can we split this topic to another JIRA and discuss it there? I have filed a JIRA.

https://issues.jboss.org/browse/JBPM-8889

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
```diff
@@ -409,4 +412,81 @@ public void beforeProcessStarted(ProcessStartedEvent event) {
...
+        manager.disposeRuntimeEngine(runtime);
+        int atDispose = timerExporations.size();
+        assertTrue(atDispose > 0);
+
+        ((AbstractRuntimeManager) manager).close(true);
+        countDownListener.reset(1);
+        countDownListener.waitTillCompleted(3000);
+        assertEquals(atDispose, timerExporations.size());
+
+        ksession.destroy();
```
tkobayas: @elguardian , regarding the recent error, https://rhba-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/KIE/job/master/job/pullrequest/job/jbpm-pullrequests/195/testReport/org.jbpm.test.functional.timer/GlobalQuartzDBTimerServiceTest/testQuartzJobDeletionOnManagerCloseWithTimerStart_1_/

We don't need ksession.destroy() here because it's already destroyed by manager.disposeRuntimeEngine()? (in case of PerRequest)

elguardian:
Jenkins retest this please.

elguardian:
failures not related...
Jenkins retest this please.

elguardian:
Jenkins retest this please.

elguardian:
Jenkins retest this please.

elguardian:
@mswiderski can we merge this ? so I can cherry pick to 7.30.x ?

mareknovotny:
@elguardian i can do that as @mswiderski  approved it already

elguardian:
this was merged but for some  unknown reason still opened.

PR #1554, opened 2019-11-05
RHPAM-2422: Avoid NPE for cases with null name nodes and boundary timers inside stage

gmunozfe:
Issue happened when there were nodes without name, inside a stage. 
After boundary timer is fired, _resolveVariable_ method may receive a null string in the first param, due to process contains nodes without setting their names.
Added check at **DynamicNode** class, _acceptEvent_ method with resolver function, and test to validate it.


elguardian:
@mswiderski @MarianMacik can we merge this and clear this out  from the backlog ?

PR #1555, opened 2019-11-06
Adding a unit test to check Quartz table cleanup after RuntimeManager…

tkobayas:
….close()

I sent this PR in association with PR https://github.com/kiegroup/jbpm/pull/1552

I think this unit test is valid in itself but it's better to discuss in PR 1552.


mareknovotny:
I am closing now as the test is included in #1552

PR #1558, opened 2019-11-12
[JBPM-8895] case management - stage boundary timer does not trigger

elguardian:
Dynamic nodes can have timers associated

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
@@ -16,17 +16,6 @@
 
 package org.jbpm.casemgmt.impl;
 
-import static java.util.stream.Collectors.toMap;
-import static org.assertj.core.api.Assertions.assertThat;
-import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
-import static org.assertj.core.api.Assertions.entry;
-import static org.assertj.core.api.Assertions.fail;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertFalse;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
-import static org.junit.Assert.assertTrue;
-
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -67,6 +56,7 @@ import org.jbpm.casemgmt.api.model.instance.StageStatus;
 import org.jbpm.casemgmt.impl.model.instance.CaseInstanceImpl;
 import org.jbpm.casemgmt.impl.objects.EchoService;
 import org.jbpm.casemgmt.impl.util.AbstractCaseServicesBaseTest;
+import org.jbpm.casemgmt.impl.util.CountDownListenerFactory;
 import org.jbpm.document.Document;
 import org.jbpm.document.service.impl.DocumentImpl;
 import org.jbpm.services.api.TaskNotFoundException;
@@ -75,7 +65,10 @@ import org.jbpm.services.api.model.ProcessInstanceDesc;
 import org.jbpm.services.api.model.VariableDesc;
 import org.jbpm.services.task.impl.model.GroupImpl;
 import org.jbpm.services.task.impl.model.UserImpl;
+import org.jbpm.test.listener.process.NodeLeftCountDownProcessEventListener;
 import org.jbpm.workflow.instance.node.MilestoneNodeInstance;
+import org.junit.After;
+import org.junit.Before;
 import org.junit.Test;
 import org.kie.api.command.ExecutableCommand;
 import org.kie.api.event.rule.MatchCreatedEvent;
@@ -97,9 +90,21 @@ import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import static java.util.stream.Collectors.toMap;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.assertj.core.api.Assertions.entry;
+import static org.assertj.core.api.Assertions.fail;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
 public class CaseServiceImplTest extends AbstractCaseServicesBaseTest {
 
     private static final Logger logger = LoggerFactory.getLogger(CaseServiceImplTest.class);
+
 
     @Override
     protected List<String> getProcessDefinitionFiles() {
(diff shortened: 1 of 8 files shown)


elguardian:
@gmunozfe can you have a look ?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -79,6 +81,12 @@ public void internalTrigger(NodeInstance from, String type) {
         if (getNodeInstanceContainer().getNodeInstance(getId()) == null) {
             return;
         }
+
+        registerBoundayEvents();
+
+    }
+
+    protected void registerBoundayEvents() {
```
gmunozfe: Typo, rename it to "registerBoundaryEvents"

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -97,8 +105,9 @@ public void internalTrigger(NodeInstance from, String type) {
 
             for (String name : getEventBasedNode().getBoundaryEvents()) {
 
-                boolean isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime().getAgenda())
-                        .isRuleActiveInRuleFlowGroup("DROOLS_SYSTEM", name, getProcessInstance().getId());
+                boolean isActive = ((InternalAgenda) getProcessInstance().getKnowledgeRuntime()
```
gmunozfe: There is no change here, keep it as before for a clear history

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -43,10 +43,12 @@
 import org.jbpm.process.instance.ProcessInstance;
 import org.jbpm.process.instance.context.variable.VariableScopeInstance;
 import org.jbpm.process.instance.impl.Action;
+import org.jbpm.process.instance.impl.JavaScriptAction;
```
gmunozfe: Not needed

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -43,10 +43,12 @@
...
 import org.jbpm.process.instance.timer.TimerInstance;
 import org.jbpm.process.instance.timer.TimerManager;
 import org.jbpm.util.PatternConstants;
 import org.jbpm.workflow.core.DroolsAction;
+import org.jbpm.workflow.core.impl.DroolsConsequenceAction;
```
gmunozfe: Not needed

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -3730,4 +3735,48 @@ public void testCaseWithStageAndBoundaryTimerFired() throws InterruptedException
...
+        ByteArrayOutputStream outContent = new ByteArrayOutputStream();
+        PrintStream originalOut = System.out;
+
+        System.setOut(new PrintStream(outContent));
+
+        String caseId = caseService.startCase(deploymentUnit.getIdentifier(), "CaseWithBoundaryTimerStage");
+        assertNotNull(caseId);
+        assertEquals(FIRST_CASE_ID, caseId);
+        
+        Thread.sleep(3000);
```
mswiderski: we should use event listeners with countdown latch instead of thread sleep. See other timer related tests.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -3730,4 +3735,48 @@ public void testCaseWithStageAndBoundaryTimerFired() throws InterruptedException
...
+            CaseStageInstance stage1 = iterator.next();
+            assertThat(stage1.getName()).isEqualTo("Stage 1");
+            assertThat(stage1.getStatus()).isEqualTo(StageStatus.Completed);
+
+            CaseStageInstance stage2 = iterator.next();
+            assertThat(stage2.getName()).isEqualTo("Stage 2");
+            assertThat(stage2.getStatus()).isEqualTo(StageStatus.Active);
+            
+            assertThat(outContent.toString(), containsString("timer on stage"));
+            assertThat(outContent.toString(), containsString("timer on task"));
```
mswiderski: I don't think we should reply on this, as mentioned above, using event listeners we can ensure that the timer (based on node info) has been triggered so no need to use the captured system out.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
gmunozfe: Thanks @mswiderski . 
Fixed at @elguardian fork for being merged there

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
gmunozfe: Ok, thanks, fixed with the same PR at Enrique's fork

elguardian:
done @mswiderski

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -3699,14 +3705,31 @@ public void testCaseWithIndexedFromVariableUpdatesResetVariable() {
...
-    
+
+    @Before
+    public void setUp() throws Exception {
+        registerListenerMvelDefinition("org.jbpm.casemgmt.impl.util.CountDownListenerFactory.get(\"" +
+                                       getClass().getSimpleName() + "1\", " +
+                                       "\"end 1\", 1)");
+        registerListenerMvelDefinition("org.jbpm.casemgmt.impl.util.CountDownListenerFactory.get(\"" +
+                                       getClass().getSimpleName() + "2\", " +
+                                       "\"Stage 1\", 1)");
```
gmunozfe: For better test performance, perhaps this can be done overriding getProcessListeners, and registering based on test method name. This was my proposal (overlapped with yours):
https://github.com/elguardian/jbpm/pull/4/commits/461ce872c916a3c0ab81b3cb80a7b3ef41e8e9e8

elguardian: that solution is a bit more confusing and error prone.

elguardian:
@kiegroup/gatekeepers can anybody merge ?

PR #1560, opened 2019-11-14
[DROOLS-4755] Support easy form url encoding on REST Requests

exabrial:
It's surprisingly difficult to send an object as a group of form url encoded parameters. This seems like a common use case (POST a bunch of key-value-pairs to a URL). 

This patch allows the use to set the `content-type` to `application/x-www-form-urlencoded` and the payload object will be automatically form url encoded.

### DIFF
--- a/jbpm-workitems/jbpm-workitems-rest/pom.xml
+++ b/jbpm-workitems/jbpm-workitems-rest/pom.xml
@@ -55,6 +55,33 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+        <groupId>com.github.exabrial</groupId>
+        <artifactId>form-binding</artifactId>
+        <!-- These are included with kie-server -->
+        <exclusions>
+            <exclusion>
+                <groupId>commons-logging</groupId>
+                <artifactId>commons-logging</artifactId>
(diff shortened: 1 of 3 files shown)


mswiderski:
@exabrial just wondering if this wouldn't be a nicer fit to be placed as separate work item in jbpm-work-items repository instead and by that to become a service task that can be easily installed via service task repository (http://mswiderski.blogspot.com/2019/01/service-task-repository-integrated-into.html), wdyt?

@MarianMacik @tsurdilo what do you guys think?

MarianMacik:
@mswiderski Completely agree. By doing that, we will keep the current REST WIH as lightweight as possible with additional features added by separate work items only in case they are needed.

exabrial:
Thank you for the feedback!

I'm fine with either, but I'll have to copy/paste the existing REST handler, and then we'll have two REST work item handlers. I think it's a pretty common use cast to POST a form to a service (nearly every OAuth protected RESTful service requires this), which is why I wrote it here.

mswiderski:
and why not to extend it and just override the methods that requires this feature?

exabrial:
Thanks, we can do that.

PR #1565, opened 2019-11-21
JBPM-8781 - API to view all of sub-cases belong to a parent case - re…

MarianMacik:
…solve ambiguous column

This should fix the KIE Server failing tests.


MarianMacik:
Fixed: kiegroup/droolsjbpm-integration#1945

PR #1566, opened 2019-11-22
RHPAM-2526 - Lots of queries against ContextMappingInfo when satisfyi…

mswiderski:
…ng Milestone


elguardian:
1 test case failure irrelevant. The other two might be relevant to this case.

elguardian:
I think the solution is less safer than bulk loading... I am going to try with different approach (just in case). In this case we are trading off one bug for another and causing the code to be less uniform in execution; To be onest I cannot think a proper way to bypass the problem caused by this. We can talk tomorrow (I have another proposal)

elguardian:
@mswiderski what about something like this #1570 ? (just a single query for loading all the context mapping info). it preserves the current logic regarding storing things in retrieving from database.

elguardian:
superseeded by https://github.com/kiegroup/jbpm/pull/1572

PR #1568, opened 2019-11-22
Fix tests for postgresql ddl scripts

gmunozfe:
Following issues during ddl script execution for postgresql have been sorted out:
- Delimiter semicolon has to be kept without splitting if found inside a dollar quoted block
- schema sql has to be executed prior to trigger sql
- missing drop table jbpm_active_clob

### DIFF
--- a/jbpm-installer/src/main/resources/db/ddl-scripts/postgresql/postgresql-jbpm-lo-trigger-clob.sql
+++ b/jbpm-installer/src/main/resources/db/ddl-scripts/postgresql/postgresql-jbpm-lo-trigger-clob.sql
@@ -1,3 +1,4 @@
+drop table if exists jbpm_active_clob;
 create table jbpm_active_clob ( loid oid );
 
 -- Triggers to protect CLOB from vacuumlo
--- a/jbpm-installer/src/test/java/org/jbpm/persistence/scripts/util/TestsUtil.java
+++ b/jbpm-installer/src/test/java/org/jbpm/persistence/scripts/util/TestsUtil.java
@@ -62,6 +62,11 @@ public final class TestsUtil {
                         return o1.getName().compareTo(o2.getName());
                     }
                 });
+                
+                if (databaseType.equals(DatabaseType.POSTGRESQL)) {
+                    //Returns first schema sql
+                    Arrays.sort(foundFiles, Comparator.<File, Boolean>comparing(s -> s.getName().contains("schema")).reversed());
+                }
             }
 
             return foundFiles;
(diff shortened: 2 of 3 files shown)


jbpm-installer/src/main/resources/db/ddl-scripts/postgresql/postgresql-jbpm-lo-trigger-clob.sql
```diff
@@ -1,3 +1,4 @@
+drop table if exists jbpm_active_clob cascade;
```
MarianMacik: Cascade shouldn't be needed here.

jbpm-installer/src/test/java/org/jbpm/persistence/scripts/util/TestsUtil.java
```diff
@@ -62,6 +62,11 @@ public int compare(File o1, File o2) {
                         return o1.getName().compareTo(o2.getName());
                     }
                 });
+                
+                if (databaseType.equals(DatabaseType.POSTGRESQL)) {
+                    //Returns first schema sql
+                    Arrays.sort(foundFiles, Comparator.<File, Boolean>comparing(s -> s.getName().contains("schema")).reversed());
+                }
```
MarianMacik: Is there any need for this sorting? What happens if we run `lo-trigger-clob` and `quartz` scripts before the schema script? Right now the order should be:
`postgresql-jbpm-lo-trigger-clob.sql`
`postgresql-jbpm-schema.sql`
`quartz_tables_postgres.sql`

and everything works fine.

gmunozfe: With that order, following exception is thrown, because that table is created in the "postgresql-jbpm-schema.sql":
`ERROR: relation "booleanexpression" does not exist`

jbpm-installer/src/main/resources/db/ddl-scripts/postgresql/postgresql-jbpm-lo-trigger-clob.sql
gmunozfe: Fixed, thanks!

jbpm-installer/src/test/java/org/jbpm/persistence/scripts/util/TestsUtil.java
MarianMacik: Ah, OK, makes sense.

PR #1570, opened 2019-11-26
[RHPAM-2526] Lots of queries against ContextMappingInfo when satisfying Milestone

elguardian:
depens on https://github.com/kiegroup/droolsjbpm-knowledge/pull/411


elguardian:
testing another solution for this jira

elguardian:
Jenkins retest this please

mswiderski:
@elguardian since the issue originates from case mgmt I am thinking if we should somehow scoped the signal name - using external event listener for milestones but then make it add a case id into the signal name. That way the returned list of process instances will only be for given case instance. Wdyt?

elguardian:
I will have a look at your idea.

MarianMacik:
Can we close this PR?

PR #1572, opened 2019-12-05
[RHPAM-2526] Lots of queries against ContextMappingInfo when satisfying Milestone

elguardian:
Modify the activation type for milestone so it is unique among cases

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
@@ -436,10 +436,16 @@ public class ProcessRuntimeImpl implements InternalProcessRuntime {
                         kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(eventType, event));
                     } else if (ruleName.startsWith( "RuleFlowStateEventSubProcess-" ) 
                             || ruleName.startsWith( "RuleFlowStateEvent-" ) 
-                            || ruleName.startsWith( "RuleFlow-Milestone-" ) 
                             || ruleName.startsWith( "RuleFlow-AdHocComplete-" )
                             || ruleName.startsWith( "RuleFlow-AdHocActivate-" )) {
                         kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(ruleName, event));
+                    } else if (ruleName.startsWith("RuleFlow-Milestone-")) {
+                        String caseId = (String) kruntime.getEnvironment().get(EnvironmentName.CASE_ID);
+                        if (caseId != null) {
+                            kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(ruleName + "-" + caseId, event));
+                        } else {
+                            kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(ruleName, event));
+                        }
                     }
                 } else {
                     String ruleName = event.getMatch().getRule().getName();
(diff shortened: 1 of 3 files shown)


elguardian:
this is ok to test

elguardian:
this ok to test. (check if this breaks something)

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -436,10 +436,12 @@ public void matchCreated(MatchCreatedEvent event) {
                         kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(eventType, event));
                     } else if (ruleName.startsWith( "RuleFlowStateEventSubProcess-" ) 
                             || ruleName.startsWith( "RuleFlowStateEvent-" ) 
-                            || ruleName.startsWith( "RuleFlow-Milestone-" ) 
                             || ruleName.startsWith( "RuleFlow-AdHocComplete-" )
                             || ruleName.startsWith( "RuleFlow-AdHocActivate-" )) {
                         kruntime.queueWorkingMemoryAction(new SignalManagerSignalAction(ruleName, event));
+                    } else if (ruleName.startsWith("RuleFlow-Milestone-")) {
+                        String caseId = (String) kruntime.getEnvironment().get(EnvironmentName.CASE_ID);
```
mswiderski: I'd add check if `kruntime.getEnvironment().get(EnvironmentName.CASE_ID)` is not null before appending it

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/MilestoneNodeInstance.java
```diff
@@ -82,9 +83,9 @@ public void removeEventListeners() {
         getProcessInstance().removeEventListener(getActivationEventType(), this, true);
     }
 
-    private String getActivationEventType() {
+    public String getActivationEventType() {
         return "RuleFlow-Milestone-" + getProcessInstance().getProcessId()
-                + "-" + getMilestoneNode().getUniqueId();
+               + "-" + getMilestoneNode().getUniqueId() + "-" + ((WorkflowProcessInstanceImpl) getProcessInstance()).getCorrelationKey();
```
mswiderski: similar here, append correlation key only when it is running in case context

elguardian:
this is ok to test.

elguardian:
test failure not related.
Jenkins retest this please.

elguardian:
@mswiderski happy with the tests. if you are happy we should proceed to merge.

mswiderski:
looks good to me, though I'd like to hear @MarianMacik feedback and maybe we should run the full downstream build or at least droolsjbpm-integration repo tests.

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
```diff
@@ -436,10 +436,16 @@ public void matchCreated(MatchCreatedEvent event) {
...
+                        if (caseId != null) {
```
MarianMacik: When can it happen that case id is null?

mswiderski: when you run with different runtime strategy than PER_CASE

elguardian:
@kiegroup/gatekeepers can anyone merge this if it is ok ?

jbpm-flow/src/main/java/org/jbpm/process/instance/ProcessRuntimeImpl.java
MarianMacik: Ah, so when you use milestone in a process, not case?

mswiderski: Correct as this is then expected behavior as we have no other context to bind to

PR #1578, opened 2020-01-08
[JBPM-8896] NPE during Process Migration when Boundary Timer is fired but UserTask not completed

elguardian:
self remove timer if there is no next boundary event in human task

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
@@ -32,6 +32,8 @@ import org.drools.core.rule.Declaration;
 import org.drools.core.spi.Activation;
 import org.drools.core.time.TimeUtils;
 import org.drools.core.time.impl.CronExpression;
+import org.drools.core.time.impl.DefaultJobHandle;
+import org.drools.core.time.impl.TimerJobInstance;
 import org.drools.core.util.MVELSafeHelper;
 import org.jbpm.process.core.ContextContainer;
 import org.jbpm.process.core.context.variable.Variable;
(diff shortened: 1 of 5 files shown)


elguardian:
this is ok to test.

elguardian:
Jenkins restest this please.

gmunozfe:
Both tests (org.jbpm.integrationtests.ProcessStateTest.testTimerState & org.jbpm.integrationtests.ProcessTimerTest.testOnEntryTimerExecutedMultipleTimes) are failing because the timer is removed though it's expected to be triggered more times, due to their period is not zero.

Maybe it can be added this condition to remove only timers without defined period:
`if (handler.getNext() == null && timerInstance.getPeriod() == 0L) `

elguardian:
Failure unrelated.
Jenkins retest this please.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/StateBasedNodeInstance.java
```diff
@@ -345,9 +347,20 @@ private void triggerTimer(TimerInstance timerInstance) {
         for (Map.Entry<Timer, DroolsAction> entry : getEventBasedNode().getTimers().entrySet()) {
             if (entry.getKey().getId() == timerInstance.getTimerId()) {
                 executeAction((Action) entry.getValue().getMetaData("Action"));
+
+                // self remove timer instance from this node as it will remove itself 
+                // from timer service once is triggered and there is no next
+                if (!(timerInstance.getJobHandle() instanceof DefaultJobHandle)) {
+                    return;
+                }
+                DefaultTimerJobInstance defaultTimerInstance = (DefaultTimerJobInstance) ((DefaultJobHandle) timerInstance.getJobHandle()).getTimerJobInstance();
```
gmunozfe: Is the casting to DefaultTimerJobInstance needed? For example, replacing _defaultTimerInstance_ in the next line by 
_((DefaultJobHandle) timerInstance.getJobHandle()).getTimerJobInstance()_

elguardian: yes it is needed JobHandle does not have method getTimerJobInstance

gmunozfe: I mean `TimerJobInstance` interface already has `getTrigger` method, so there's no need to cast to `DefaultTimerJobInstance`, if you don't create an intermediate object (but it's not important, just nitpicking)

```suggestion
if (((DefaultJobHandle) timerInstance.getJobHandle()).getTimerJobInstance().getTrigger().hasNextFireTime() == null) {
```

gmunozfe:
@elguardian I've created a test with Quartz timers and springboot, and it's failing because jobHandler is null, so timer is not removed

[droolsjbpm-integration#1980](https://github.com/kiegroup/droolsjbpm-integration/pull/1980)

gmunozfe:
It seems droolsjbpm-integration test with Quartz timers is working with this new commit. Thanks @elguardian !

gmunozfe:
jenkins retest this

gmunozfe:
This is ok to merge

elguardian:
@kiegroup/gatekeepers can you merge ?

PR #1581, opened 2020-01-16
RHPAM-2556 - Case reopen capability

mswiderski:


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
```diff
@@ -79,20 +79,22 @@ public Void execute(Context context) {
...
         }
-        ksession.update(factHandle, caseFile);
+        
         
         logger.debug("Starting process instance for case {} and case definition {}", caseId, caseDefinitionId);
         CorrelationKey correlationKey = correlationKeyFactory.newCorrelationKey(caseId);
         Map<String, Object> params = new HashMap<>();
         // set case id to allow it to use CaseContext when creating runtime engine
         params.put(EnvironmentName.CASE_ID, caseId);
         long processInstanceId = processService.startProcess(deploymentId, caseDefinitionId, correlationKey, params);
```
gmunozfe: Hi @mswiderski 
There are 3 tests failing in _CaseServiceImplTest_ class:

_testStartThenReopenEmptyCase
testStartThenReopenEmptyCaseUpdateData
testStartThenReopenEmptyCaseUpdateDescription_

_expected:<[my first case]> but was:<[#{name}]>_

It seems like when starting the process, {name} variable has no value

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
```diff
@@ -79,20 +79,22 @@ public Void execute(Context context) {
...
+        ksession.insert(caseFile);
```
gmunozfe: Shouldn't _caseFile_ be inserted  before starting the process?

mswiderski: no, that's the thing if it is then it will fire rules before any of the listeners is triggered

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
mswiderski: working on a fix, thanks for checking on it

mswiderski: pushed a fix so let's see how it goes in Jenkins now...

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
gmunozfe: yes, I've tested that approach, but some assertions are failing, because without the caseFile_ prefix, a new variable is created, so expecting 3 instead of 4:
caseFile_name vs name (new one).

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
gmunozfe: I've tested that approach, but some assertions are failing, because expecting 3 variables instead of 4 (because this time -inserting it like that- doesn't have caseId_ prefix):
variableId=caseFile_name  vs. variableId=name

gmunozfe: And if add them with the prefix
for (Map.Entry<String, Object> entry : caseData.entrySet()) {
            params.put(VariableScope.CASE_FILE_PREFIX + entry.getKey(), entry.getValue());
        }
Then, getVariable at VariableScopeInstance class, don't transform it as
caseFiles.size() == 0

mswiderski: you're right, I refactored this to suit both needs, have a look again

gmunozfe: thanks Maciej! It seems now it's working (though let Jenkins to finish)

mswiderski:
@gmunozfe @MarianMacik are there any other comments or do we consider this good to be merged?

gmunozfe:
This is OK to merge @mswiderski 
SonarCloud is not calculating correctly the test coverage when more PRs are added.

elguardian:
@kiegroup/gatekeepers merge plz

PR #1583, opened 2020-01-24
Fixing a column name error in the SQL

bbalakriz:
Looking at https://github.com/kiegroup/jbpm/blob/cb8c01be15c79524c0e3218db0f082e67506d404/jbpm-installer/src/main/resources/db/ddl-scripts/mysql5/mysql5-jbpm-schema.sql#L190, the column name for process instance id in ExecutionErrorInfo is PROCESS_INST_ID

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/ExecutionErrorCleanupCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/ExecutionErrorCleanupCommand.java
@@ -115,7 +115,7 @@ public class ExecutionErrorCleanupCommand implements Command, Reoccurring {
 		Map<String, Object> parameters = new HashMap<>();
         StringBuilder cleanUpErrorsQuery = new StringBuilder();
         
-        cleanUpErrorsQuery.append("delete from ExecutionErrorInfo where processInstanceId in "
+        cleanUpErrorsQuery.append("delete from ExecutionErrorInfo where PROCESS_INST_ID in "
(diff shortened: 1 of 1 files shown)


jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/ExecutionErrorCleanupCommand.java
```diff
@@ -115,7 +115,7 @@ public ExecutionResults execute(CommandContext ctx) throws Exception {
 		Map<String, Object> parameters = new HashMap<>();
         StringBuilder cleanUpErrorsQuery = new StringBuilder();
         
-        cleanUpErrorsQuery.append("delete from ExecutionErrorInfo where processInstanceId in "
+        cleanUpErrorsQuery.append("delete from ExecutionErrorInfo where PROCESS_INST_ID in "
```
MarianMacik: Hi, the string here is JPQL (Java Persistence Query Language), not SQL. JPQL operates on Java entities which means that the identifier here should be the name of the Java field:

https://github.com/kiegroup/jbpm/blob/cb8c01be15c79524c0e3218db0f082e67506d404/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/jpa/ExecutionErrorInfo.java#L94-L98

So the query here is correct.

mswiderski:
agree with @MarianMacik and the query is correct. closing this PR. Feel free to open a email thread on the mailing list to describe what is the issue you are running into.

PR #1585, opened 2020-01-27
fixed last http left overs and URLs

mbiarnes:
from this version it seems that some URLs are not accessible any more.
Changed all left over http to https.
Changed https://repository.jboss.org/ to https://origin-repository.jboss.org/ since the origin-repository gets updated immediately.
Script was tested now and it works properly again.

### DIFF
--- a/jbpm-installer/src/main/resources/build.properties
+++ b/jbpm-installer/src/main/resources/build.properties
@@ -18,57 +18,57 @@ release.version=7.33.0-SNAPSHOT
 #   jBPM.version=${snapshot.version}
 #   jBPM.url=https://repository.jboss.org/nexus/service/local/artifact/maven/redirect?r=snapshots&g=org.drools&a=droolsjbpm-bpms-distribution&v=${jBPM.version}&e=zip&c=bin
 jBPM.version=${snapshot.version}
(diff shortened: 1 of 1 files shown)


jbpm-installer/src/main/resources/build.properties
```diff
@@ -18,57 +18,57 @@ release.version=7.33.0-SNAPSHOT
 #   jBPM.version=${snapshot.version}
 #   jBPM.url=https://repository.jboss.org/nexus/service/local/artifact/maven/redirect?r=snapshots&g=org.drools&a=droolsjbpm-bpms-distribution&v=${jBPM.version}&e=zip&c=bin
 jBPM.version=${snapshot.version}
-jBPM.url=https://repository.jboss.org/nexus/service/local/artifact/maven/redirect?r=snapshots&g=org.drools&a=droolsjbpm-bpms-distribution&v=${jBPM.version}&e=zip&c=bin
+jBPM.url=https://origin-repository.jboss.org/nexus/service/local/artifact/maven/redirect?r=snapshots&g=org.drools&a=droolsjbpm-bpms-distribution&v=${jBPM.version}&e=zip&c=bin
```
cristianonicolai: @mbiarnes do we also need to change to _origin_?

MarianMacik: This is probably to avoid proxy issues which JBoss repository have had lately.

PR #1590, opened 2020-02-03
[DROOLS-4999] make jbpm to work with the executable model

mariofusco:


mariofusco:
ok to test

mariofusco:
ok to test

PR #1591, opened 2020-02-05
JBPM-9017 - MI Data Input and Output variable types are not handled c…

MarianMacik:
…orrectly

* Tests for strict validation with a bad process definition will be added if there is no regression.

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
@@ -665,41 +665,26 @@ public class ActivityTest extends JbpmBpmn2TestCase {
 
     @Test
     public void testCallActivityMI() throws Exception {
-        KieBase kbase = createKnowledgeBaseWithoutDumper("BPMN2-CallActivityMI.bpmn2",
-                "BPMN2-CallActivitySubProcess.bpmn2");
-        ksession = createKnowledgeSession(kbase);
-        final List<Long> subprocessStarted = new ArrayList<Long>();
-        ksession.addEventListener(new DefaultProcessEventListener() {
+        runAndAssertMultiInstanceProcess("BPMN2-CallActivityMI.bpmn2");
+    }
 
-            @Override
-            public void beforeProcessStarted(ProcessStartedEvent event) {
-                if (event.getProcessInstance().getProcessId().equals("SubProcess")) {
-                    subprocessStarted.add(event.getProcessInstance().getId());
-                }
-            }
+    @Test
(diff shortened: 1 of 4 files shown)


mswiderski:
@MarianMacik so this is to validate type of the variable on the target node instance like subprocess or task, right? If so it makes perfect sense to add it

Maybe we should also validate on the output collection 
https://github.com/kiegroup/jbpm/blob/9753346b45e674f2ad3e58e6ce3e952d34be1771/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/ForEachNodeInstance.java#L239

Wdyt?

MarianMacik:
> @MarianMacik so this is to validate type of the variable on the target node instance like subprocess or task, right? If so it makes perfect sense to add it

Not sure though what you mean by task. I thought that you can only have multi instance loop only with subprocesses.

But probably yes, it checks that if the input collection contains String elements that the input variable (where each ellement from input collection is assigned to) is of type String as well. Meaning that we don't assign elements of type String to a variable of type Boolean for example.




And btw here:
https://github.com/kiegroup/jbpm/blob/9753346b45e674f2ad3e58e6ce3e952d34be1771/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/ForEachNodeInstance.java#L239

I think the validation is not needed as we just set output collection already full of elements. However, we could check that the output variable, which was created by the subprocess, is of the type defined in the process definition before we add it to the collection, i.e. here:
https://github.com/kiegroup/jbpm/blob/9753346b45e674f2ad3e58e6ce3e952d34be1771/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/ForEachNodeInstance.java#L198

Wdyt?

I think what you have suggested is to check if the type of the collection itself is the same as the one defined in the definition. Currently we use List only in the engine, but in theory it can happen that somebody may try to use something else like Set or so.

And one more thing, in case there is a sequential iteration, we have another set here:

https://github.com/kiegroup/jbpm/blob/9753346b45e674f2ad3e58e6ce3e952d34be1771/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/ForEachNodeInstance.java#L220
But another check here shouldn't be needed I guess because if we check it with the first element and everything is OK, the process definition/instance is still the same, so we don't need to recheck every time I guess.

mswiderski:
> Not sure though what you mean by task. I thought that you can only have multi instance loop only with subprocesses.

you can have individual tasks with multi instance characteristics 

output collection check might be just in case the variable that the output should be mapped to is of different type, but this might be really an edge case so might not be worth to add.

so we can just stay with what is here

mswiderski:
@gmunozfe would you mind to have a quick look?

MarianMacik:
Thanks @gmunozfe!

PR #1592, opened 2020-02-06
RHPAM-2555 - Require the ability to define a set of mandatory case / …

mswiderski:
…process creation variables

@elguardian @gmunozfe would you mind to have a look?

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ package org.jbpm.bpmn2;
+
+
+import static org.assertj.core.api.Assertions.assertThatExceptionOfType;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+
+import org.jbpm.bpmn2.objects.TestWorkItemHandler;
+import org.jbpm.process.core.context.variable.VariableViolationException;
+import org.junit.After;
+import org.junit.BeforeClass;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
+import org.kie.api.KieBase;
+import org.kie.api.runtime.KieSession;
+import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.api.runtime.process.WorkItem;
+
+@RunWith(Parameterized.class)
+ public class VariableTagsTest extends JbpmBpmn2TestCase {
+
+     @Parameters
+     public static Collection<Object[]> persistence() {
+         Object[][] data = new Object[][] { { false }, { true } };
+         return Arrays.asList(data);
+     };
+
+     private KieSession ksession;
+     private KieSession ksession2;
+
+     public VariableTagsTest(boolean persistence) throws Exception {
+         super(persistence);
+     }
+
+     @BeforeClass
+     public static void setup() throws Exception {
+         setUpDataSource();
+     }
+
+     @After
+     public void dispose() {
+         if (ksession != null) {
+             abortProcessInstances(ksession);
+             ksession.dispose();
+             ksession = null;
+         }
+         if (ksession2 != null) {
+             ksession2.dispose();
+             ksession2 = null;
+         }
+     }
+
+     @Test
+     public void testProcessWithMissingRequiredVariable() throws Exception {
+         KieBase kbase = createKnowledgeBase("variable-tags/approval-with-required-variable-tags.bpmn2");
+         KieSession ksession = createKnowledgeSession(kbase);
+         TestWorkItemHandler workItemHandler = new TestWorkItemHandler();
+         ksession.getWorkItemManager().registerWorkItemHandler("Human Task", workItemHandler);
+
+         assertThatExceptionOfType(VariableViolationException.class).isThrownBy(() -> ksession.startProcess("approvals"));
+
+         ksession.dispose();
+     }
+
+     @Test
+     public void testProcessWithRequiredVariable() throws Exception {
+         KieBase kbase = createKnowledgeBase("variable-tags/approval-with-required-variable-tags.bpmn2");
+         KieSession ksession = createKnowledgeSession(kbase);
+         TestWorkItemHandler workItemHandler = new TestWorkItemHandler();
+         ksession.getWorkItemManager().registerWorkItemHandler("Human Task", workItemHandler);
+
+         Map<String, Object> parameters = new HashMap<>();
+         parameters.put("approver", "john");
+
+         ProcessInstance processInstance = ksession.startProcess("approvals", parameters);
+         assertTrue(processInstance.getState() == ProcessInstance.STATE_ACTIVE);
+         ksession = restoreSession(ksession, true);
+         ksession.getWorkItemManager().registerWorkItemHandler("Human Task", workItemHandler);
+         WorkItem workItem = workItemHandler.getWorkItem();
+         assertNotNull(workItem);
+         ksession.getWorkItemManager().completeWorkItem(workItem.getId(), null);
+
+         workItem = workItemHandler.getWorkItem();
+         assertNotNull(workItem);        
+         ksession.getWorkItemManager().completeWorkItem(workItem.getId(), null);
+
+         assertProcessInstanceFinished(processInstance, ksession);
+         ksession.dispose();
+     }
+
+     @Test
+     public void testProcessWithReadonlyVariable() throws Exception {
+         KieBase kbase = createKnowledgeBase("variable-tags/approval-with-readonly-variable-tags.bpmn2");
+         KieSession ksession = createKnowledgeSession(kbase);
+         TestWorkItemHandler workItemHandler = new TestWorkItemHandler();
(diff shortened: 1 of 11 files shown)


jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
```diff
@@ -0,0 +1,136 @@
...
+         WorkItem workItem = workItemHandler.getWorkItem();
+         assertNotNull(workItem);
+         ksession.getWorkItemManager().completeWorkItem(workItem.getId(), null);
+
+         workItem = workItemHandler.getWorkItem();
+         assertNotNull(workItem);        
+         ksession.getWorkItemManager().completeWorkItem(workItem.getId(), null);
+
+         assertProcessInstanceFinished(processInstance, ksession);
+         ksession.dispose();
```
gmunozfe: Not needed

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
```diff
@@ -0,0 +1,136 @@
...
+         ProcessInstance processInstance = ksession.startProcess("approvals", parameters);
+         assertTrue(processInstance.getState() == ProcessInstance.STATE_ACTIVE);        
+         WorkItem workItem = workItemHandler.getWorkItem();
+         assertNotNull(workItem);
+
+         assertThatExceptionOfType(VariableViolationException.class).isThrownBy(() -> ksession.getWorkItemManager().completeWorkItem(workItem.getId(), Collections.singletonMap("ActorId", "john")));
+         ksession.abortProcessInstance(processInstance.getId());
+
+         assertProcessInstanceFinished(processInstance, ksession);
+         ksession.dispose();
```
gmunozfe: Not needed

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableViolationException.java
```diff
@@ -0,0 +1,63 @@
...
+public class VariableViolationException extends RuntimeException {
+
+    private static final long serialVersionUID = 8031225233775014572L;
+
+    private final Long processInstanceId;
+    private final String variableName;
+    private final String errorMessage;
+
+    public VariableViolationException(Long processInstanceId, String variableName, String errorMessage) {
+        super("Variable '" + variableName + "' in process instance '" + (processInstanceId == null ? "unknown" : processInstanceId) + "' violated");
```
gmunozfe: Here it's overwriting the errorMessage passed by the caller.
```suggestion
        super(errorMessage);
```

Currently, it's displaying for case scenarios:

`Variable 'caseFile_s' in process instance '1' violated`

instead of the errorMessage tailored for variables starting with CASE_FILE_PREFIX: 

`Case file item 'caseFile_s' is required but not set`

If the process instance id is useful to be shown, then it may come from the caller itself.

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -162,5 +168,41 @@ public void setContextInstanceContainer(ContextInstanceContainer contextInstance
...
 	}
+    
+    public void enforceRequiredVariables() {
+        VariableScope variableScope = getVariableScope();
+        for (Variable variable : variableScope.getVariables()) {
+            String name = variable.getName();
+            if (variableScope.isRequired(name)) {  
+                // check case file if it is prefixed
+                if (name.startsWith(VariableScope.CASE_FILE_PREFIX)) {
+                    if (!findCaseData(variable.getName())) {
```
gmunozfe: `variable.getName() `is already retrieved at `name`

```suggestion
                    if (!findCaseData(name)) {
```

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -162,5 +168,41 @@ public void setContextInstanceContainer(ContextInstanceContainer contextInstance
...
+                        throw new VariableViolationException(getProcessInstance().getId(), name, "Case file item '" + name + "' is required but not set");
```
gmunozfe: It may be added the processInstanceId into the errorMessage, as this will be pass to the parents.

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -162,5 +168,41 @@ public void setContextInstanceContainer(ContextInstanceContainer contextInstance
...
+                        
+                    }
+                    // otherwise check variables                    
+                } else if (!variables.containsKey(name)) {
+                    throw new VariableViolationException(getProcessInstance().getId(), variable.getName(), "Variable '" + variable.getName() + "' is required but not set");
```
gmunozfe: Same here, `variable.getName() `is already at `name`, and also processInstanceId can be added into the errorMessage

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -162,5 +168,41 @@ public void setContextInstanceContainer(ContextInstanceContainer contextInstance
     		this.variableInstanceIdPrefix = ((CompositeContextNodeInstance) contextInstanceContainer).getUniqueId();
     	}
 	}
+    
+    public void enforceRequiredVariables() {
+        VariableScope variableScope = getVariableScope();
+        for (Variable variable : variableScope.getVariables()) {
+            String name = variable.getName();
+            if (variableScope.isRequired(name)) {
```
gmunozfe: For reducing indentation, it can return if condition is opposite
```suggestion
            if (!variableScope.isRequired(name)) {
              return;
            }  
```

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
```diff
@@ -80,6 +90,9 @@ public void setValue(final Object value) {
 
     public void setMetaData(String name, Object value) {
         this.metaData.put(name, value);
+        if (VARIABLE_TAGS.equals(name) && value != null) {
```
gmunozfe: It seems like "setMetaData" is never called for "customTags" name, so tags is never cached, if this was the purpose of these lines.

elguardian: look at the bpmn2. you will understand the customTags

gmunozfe: My point is that in the PropertyHandler class (correct me if I'm wrong), it's missing some code to read the metaData, and in case of "customTags" is present, then invoke setMetaData with that key.

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableViolationException.java
mswiderski: that's the whole point, the message that comes from parent is generic that indicates about violation, while it has extra errorMessage property that says the specific reason for it.

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
mswiderski: fixed

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
mswiderski: these are all available as dedicated methods and should be used as such instead of bloated message from the generic runtime exception class.

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
mswiderski: fixed

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
mswiderski: it cannot return as it must check all variables

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
mswiderski: this is set via org.jbpm.bpmn2.xml.MetaDataHandler.MetaDataWrapper

gmunozfe: But this condition is never matched:
VARIABLE_TAGS.equals(name)
Is this to cache tags?

gmunozfe:
@mswiderski I think you amended with previous commit and wiped out changes for test class

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
gmunozfe: you're right, I should have said "continue"

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
mswiderski: this is an extra way of setting tags that can be done programatically like fluent api so it should be handled too.

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
mswiderski: if you don't mind I'd prefer to keep just single control flow

mswiderski:
@gmunozfe you're right, I did override it. just fixed and included extra tests, thanks

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableViolationException.java
gmunozfe: ok, my confusion was because in the stacktrace I only see the parent (generic) message

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/Variable.java
gmunozfe: ok, thanks for the clarification

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
gmunozfe: ok

gmunozfe:
jenkins retest this

gmunozfe:
Looks good to me. Thanks @mswiderski !

PR #1593, opened 2020-02-10
RHPAM-2554 - Require the ability to define mandatory process /case va…

mswiderski:
…riables that can only be changed by Admin

depends on
https://github.com/kiegroup/droolsjbpm-knowledge/pull/422
https://github.com/kiegroup/drools/pull/2756 

@gmunozfe @elguardian could you please review?

Overall idea is to register additional event listener that can react on beforeVariableChange event to enforce any security restrictions based on custom tags associated with variables.

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/VariableTagsTest.java
@@ -36,6 +36,8 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 import org.kie.api.KieBase;
+import org.kie.api.event.process.DefaultProcessEventListener;
+import org.kie.api.event.process.ProcessVariableChangedEvent;
 import org.kie.api.runtime.KieSession;
 import org.kie.api.runtime.process.ProcessInstance;
 import org.kie.api.runtime.process.WorkItem;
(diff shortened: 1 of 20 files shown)


mswiderski:
Jenkins retest this

gmunozfe:
@mswiderski there are 3 compilations failures.

At `AddDataCaseFileInstanceCommand`, `ReopenCaseCommand` when invoking `fireAfterVariableChanged`, it's missing the new param (List<String>):

Not sure if there's another easier way to retrieve it:
`VariableScope variableScope = ((VariableScope) ((RegistryContext) context).get(VariableScope.VARIABLE_SCOPE));`

And then, inside the fireAfterVariableChanged call:

`variableScope == null ? null : variableScope.tags(name)`

Last failure is at CachedProcessVariableChangedEvent (jbpm-test-coverage) that missing unimplemented methods:

       @Override
        public List<String> getTags() {
            return Collections.emptyList();
        }

        @Override
        public boolean hasTag(String tag) {
            return false;
        }


If you agree, I can create a PR to your branch.

mswiderski:
@gmunozfe thanks a lot, and go ahead with PR :)

elguardian:
@mswiderski  looks good to me.

gmunozfe:
@mswiderski I've created several tests for case scenarios:

- The way I proposed for retrieving `VariableScope` from Context was not working, the new proposal does.

- For the commands `addDataToCaseFile` and `reopenCase`, only the `fireAfterVariableChanged` was invoked, so there is no validation in these scenarios. I guess `fireBeforeVariableChanged` should also be called, and then these new tests are passing.

I am creating a PR to your branch for you to review if this proposal is valid or not.

elguardian:
Jenkins retest this please

elguardian:
@gmunozfe any progress ?

gmunozfe:
> @gmunozfe any progress ?

sorry, I was trying to test a case with the 3 constraints: required, readonly and restricted.

However, readonly is not validated, because `oldValue` has been already set:

https://github.com/kiegroup/jbpm/blob/b3a1d1ccd7fe513b4d3a1f8b8708285308e12ac3/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L108

Checking why that value is not null, as expected...
^ @mswiderski

gmunozfe:
https://github.com/kiegroup/jbpm/blob/b3a1d1ccd7fe513b4d3a1f8b8708285308e12ac3/jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java#L83

caseFile already has that variable with its value, therefore the setVariable is considering the operation as a modification, violating readonly policy.

mswiderski:
@gmunozfe I would say that this is expected behaviour, since case file variable can be fetched as process variable and since it was marked as readonly it should cause the exception. Or am I misunderstanding something here...

gmunozfe:
OK, thanks @mswiderski I was thinking that "readonly" meant "only can be assigned once"

elguardian:
@gmunozfe @mswiderski is this ready to merge ?

mswiderski:
@elguardian yes, from my point of view

gmunozfe:
@elguardian sure, this is ok to merge

PR #1594, opened 2020-02-14
Update license file name

yurloc:
https://spdx.org/licenses/Apache-2.0.html


MarianMacik:
@yurloc Just a question, why is this needed? ASL is Apache Software License.

yurloc:
"Apache Software License" was the name of the license used in versions 1.x. Version 2.0 is called just the Apache License.

https://opensource.org/licenses/apachepl-1.1.php

MarianMacik:
@yurloc OK, thanks for clarification. Good catch.

PR #1595, opened 2020-02-14
[JBPM-8910] How to use a "OR" query operator in a custom query in the query-parameters

elguardian:
modify some function to allow AST expressions for filtering

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
@@ -16,10 +16,6 @@
 
 package org.jbpm.kie.services.impl.query.preprocessor;
 
-import static org.dashbuilder.dataset.filter.FilterFactory.AND;
-import static org.dashbuilder.dataset.filter.FilterFactory.in;
-import static org.jbpm.services.api.query.QueryResultMapper.COLUMN_POTOWNER;
-
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
@@ -30,10 +26,15 @@ import org.dashbuilder.dataset.DataSetMetadata;
 import org.dashbuilder.dataset.filter.ColumnFilter;
 import org.dashbuilder.dataset.filter.CoreFunctionFilter;
 import org.dashbuilder.dataset.filter.DataSetFilter;
+import org.dashbuilder.dataset.filter.LogicalExprFilter;
 import org.kie.api.task.UserGroupCallback;
 import org.kie.internal.identity.IdentityProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import static org.dashbuilder.dataset.filter.FilterFactory.AND;
+import static org.dashbuilder.dataset.filter.FilterFactory.in;
+import static org.jbpm.services.api.query.QueryResultMapper.COLUMN_POTOWNER;
 
 public class UserAndGroupsTasksPreprocessor extends UserTasksPreprocessor {
 
@@ -65,10 +66,10 @@ public class UserAndGroupsTasksPreprocessor extends UserTasksPreprocessor {
 
         if (lookup.getFirstFilterOp() != null) {
             List<String> potOwners = new ArrayList<String>();
-            List<ColumnFilter> columnFilters = lookup.getFirstFilterOp().getColumnFilterList();
-            Iterator<ColumnFilter> it= columnFilters.iterator();
+            List<CoreFunctionFilter> columnFilters = extractCoreFunctionFilter(lookup.getFirstFilterOp().getColumnFilterList());
+            Iterator<CoreFunctionFilter> it = columnFilters.iterator();
             while (orgEntities.isEmpty() && it.hasNext()) {
-                CoreFunctionFilter column = (CoreFunctionFilter)it.next();
+                CoreFunctionFilter column = it.next();
 
                 if (column.getColumnId().toUpperCase().equals(columnId)) {
                     potOwners.addAll(column.getParameters());
@@ -77,9 +78,12 @@ public class UserAndGroupsTasksPreprocessor extends UserTasksPreprocessor {
                         addUserAndGroupsFromIdentityProvider(orgEntities, potOwner);
                     }
 
-                    it.remove();
+                    //  we have now a tree as expression. we need to traverse the entire tree for finding the filter.
+                    removeCoreFunctionFilter(lookup.getFirstFilterOp().getColumnFilterList(), column);
                 }
+
             }
+
 
             if (orgEntities.isEmpty()) {
                 addUserAndGroupsFromIdentityProvider(orgEntities, identityProvider.getName());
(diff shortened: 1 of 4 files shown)


elguardian:
Jenkins retest this please.

elguardian:
Jenkins retest this please.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
```diff
@@ -105,4 +109,27 @@ private void addUserAndGroupsFromIdentityProvider(List<Comparable> orgEntities,
         orgEntities.addAll(Optional.ofNullable(userGroupCallback.getGroupsForUser(userId)).orElse(new ArrayList<>()));
         orgEntities.add(userId);
     }
+
+    private void removeCoreFunctionFilter(List<ColumnFilter> filters, ColumnFilter removedFilter) {
+        if (filters.remove(removedFilter)) {
```
gmunozfe: If the list of filters contains other filters not related to the `columnId` (`POTOWNER`), these would be removed here and I think this is not expected.

My suggestion is to control first that the `removedFilter` is related to the `columnId`:

```suggestion
       if ((removedFilter.getColumnId().toUpperCase().equals(columnId)) && filters.remove(removedFilter)) {
```
Notice that current behavior works for filters like this (POTOWNER column first), because rest of filters won't be deleted:
`(POTOWNER in potOwner AND COLUMN2 = value2)`
but not for
`(COLUMN2 = value2 AND POTOWNER in potOwner)`
when first filter would be deleted

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
```diff
@@ -105,4 +109,27 @@ private void addUserAndGroupsFromIdentityProvider(List<Comparable> orgEntities,
...
+
+    }
+    private List<CoreFunctionFilter> extractCoreFunctionFilter(List<ColumnFilter> filters) {
+        List<CoreFunctionFilter> list = new ArrayList<>();
+        for (ColumnFilter filter : filters) {
+            if (filter instanceof CoreFunctionFilter) {
+                list.add((CoreFunctionFilter) filter);
+            } else if (filter instanceof LogicalExprFilter) {
+                list.addAll(extractCoreFunctionFilter(((LogicalExprFilter) filter).getLogicalTerms()));
+            }
```
gmunozfe: What happens for other instances of filters that are not `CoreFunctionFilter` neither `LogicalExprFilter`?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
```diff
@@ -77,8 +78,11 @@ public void preprocess(DataSetLookup lookup) {
                         addUserAndGroupsFromIdentityProvider(orgEntities, potOwner);
                     }
 
-                    it.remove();
                 }
+
+                // we need to look for recursively for the filter
```
mswiderski: this comment is quite misleading to me... looks like it missing what we need to look for?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
```diff
@@ -77,8 +78,11 @@ public void preprocess(DataSetLookup lookup) {
...
+                removeCoreFunctionFilter(lookup.getFirstFilterOp().getColumnFilterList(), column);
```
mswiderski: shouldn't this be placed in exact same location as 'it.remove()' before?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
mswiderski: good point, either this or as mentioned in my other comment why this is taken out from the if ....

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
elguardian: the other won't apply here or it should not be here.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
elguardian: now we have a tree. so we need to traverse recursively for the filter.
changed it

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
elguardian: you are right.
changed it.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessor.java
elguardian: changed. it must be the AND for backward compatibility. Also it looks like it is a way to restrict the query (not just change the column filter) otherwise it would not be added in case there are not filters.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessorTest.java
```diff
@@ -138,6 +140,39 @@ public void testPotOwnerFilterNoIdentityProvider() {
...
+        String potOwner = "potOwner";
+        when(userGroupCallback.getGroupsForUser(potOwner)).thenReturn(Arrays.asList("role1", "role2"));
+
+        List<String> potOwners = new ArrayList<String>();
+        potOwners.add(potOwner);
+
+        QueryParam queryParam = new QueryParam(COL_ID, "IN", potOwners);
+        QueryParam queryParamTerm1 = new QueryParam("MY_COLUMN", "EQUALS_TO", Arrays.asList("1234"));
+        QueryParam queryParamTerm2 = new QueryParam("MY_COLUMN", "NOT_IN", Arrays.asList(1, 2, 3, 4));
+        QueryParam queryParamOR = new QueryParam(null, "OR", Arrays.asList(queryParamTerm1, queryParamTerm2, queryParam));
```
gmunozfe: So, in this test, the whole filter to preprocess is

(MY_COLUMN = 1234 OR MY_COLUMN not_in 1, 2, 3, 4 **OR** POTOWNER in potOwner)

but after preprocessing, the filters are split in a list(because more values have been added for POTOWNER column) but they are going to be rebuilt with AND, no?

((MY_COLUMN = 1234 OR MY_COLUMN not_in 1, 2, 3, 4) **AND** (POTOWNER in role1, role2, potOwner))

Or am I misunderstanding anything?

elguardian: correct. This should the right approach how this filter works.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/impl/query/preprocessor/UserAndGroupsTasksPreprocessorTest.java
```diff
@@ -138,6 +140,39 @@ public void testPotOwnerFilterNoIdentityProvider() {
...
+        ColumnFilter columnFilter = (ColumnFilter) coreFunctionQueryParamBuilder.build();
+
+        DataSetFilter filter = new DataSetFilter();
+        filter.addFilterColumn(columnFilter);
+        dataSetLookup.addOperation(filter);
+
+        userAndGroupsTasksPreprocessor.preprocess(dataSetLookup);
+        List<ColumnFilter> filters = dataSetLookup.getFirstFilterOp().getColumnFilterList();
+
+        assertEquals(filters.size(), 2);
```
gmunozfe: Swap these arguments to be: expected, actual

PR #1596, opened 2020-02-18
Fixed issue when having same dates - deleteLogsByDate method

afalhambra:
deleteLogsByDate method fixed by adding a delay of 1sec when starting up BPM processes in order to avoid issue with equal dates.

Previous fix didn't work completely as there were cases where some of the dates were equals between them and some of them not. For instance, out of 4 dates, 2 were equals between them and the other 2 were equals between them but different than the first two. So it is difficult to manage this scenario. 
One option would've been to use same data structure, either a hashset or a list for both, startDates and resultList but I think the easiest solution is to add a small delay of 1 second when starting up the processes. 
By doing this, we make sure the start dates are completely different one each other.

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ProcessInstanceLogCleanTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ProcessInstanceLogCleanTest.java
@@ -22,6 +22,7 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.assertj.core.api.Assertions;
@@ -90,7 +91,7 @@ public class ProcessInstanceLogCleanTest extends JbpmTestCase {
     }
 
     @Test
-    public void deleteLogsByProcessName() {
+    public void deleteLogsByProcessName() throws InterruptedException {
         KieSession kieSession = createKSession(HELLO_WORLD_PROCESS);
 
         startProcess(kieSession, HELLO_WORLD_PROCESS_ID, 2);
@@ -123,7 +124,7 @@ public class ProcessInstanceLogCleanTest extends JbpmTestCase {
     }
 
     @Test
-    public void deleteLogsByProcessId() {
+    public void deleteLogsByProcessId() throws InterruptedException {
         KieSession kieSession = createKSession(HELLO_WORLD_PROCESS);
 
         startProcess(kieSession, HELLO_WORLD_PROCESS_ID, 3);
@@ -156,7 +157,7 @@ public class ProcessInstanceLogCleanTest extends JbpmTestCase {
     }
 
     @Test
-    public void deleteLogsByVersion() {
+    public void deleteLogsByVersion() throws InterruptedException {
         KieSession kieSession = createKSession(HELLO_WORLD_PROCESS);
         startProcess(kieSession, HELLO_WORLD_PROCESS_ID, 7);
         disposeRuntimeManager();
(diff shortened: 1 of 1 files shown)


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ProcessInstanceLogCleanTest.java
```diff
@@ -223,7 +224,7 @@ public void deleteLogsWithStatusActive() {
 
     @Test
     @BZ("1188702")
-    public void deleteLogsByDate() {
+    public void deleteLogsByDate() throws InterruptedException {
```
MarianMacik: If only this test is problematic, I would add a delay just here, otherwise we will be prolonging all tests in the class.

afalhambra: problem is only happening for that particular test method, however, the problem itself is the call to the method 

`startProcess(kieSession, HELLO_WORLD_PROCESS_ID, 4);`

That's the one which is starting 4 processes at same time - for the code above.

So far, the issue is only happening for that test method, but it might affect future test cases calling the same "startProcess" method.

Another option, would be to create a new method overloaded called "startProcess", passing a delay in miliseconds for instance.

I'd prefer this option - what do you think @MarianMacik?

gmunozfe: yes, having a different method seems more clear.

One second delay is too much, perhaps 5-10 ms is enough as precision is per millisecond.

MarianMacik: +1 @afalhambra

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ProcessInstanceLogCleanTest.java
```diff
@@ -385,9 +380,10 @@ private void abortProcess(KieSession kieSession, long pid) {
         }
     }
 
-    private List<ProcessInstance> startProcess(KieSession kieSession, String processId, int count) {
+    private List<ProcessInstance> startProcess(KieSession kieSession, String processId, int count) throws InterruptedException {
         List<ProcessInstance> piList = new ArrayList<ProcessInstance>();
         for (int i = 0; i < count; i++) {
+            TimeUnit.MILLISECONDS.sleep(1000);
```
gmunozfe: Reduce sleep time (probably a few milliseconds is enough)

afalhambra: agree

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/log/ProcessInstanceLogCleanTest.java
afalhambra: Done. Updated.

afalhambra:
> I had one other option in my mind without the sleep - check if

But problem with that is the following scenario which I was able to reproduce in Jenkins with further logs enabled:

1st Date = 22/02/2020 00:00:01.0001
2nd Date = 22/02/2020 00:00:01.0001
3rd Date = 22/02/2020 00:00:01.0002
4th Date = 22/02/2020 00:00:01.0002

As you can see under this scenario, first two dates are equals, so are last two. In this case, not all 3 will be deleted...

This is issue I noticed on Jenkins.

MarianMacik:
In that case all 4 will be deleted. (you construct the startDates Set froom 2nd, 3rd, 4th which will contain both dates) And you know it by the fact that 1st Date will be in the Set made from 2nd, 3rd and 4th Date. So if it is there, all 4 are deleted. If not, then just 3 of them will be deleted. But maybe I am missing something :)

PR #1599, opened 2020-02-24
6.5.x

Sandeepg20:
Am just looking at the timer stuck issue on jbpm6.5 with perprocess instance default timer


MarianMacik:
Hi @Sandeepg20. `6.5.x` is no longer under development, but please feel free to create a JIRA describing your issue. I unfortunately cannot see your commit in the PR, so I am not sure what the issue really is about.

PR #1603, opened 2020-03-09
[RHPAM-2774] AchievedAtDate for milestone is not preserved when case is reopen

elguardian:
modify query for milestones as a plain value object

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
@@ -340,44 +340,47 @@ public class CaseRuntimeDataServiceImpl implements CaseRuntimeDataService, Deplo
 
     @Override
     public Collection<CaseMilestoneInstance> getCaseInstanceMilestones(String caseId, boolean achievedOnly, QueryContext queryContext) {
-        ProcessInstanceDesc pi = runtimeDataService.getProcessInstanceByCorrelationKey(correlationKeyFactory.newCorrelationKey(caseId));        
+        ProcessInstanceDesc pi = runtimeDataService.getProcessInstanceByCorrelationKey(correlationKeyFactory.newCorrelationKey(caseId));
         if (pi == null || !pi.getState().equals(ProcessInstance.STATE_ACTIVE)) {
             throw new CaseNotFoundException("No case instance found with id " + caseId + " or it's not active anymore");
         }
         CorrelationKey correlationKey = correlationKeyFactory.newCorrelationKey(caseId);
-        
-        Collection<org.jbpm.services.api.model.NodeInstanceDesc> nodes = runtimeDataService.getNodeInstancesByCorrelationKeyNodeType(correlationKey, 
-                                                                                                Arrays.asList(ProcessInstance.STATE_ACTIVE), 
-                                                                                                Arrays.asList("MilestoneNode"), 
-                                                                                                queryContext);
-        
-        Collection<Long> completedNodes = nodes.stream().filter(n -> ((NodeInstanceDesc)n).getType() == 1).map(n -> n.getId()).collect(toList());
-        Predicate<org.jbpm.services.api.model.NodeInstanceDesc> filterNodes = null;
-        if (achievedOnly) {            
-            filterNodes = n -> ((NodeInstanceDesc)n).getType() == 1;             
-        } else {
-            filterNodes = n -> ((NodeInstanceDesc)n).getType() == 0;
-        }
-        List<String> foundMilestones = new ArrayList<>();
-        
-        List<CaseMilestoneInstance> milestones = nodes.stream()
-        .filter(filterNodes)
-        .map(n -> {
-            foundMilestones.add(n.getName());
-            return new CaseMilestoneInstanceImpl(String.valueOf(n.getId()), n.getName(), completedNodes.contains(n.getId()), n.getDataTimeStamp());        
-        })
-        .collect(toList());
-        
+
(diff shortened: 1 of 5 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
```diff
@@ -340,43 +340,36 @@ public CaseDefinition getCase(String deploymentId, String caseDefinitionId) {
...
+                                                                                                                                     queryContext);
+
+        // achieved milestones
+        Map<String, List<CaseMilestoneInstance>> milestonesGroup = nodes.stream()
+                                                                        .map(n -> (NodeInstanceDesc) n)
+                                                                        .filter(n -> n.getType() == 1)
+                                                                        .map(n -> new CaseMilestoneInstanceImpl(String.valueOf(n.getId()), n.getName(), true, n.getDataTimeStamp()))
+                                                                        .collect(Collectors.groupingBy(CaseMilestoneInstance::getName));
+
+        List<CaseMilestoneInstance> milestones = milestonesGroup.values().stream().map(e -> e.stream().min((o1, o2) -> o1.getAchievedAt().compareTo(o2.getAchievedAt())).get()).collect(toList());
```
gmunozfe: The map contains lists with one element, so ordering is applying to that list and therefore not working.
Perhaps you can order after generating the list of CaseMilestoneInstance.

`milestones.sort(Comparator.comparing(CaseMilestoneInstance::getAchievedAt));`

elguardian: Not really sure I understand your comments. If a case is cancelled and then reopen it should contain at least two elements and so on ?... do you mind to check again ?

gmunozfe: I meant for the failing test, which contained only one element per key, and the list obtained from the map was not ordered. Now, it's working fine, thanks!

gmunozfe:
Check this scenario:

No milestone achieved before cancelling case. After reopening, milestone is achieved.
When retrieving milestones, however, _achievedAt_ date corresponds to _processInstanceId=1_, when during that instance, no milestone was achieved (it was achieved for _proccessInstanceId=2_).

It seems like that node is updated to _completed=true_, when it shouldn't:

_NodeInstanceDesc{id=1, nodeId=PID__d78219e8-fdde-4c24-9ed6-8864d7b2d26b, nodeUniqueId=PID__d78219e8-fdde-4c24-9ed6-8864d7b2d26b, name=All work done, deploymentId=org.jbpm.cases:case-module:1.0.0, **processInstanceId=1**, type=MilestoneNode, **completed=true**, dataTimeStamp=2020-03-09 18:04:02.748}]_

elguardian:
you were right @gmunozfe but there was not a way to know the type of node exit. I just added the flat to the node instance log to be able to discriminate between them. This should work.

gmunozfe:
works fine, thanks @elguardian, awesome job!

elguardian:
@gmunozfe I need you check the scripting node instance log new column and feedback

jbpm-installer/src/test/resources/ddl60/postgresql/postgresql-jbpm-schema.sql
```diff
@@ -130,6 +130,7 @@
         processInstanceId int8 not null,
         type int4 not null,
         workItemId int8,
+        skipable boolean,
```
gmunozfe: shouldn't it be "cancel"?

jbpm-installer/src/test/resources/ddl60/db2/db2-jbpm-schema.sql
```diff
@@ -130,6 +130,7 @@
         processInstanceId bigint not null,
         type integer not null,
         workItemId bigint,
+        cancel smallint not null,
```
gmunozfe: do we need to update ddl60 schemas? my understanding is that this is used to test migration from 6.0 to 7.x

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/CaseRuntimeDataServiceImpl.java
```diff
@@ -340,44 +340,47 @@ public CaseDefinition getCase(String deploymentId, String caseDefinitionId) {
...
-        
-        return applyPagination(milestones, queryContext);
+
+        List<CaseMilestoneInstance> sortedMilestones = milestones.stream().sorted(this::compareMilestones).collect(toList());
+        return applyPagination(sortedMilestones, queryContext);
+    }
+
+    private int compareMilestones(CaseMilestoneInstance o1, CaseMilestoneInstance o2) {
+        if (o1.getAchievedAt() == null) {
+            return -1;
```
gmunozfe: if both are null, shouldn't it return 0?

elguardian: null <> null = true... null == null = false... depends on our view.

jbpm-installer/src/test/resources/ddl60/db2/db2-jbpm-schema.sql
elguardian: this flag will be changed in exchange for new type of nodeinstancelog

elguardian:
depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/431

PR #1606, opened 2020-03-09
[RHPAM-2782] Error boundary event not working as expected with reusable subprocess node.

elguardian:
error boundary compensation event is not thrown.


elguardian:
you were right @gmunozfe but there was not a way to know the type of node exit. I just added the flat to the node instance log to be able to discriminate between them. This should work.

PR #1607, opened 2020-03-11
[RHPAM-2789] CorrelationKeyInfo table is used even without using correlation key functionality

elguardian:
remove select for CorrelationKeyInfo and adding unique index

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
@@ -2148,27 +2148,30 @@ public class ActivityTest extends JbpmBpmn2TestCase {
     @RequirePersistence
     @Test
     public void testCallActivityChain() throws Exception {
+
         try {
-            System.setProperty("org.jbpm.correlationkey.length", "70");
+            System.setProperty("org.jbpm.correlationkey.length", "90");
             KieBase kbase = createKnowledgeBase("correlationkey/Process1.bpmn2",
-                    "correlationkey/Process2.bpmn2",
-                    "correlationkey/Process3.bpmn2",
-                    "correlationkey/Process4.bpmn2",
-                    "correlationkey/Process5.bpmn2");
+                                                "correlationkey/Process2.bpmn2",
+                                                "correlationkey/Process3.bpmn2",
+                                                "correlationkey/Process4.bpmn2",
+                                                "correlationkey/Process5.bpmn2");
             ksession = createKnowledgeSession(kbase);
-            Map<String, Object> params = new HashMap<String, Object>();        
+            Map<String, Object> params = new HashMap<String, Object>();
             ProcessInstance processInstance = ksession.startProcess(
-                    "src.Process1", params);
+                                                                    "src.Process1", params);
             assertProcessInstanceCompleted(processInstance);
-            
+
             ProcessInstanceLog log = logService.findProcessInstances("src.Process5").get(0);
             assertNotNull(log);
             assertNotNull(log.getCorrelationKey());
             assertTrue(log.getCorrelationKey().startsWith(processInstance.getId() + ":src.Process2:"));
             assertTrue(log.getCorrelationKey().contains(":src.Process4"));
+            assertTrue(log.getCorrelationKey().length() <= 90);
         } finally {
             System.clearProperty("org.jbpm.correlationkey.length");
         }
+
     }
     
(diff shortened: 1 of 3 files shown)


elguardian:
This is ok to test.

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
```diff
@@ -2148,27 +2148,24 @@ public void testBusinessRuleTaskException() throws Exception {
...
-            assertProcessInstanceCompleted(processInstance);
-            
-            ProcessInstanceLog log = logService.findProcessInstances("src.Process5").get(0);
-            assertNotNull(log);
-            assertNotNull(log.getCorrelationKey());
-            assertTrue(log.getCorrelationKey().startsWith(processInstance.getId() + ":src.Process2:"));
-            assertTrue(log.getCorrelationKey().contains(":src.Process4"));
-        } finally {
-            System.clearProperty("org.jbpm.correlationkey.length");
-        }
```
MarianMacik: Would be probably good to correct (and retain) such a test.

elguardian: it is a faulty test in my opinion due to the way of constructing the correlation key is creating keys bigger than 70.

MarianMacik: Well, that is the point of the test - or should be - to check if it is trimmed in case a system property is defined.

elguardian: it is a faulty test.

persist correlation info: null
Process1
persist correlation info: 1:src.Process2:1584350223097
Process2
persist correlation info: 1:src.Process2:1584350223097:src.Process3:1584350223109
Process3
2020-03-16 10:17:03,119 [main] WARN  CorrelationKey content was trimmed as it was too long (more than 70 characters)
persist correlation info: 1:src.Process2:1584350223097:src.Process3:1584350223109:src.Process4:1
2020-03-16 10:17:03,124 [main] WARN  CorrelationKey content was trimmed as it was too long (more than 70 characters)
Process4
2020-03-16 10:17:03,130 [main] WARN  CorrelationKey content was trimmed as it was too long (more than 70 characters)
2020-03-16 10:17:03,130 [main] WARN  CorrelationKey content was trimmed as it was too long (more than 70 characters)
persist correlation info: 1:src.Process2:1584350223097:src.Process3:1584350223109:src.Process4:1
2020-03-16 10:17:03,132 [main] WARN  SQL Error: 23505, SQLState: 23505

The trim is correct, the problem is that process 4 (already excceds the 70 positions) and process 5 generates the same as process 4 correlation key that is the reason it is  failing now when you haven an index in database.

MarianMacik: than I guess we can limit it to for example to 100 and it should be fine, wdyt?

elguardian: yeah I fixed the test. just to give you the bits. The test was working because another bug in 

https://github.com/kiegroup/jbpm/blob/master/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java#L118

this uses this function:
https://github.com/kiegroup/jbpm/blob/master/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java#L142

if you see it uses the externalForm of the correlation Key... the problem is that the external form is not being trimmed like the name
https://github.com/kiegroup/jbpm/blob/d7a0195ea2f8cb99930d1c875fa28d24e1c6be8e/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/correlation/CorrelationKeyInfo.java#L153

so it can happen that while the name is trimmed you are looking for a key not trimmed not finding the right the right one. That is the reason is a faulty test. 

I did follow you recomendation and added the check

MarianMacik: I see... Well, the original issue was that only process instance log were affected for the reporter of the issue, so in fact CorrelationKeyInfo class change wasn't even needed in the end and was done just for consistency (which in fact wasn't consistent as you found out :) ) Good catch!

PR #1609, opened 2020-03-17
[RHPAM-2805] memory growth when starting a high number of process instances with a high timers

elguardian:
removing transient objects when the timer is created.


elguardian:
@gmunozfe in this case is not necessary as there is no file descriptors involved.

PR #1612, opened 2020-03-19
[7.7.x] [RHPAM-2782] Error boundary event not working as expected with reusable subprocess node.

elguardian:
error boundary compensation event is not thrown.


elguardian:
cherry picked from https://github.com/kiegroup/jbpm/commit/50a61c9018edab159a7d8f8ae1bdf658270ed5cb

PR #1617, opened 2020-03-24
JBPM-9072 single test

hasys:
Hi @elguardian, @MarianMacik,

I migrated one unit test from https://github.com/kiegroup/kogito-runtimes/pull/314/files#diff-15fa3183e6865094973de2150bbc3b78R1862 to be sure that it is not implemented in jbpm (non-Kogito). This PR can be used as base for future test migration / implementation.

Thank you!

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/ActivityTest.java
@@ -2226,4 +2226,30 @@ public class ActivityTest extends JbpmBpmn2TestCase {
         assertEquals("new value", listOut.get(0));
         assertEquals("new value", listOut.get(1));
     }
+
+    @Test
+    public void testUserTaskWithExpressionsForIO() throws Exception {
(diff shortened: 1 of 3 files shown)


elguardian:
@hasys I am bit lost with this one... what do you intent ? you said " to be sure that it is not implemented in jbpm" so this feature should not be in jbpm right ? then why this PR ¿?

hasys:
Hi @elguardian,

sorry for confusion. I mean this feature is implemented in Kogito runtime, and I expected it to work in Business Central as well. So I started to test it in Business Central but it didn't work. But I wasn't sure is it because of issues in our BPMN Designer or in the engine, so I ported this test from Kogito to check that issue is in jbpm engine.

Since test if failing right in the place where this feature is tested I assume that test is correct and feature is not implemented. So you can use this PR as starting point to port all other tests from Kogito and than implement it in jbpm engine.

Hope it makes more sense, thank you @elguardian!

gmunozfe:
@hasys probably it could be a "draft pull request" till implementation is completed

hasys:
@gmunozfe I am not an assignee, I just shared my findings. Feel free to close this PR if it is not needed. Next time will open a draft instead, good point.

Thank you!

PR #1620, opened 2020-03-27
[RHPAM-2839] Event Sub-process with timer start node fire if an intermediate timer from process is triggered

elguardian:
discrimate wheter the timer is triggered for subprocess or not

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventSubProcessNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventSubProcessNodeInstance.java
@@ -19,6 +19,7 @@ import java.util.List;
 import java.util.stream.Collectors;
 
 import org.jbpm.process.instance.ProcessInstance;
+import org.jbpm.process.instance.timer.TimerInstance;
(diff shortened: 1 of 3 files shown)


jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventSubProcessNodeInstance.java
```diff
@@ -54,14 +56,22 @@ public void signalEvent(String type, Object event) {
...
         }
         super.signalEvent(type, event);
     }
+
+    private boolean containsTimerEvent(String type, Object event) {
+        if (!type.equals("timerTriggered") || !(event instanceof TimerInstance)) {
+            return false;
+        }
+        TimerInstance timerInstance = (TimerInstance) event;
+        return getTimerInstances().contains(timerInstance.getId());
```
gmunozfe: May `getTimerInstances` return null and this line throws a NPE?

elguardian: u r right. good catch

PR #690, opened 2016-11-25
JBPM-5423 - remove singleResult flag from QueryFilter

mswiderski:

PR #1621, opened 2020-03-30
[JBPM-9085],[JBPM-9086] - fixed failing tests

mareknovotny:
https://issues.redhat.com/browse/JBPM-9085

just in case adding check for null value

### DIFF
--- a/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/HumanTaskQueryFilterTest.java
+++ b/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/HumanTaskQueryFilterTest.java
@@ -123,19 +123,6 @@ public class HumanTaskQueryFilterTest extends JbpmTestCase {
         abortHumanTaskProcess(3);
     }
 
-    @Test
-    @BZ("1132145")
-    public void testSingleResult() {
-        startHumanTaskProcess(4, "john's task", "john");
-
-        List<TaskSummary> taskList = taskService.getTasksAssignedAsPotentialOwner("john", null, null,
-                new QueryFilter(0, 0));
-        logger.debug("### Potential owner task list: " + taskList);
-        Assertions.assertThat(taskList).hasSize(1);
-
-        abortHumanTaskProcess(4);
-    }
-
     /**
      * TODO - Need to pass in parameters which do make sense.
      */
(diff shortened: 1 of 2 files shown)


MarianMacik:
jenkins retest this

gmunozfe:
@mareknovotny  these failing tests weren't being executed because of BZ annotation. Are they needed now?

If yes, _testFilterParams_ test may be modified like this (line [147](https://github.com/kiegroup/jbpm/blob/master/jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/HumanTaskQueryFilterTest.java#L147)):
`QueryFilter queryFilter = new QueryFilter("t.id=1 AND t.priority=0", null, "t.name", true);`

The other one, _testSingleResult_ I think it's no longer supported, and could be deleted: https://bugzilla.redhat.com/show_bug.cgi?id=1132145
There's no "singleResult" param at QueryFilter

MarianMacik:
Ah, so that's why it suddenly started failing :) I would then recommend this:

testSingleResult
* Feature was removed in https://github.com/kiegroup/jbpm/pull/690, so I believe we should delete also the test.

testFilterParams
* I believe that using both param names and param values in the first argument is abusing of the API, therefore we should change it to 

`QueryFilter queryFilter = new QueryFilter("t.id=:id AND t.priority=:priority", {Map with "id"->1, "priority"->0}, "t.name", true);`

What do you think?

gmunozfe:
@MarianMacik I agree, filterParams are conditions with named parameters and map contains their values, so probably it has no sense to pass a null map, and therefore, null pointer exception is useful in this case (following Mario's tweet "There's no line of code deserving more thinking and explanations than a null check." ;-)

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/task/HumanTaskQueryFilterTest.java
```diff
@@ -123,28 +123,17 @@ public void testAscendingOrder() {
...
      * TODO - Need to pass in parameters which do make sense.
      */
     @Test
     @BZ("1132157")
     public void testFilterParams() {
         startHumanTaskProcess(10, "john's task", "john");
-
-        QueryFilter queryFilter = new QueryFilter("x=1,y=2", null, "t.name", true);
+        Map<String, Object> parameters = new HashMap<String, Object>();
+        parameters.put("id", 1);
```
gmunozfe: It should be "1L"

mareknovotny:
@MarianMacik I would keep that null check there as that not replaced parameters will be even more clearer than just NPE thrown there. I am always fan of checks if it is not directly clear on first view or IAE should be thrown there instead.

mareknovotny:
just in case you haven't noticed this failures start to shown over weekend or so. one of it is seanable at 
kiegroup/droolsjbpm-build-bootstrap/pull/12

mareknovotny:
skip ci

Ginxo:
tested locally and they are working. LGTM

PR #1623, opened 2020-04-01
[BAPL-1533] Adding case id expressions for generating case Id

elguardian:


elguardian:
Jenkins this is ok to test.

PR #1628, opened 2020-04-03
[RHPAM-2813] Allow to use 'user' query parameter with /contents/output REST API.

elguardian:
kie api service in jbpm to store task event with another user.

### DIFF
--- a/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/UserTaskService.java
+++ b/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/UserTaskService.java
@@ -496,6 +496,15 @@ public interface UserTaskService {
     
     /**
      * Saves gives values as content of a task, applies to task output as input cannot be altered
+     * @param taskId
+     * @param values
+     * @return
+     * @throws TaskNotFoundException in case task with given id was not found
+     */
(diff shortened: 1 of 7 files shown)


elguardian:
depends on https://github.com/kiegroup/droolsjbpm-knowledge/pull/435

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/UserTaskService.java
```diff
@@ -494,6 +494,15 @@
      */
     Long saveContent(Long taskId, Map<String, Object> values);
     
+    /**
+     * Saves gives values as content of a task, applies to task output as input cannot be altered
```
afalhambra, line 498: This method description might be reworded, as to differ saveContenFromUser from the saveContent method.

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/UserTaskService.java
```diff
@@ -494,6 +494,15 @@
...
+     * @param taskId
+     * @param values
```
afalhambra, line 500: Please, add the missing parameter "userId" in the doc.

```suggestion
     * @param taskId
     * @param userId
     * @param values
```

elguardian:
@afalhambra can you review ?

PR #1629, opened 2020-04-03
[RHPAM-2839] Event Sub-process with timer start node fire if an intermediate timer from process is triggered

elguardian:
discrimate wheter the timer is triggered for subprocess or not


elguardian:
cherry picked from 
https://github.com/kiegroup/jbpm/commit/d5772933b9bb1478efef310423f7a078bf98a5c9

PR #1630, opened 2020-04-06
BXMSPROD-702 jbpm-installer excluded from productized profile

Ginxo:
https://issues.redhat.com/browse/BXMSPROD-702
My only concern is about the jbpm-installer sources remain in the jbpm-distribution assembly zip `jbpm-%version%-src-zip` file

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,7 +107,6 @@
         <module>jbpm-case-mgmt</module>
         <module>jbpm-test-coverage</module>
         <module>jbpm-examples</module>
-        <module>jbpm-installer</module>
         <module>jbpm-container-test</module>
         <module>jbpm-event-emitters</module>
         <module>jbpm-xes</module>
@@ -278,6 +277,18 @@
                     </plugins>
                 </pluginManagement>
             </build>
+        </profile>
+        <profile>
+            <id>jbpm-installer-exclude-productized</id>
+            <activation>
+                <property>
+                    <name>productized</name>
+                    <value>!true</value>
+                </property>
+            </activation>
+            <modules>
+                <module>jbpm-installer</module>
+            </modules>
         </profile>
     </profiles>
 
--- a/jbpm-distribution/pom.xml
+++ b/jbpm-distribution/pom.xml
@@ -72,31 +72,6 @@
           </execution>
         </executions>
       </plugin>
-      <plugin>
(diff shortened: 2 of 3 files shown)


mareknovotny:
I tried that it seems works only with property activation,if you use `-P productized` it is still included in build reactor

mareknovotny:
> https://issues.redhat.com/browse/BXMSPROD-702
> My only concern is about the jbpm-installer sources remain in the jbpm-distribution assembly zip `jbpm-%version%-src-zip` file
we need to exclude it there too

pom.xml
```diff
@@ -279,6 +278,29 @@
...
+                <property>
+                    <name>productized</name>
+                    <value>false</value>
+                </property>
+            </activation>
+            <modules>
+                <module>jbpm-installer</module>
+            </modules>
+        </profile>
     </profiles>
```
mareknovotny, line 304: I think both cases/conditions can be covered by simple 
`<value>!true</value>`

Ginxo: let me try it

Ginxo: @mareknovotny much simpler with
```
<name>productized</name>
<value>!true</value>
```
thanks a lot

Ginxo:
@mareknovotny also excluded from src distribution zip file

pom.xml
```diff
@@ -279,6 +278,18 @@
                 </pluginManagement>
             </build>
         </profile>
+        <profile>
+            <id>jbpm-installer-exclude-productized</id>
```
afalhambra, line 282: Why not using same id profile as per jbpm-distribution/pom.xml in line 404 
```suggestion
            <id>jbpm-installer-include</id>
```

afalhambra, line 282: Why not using same id profile as per jbpm-distribution/pom.xml in line 404 
```suggestion
            <id>jbpm-installer-include</id>
```

mareknovotny:
There could be another idea how to do that in more simpler way https://github.com/kiegroup/jbpm/pull/1631, I am not sure that we need full profile there, but kept it for not changing the usages
Also jbpm-distribution could be needed somewhere in assemblies, but we can look if that is safe

Ginxo:
replaced by https://github.com/kiegroup/jbpm/pull/1631

PR #1634, opened 2020-04-16
[JBPM-9118] Support disabling of Notification Listener

elguardian:
add flags for include/excluding notification listeners

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
@@ -16,8 +16,11 @@
 package org.jbpm.services.task.deadlines.notifications.impl;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 import java.util.ServiceLoader;
+import java.util.function.Predicate;
 
 import org.jbpm.services.task.deadlines.NotificationListener;
 import org.kie.internal.task.api.UserInfo;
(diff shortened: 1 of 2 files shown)


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
```diff
@@ -30,18 +32,59 @@
...
-    
+
+    private List<String> excludeLists;
+
+    private List<String> includeLists;
+
+    public static List<String> propertyToList(String propertyName) {
+        String name = System.getProperty(propertyName);
+        if (name == null) {
+            return null;
```
gmunozfe, line 55: Why not return `Collections.emptyList()` ?
Then, at createPredicate you can check against "is not empty"...

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
```diff
@@ -30,18 +32,59 @@
...
         for (NotificationListener listener : listenersLoaded) {
-            listeners.add(listener);
+            if (predicate.test(listener.getClass().getName())) {
+                listeners.add(listener);
+            }
+        }
+    }
+
+    private Predicate<String> createPredicate() {
+        if(includeLists != null) {
```
gmunozfe, line 82: Missing a blank after "if"
If `propertyToList` returns empty list, check `if (!includeLists.isEmpty())`

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
```diff
@@ -30,18 +32,59 @@
...
+            return e -> includeLists.contains(e);
+        } else if (excludeLists != null) {
```
gmunozfe, line 84: Same here

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
```diff
@@ -30,18 +32,59 @@
...
+            return null;
+        }
+        return Arrays.asList(name.split(","));
+    }
+
     private NotificationListenerManager() {
+        init();
+    }
+
+    public void reset() {
```
gmunozfe, line 64: Is this method really needed?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
```diff
@@ -30,18 +32,59 @@
...
+        init();
```
gmunozfe, line 65: Is this method really needed? It's similar to make init public

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManagerTest.java
```diff
@@ -0,0 +1,52 @@
+package org.jbpm.services.task.deadlines.notifications.impl;
```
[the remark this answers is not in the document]
elguardian: done

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
elguardian: not equivalent... empty string is not equivalent to null. Think aboud include = ""... this will exclude everthing as the predicate will never test true.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
elguardian: ok with the if. The other one I already explain why not.

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
elguardian: not the same thing

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManager.java
elguardian: yep... this is extracting a method without changing the constructor and reusing the code for reset which helpfull when testing.

elguardian:
@gmunozfe done.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/deadlines/notifications/impl/NotificationListenerManagerTest.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    @Test
+    public void testBothLists() {
+        System.setProperty(NotificationListenerManager.KIE_LISTENER_EXCLUDE, EMAIL_LISTENER);
+        System.setProperty(NotificationListenerManager.KIE_LISTENER_INCLUDE, EMAIL_LISTENER);
+        NotificationListenerManager.get().reset();
+        assertThat(NotificationListenerManager.get().getNotificationListeners().size(), is(1));
+        assertThat(NotificationListenerManager.get().getNotificationListeners().get(0).getClass().getName(), is(EMAIL_LISTENER));
+    }
```
MarianMacik, line 67: What about adding a test which will test excluding/including more than one?

elguardian: done

PR #1635, opened 2020-04-16
[RHPAM-2861] JPA vs JMS audit produces different logs

elguardian:


jbpm-audit/src/test/java/org/jbpm/process/audit/jms/AsyncAuditLogProducerTest.java
```diff
@@ -233,12 +235,19 @@ public void testAsyncAuditLoggerComplete() throws Exception {
         List<ProcessInstanceLog> processInstances = logService.findProcessInstances("com.sample.ruleflow");
         Assertions.assertThat(processInstances.size()).isEqualTo(1);
         List<NodeInstanceLog> nodeInstances = logService.findNodeInstances(processInstance.getId());
+        Collections.sort(nodeInstances, (o1, o2) -> o1.getDate().compareTo(o2.getDate()));
```
gmunozfe, line 238: May getId be better to use in comparison than getDate ?  Anyway, is it needed to sort the collection?

jbpm-audit/src/main/java/org/jbpm/process/audit/jms/AsyncAuditLogProducer.java
```diff
@@ -161,10 +159,11 @@ public void afterNodeTriggered(ProcessNodeTriggeredEvent event) {
     	}
     }
 
+
     @Override
     public void beforeNodeLeft(ProcessNodeLeftEvent event) {
-
-        
+        NodeInstanceLog log = (NodeInstanceLog) builder.buildEvent(event, null);
+        sendMessage(log, AFTER_NODE_LEFT_EVENT_TYPE, 1);
```
gmunozfe, line 166: It should be BEFORE_NODE_LEFT_EVENT_TYPE, shouldn't it?
So, now, it's logging it before leaving the node

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
```diff
@@ -176,15 +175,7 @@ public final void trigger(NodeInstance from, String type) {
     	}
     	
     	if (from != null) {
-    	    int level = ((org.jbpm.workflow.instance.NodeInstance)from).getLevel();
-    	    ((org.jbpm.workflow.instance.NodeInstanceContainer)getNodeInstanceContainer()).setCurrentLevel(level);
```
gmunozfe, line 180: Setting the level is not longer needed?

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
```diff
@@ -176,15 +175,7 @@ public final void trigger(NodeInstance from, String type) {
...
-	    	Collection<Connection> incoming = getNode().getIncomingConnections(type);
-	    	for (Connection conn : incoming) {
-	    	    if (conn.getFrom().getId() == from.getNodeId()) {
-	    	        this.metaData.put("IncomingConnection", conn.getMetaData().get("UniqueId"));
-	    	        break;
-	    	    }
-	    	}
+            this.metaData.put("IncomingConnection", from.getNode().getMetaData().get("UniqueId"));
```
gmunozfe, line 178: "UniqueId" could be defined as constant (as it is used three times in this class) (SonarCloud code smell)

jbpm-audit/src/test/java/org/jbpm/process/audit/jms/AsyncAuditLogProducerTest.java
elguardian: that line was  not needed (removed)

jbpm-audit/src/main/java/org/jbpm/process/audit/jms/AsyncAuditLogProducer.java
elguardian: this can cause an incompatibility. So I wouldn't do that at least for now.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
elguardian: nope. this is really complex to explain, but this piece of code creates new NodeInstanceImpl under the good so it is not good either.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/NodeInstanceImpl.java
elguardian: ok

elguardian:
done @gmunozfe

PR #1637, opened 2020-04-21
[RHPAM-2903] 255 characters limitation in ExecutionErrorInfo database table

elguardian:
trim string in execution error info so it does truncate when info does not fit the column

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/jpa/ExecutionErrorInfo.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/jpa/ExecutionErrorInfo.java
@@ -40,6 +40,8 @@ public class ExecutionErrorInfo extends ExecutionError implements Serializable {
 
 	private static final long serialVersionUID = 6669858787722894023L;
 	
+    private final int ERROR_LOG_LENGTH = Integer.parseInt(System.getProperty("org.kie.jbpm.error.log.length", "255"));
+
 	private Long id;
 
     public ExecutionErrorInfo() {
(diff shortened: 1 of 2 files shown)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/jpa/ExecutionErrorInfo.java
```diff
@@ -103,6 +105,15 @@ public Long getActivityId() {
         return super.getActivityId();
     }
 
+    @Override
+    public void setErrorMessage(String errorMessage) {
+        String trimmedErrorMessage = errorMessage;
+        if (trimmedErrorMessage != null && trimmedErrorMessage.length() > ERROR_LOG_LENGTH) {
+            trimmedErrorMessage = trimmedErrorMessage.substring(0, ERROR_LOG_LENGTH);
+        }
+        super.setErrorMessage(trimmedErrorMessage);
```
gmunozfe, line 114: Perhaps we may use _Apache Commons Lang_ **abbreviate** method, adding 3 dots at the end, to indicate that it's trimmed. Notice that this method would throw _IllegalArgumentException_ if length is equal or less than 3 (should not be a common length to use). What do you think?

```suggestion
        super.setErrorMessage(StringUtils.abbreviate(errorMessage, ERROR_LOG_LENGTH));
```

elguardian: Assuming the exception is not equivalent. So In spite is good catch I would stick to this implementation rather than throw a potential runtime exception. (we would trade off one exception for another even if it is unlikely to happen)

elguardian:
Jenkins retest this please

PR #1638, opened 2020-04-24
[RHPAM-2805] memory growth when starting a high number of process instances with a high timers

elguardian:
removing transient objects when the timer is created.

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -16,6 +16,11 @@
 
 package org.jbpm.services.ejb.timer;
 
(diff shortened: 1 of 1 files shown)


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -204,6 +210,22 @@ public void internalSchedule(TimerJobInstance timerJobInstance) {
 	}
 
 
+    private Serializable removeTransientFields(Serializable info) {
+        // removing transient fields
```
MarianMacik, line 214: Wouldn't it be easier to just set these transient fields to null?

elguardian: It is fairly more complex do that. You cannot guarantee how deep is the  tree and it is error prone

PR #1642, opened 2020-05-04
[RHPAM-2861] Stabilize mysql JMS test ordering also by nodeInstanceId

gmunozfe:
At **mysql/mariadb** execution of _AsyncAuditLogProducerTest#testAsyncAuditLoggerComplete_, the nodes returned by **findNodeInstances** are
 `{Start, Task, End, Start, Task, End}`
instead of expected
`{Start, Start, Task, Task, End, End}`

:grey_exclamation: Notice that _log_date_ is the same when running tests, as execution is very fast.

:point_right: Sorted in the test by **nodeInstanceId** and id to fix it.


gmunozfe:
thanks @elguardian! reordered only in the test

PR #1646, opened 2020-05-11
[JBPM-9147] getTaskById does not return formName

gmunozfe:
Current getTaskInstanceById query didn't fetch formName column (belonging to Task table).
Convert to named-native-query for supporting the join operation and added constructor with this param.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
+++ b/jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
@@ -698,86 +698,88 @@
     <!-- hint name="org.hibernate.timeout" value="200"/ -->
   </named-query>
   
-  <named-query name="getTaskInstanceById">
+  <named-native-query name="getTaskInstanceById">
     <query>
         select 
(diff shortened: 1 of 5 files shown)


gmunozfe:
jenkins retest this

jbpm-services/jbpm-kie-services/src/main/resources/META-INF/Servicesorm.xml
```diff
@@ -698,86 +698,88 @@
     <!-- hint name="org.hibernate.timeout" value="200"/ -->
   </named-query>
   
-  <named-query name="getTaskInstanceById">
+  <named-native-query name="getTaskInstanceById">
     <query>
         select 
-	      new org.jbpm.kie.services.impl.model.UserTaskInstanceDesc(
```
elguardian, line 704: why this you remove this ? It is better to retrieve the custom object UserTaskInstance (type safe)  that an array of objects. The transformation should be the other way round.

gmunozfe:
Closing this, because the JPA 2.1 with named-queries is the right approach (will open a different PR), thanks a lot @elguardian !!

PR #1647, opened 2020-05-12
[JBPM-9147] getTaskById does not return formName

gmunozfe:
Include _formName_ into the UserTaskInstanceDesc object returned by getTaskById, getTaskByWorkItemId and getProcessInstanceById.


gmunozfe:
jenkins retest this

PR #1650, opened 2020-05-14
[JBPM-9155] GROUPS is a reserved word since mysql 8.0.2

gmunozfe:
Table name uses a reserved word for mysql.
If quoted with backticks (`), then other dbs (mssql, db2) don't work (incorrect syntax).


afalhambra:
jenkins retest this please

afalhambra:
Jenkins retest this please

gmunozfe:
@elguardian can you also check and merge if you agree?

gmunozfe:
> have you tried to use invert comas ?... right now seems to me like you are generating an backward incompatibilty. I would explore other options before changing the name of the table.

I've just tried: for mysql 8 worked, but not for mssql, db2, etc.:
_com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near '`'._

afalhambra:
> > have you tried to use invert comas ?... right now seems to me like you are generating an backward incompatibilty. I would explore other options before changing the name of the table.
> 
> I've just tried: for mysql 8 worked, but not for mssql, db2, etc.:
> _com.microsoft.sqlserver.jdbc.SQLServerException: Incorrect syntax near '`'._

@elguardian - why backward compatibility? I don't see any issues regarding that. `Groups` table is created and drop in the test itself and there is no reference to any jbpm persistence entity.

PR #1655, opened 2020-05-20
JBPM-9126 - Missing "subject" attribute in the UserTaskInstanceDesc interface

fjtirado:
Unfortunately, although I have imported all code styles, I cannot avoid empty spaces differences


tsurdilo:
Please add the title of the jira in your pr title as well, for example

JBPM-1234 - This is the title of the Jira

elguardian:
rebase plz. don't remove constructors. Add summary to the commit. 
Also explain what it does.
[jira key] - summary

fjtirado:
> 
> 
> rebase plz. don't remove constructors. Add summary to the commit.
> Also explain what it does.
> [jira key] - summary

This class in internal (it is located in impl package), so it is theoretically allowed to change existing constructor. 
In this particular case, after analyzing the usage of the existing constructor, in order to avoid a constructor "explosion" in a class which already have plenty of them, it was preferred to add the subject parameter to the proper constructor and change the classes using it: UserTaskInstanceQueryMapper and the xml file building instances after query, ServiceOrm.xml

PR #1656, opened 2020-05-20
JBPM-9138 - RESTWorkItemHandler doesn't support proxy

fjtirado:
Creating new constructor with additional useSystemProperties parameter,
which allow creating connection using system properties.

Depends on [droolsjbpm-build-bootstrap#1355](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1355)

### DIFF
--- a/jbpm-workitems/jbpm-workitems-rest/pom.xml
+++ b/jbpm-workitems/jbpm-workitems-rest/pom.xml
@@ -19,7 +19,7 @@
   <properties>
     <java.module.name>org.jbpm.workitem.rest</java.module.name>
   </properties>
-
+  
   <dependencies>
     <dependency>
       <groupId>org.apache.cxf</groupId>
@@ -110,6 +110,11 @@
     <dependency>
       <groupId>javax.activation</groupId>
       <artifactId>activation</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.mock-server</groupId>
+      <artifactId>mockserver-netty</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
+++ b/jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
@@ -29,6 +29,7 @@ import java.util.function.BiConsumer;
 
 import javax.xml.bind.JAXBContext;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.apache.http.Consts;
 import org.apache.http.HttpEntity;
 import org.apache.http.HttpHeaders;
@@ -83,8 +84,8 @@ import org.kie.api.runtime.process.WorkItemManager;
 import org.kie.internal.runtime.Cacheable;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-
-import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.http.impl.conn.DefaultProxyRoutePlanner;
+import org.apache.http.conn.routing.HttpRoutePlanner;
 
 /**
  * WorkItemHandler that is capable of interacting with REST service. Supports both types of services
@@ -156,6 +157,7 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler imple
     private static final Logger logger = LoggerFactory.getLogger(RESTWorkItemHandler.class);
     private static final int DEFAULT_TOTAL_POOL_CONNECTIONS = 500;
     private static final int DEFAULT_MAX_POOL_CONNECTIONS_PER_ROUTE = 50;
+    protected static final String USE_SYSTEM_PROPERTIES = "org.kie.workitem.rest.useSystemProperties";
 
     public static final String PARAM_AUTH_TYPE = "AuthType";
     public static final String PARAM_CONNECT_TIMEOUT = "ConnectTimeout";
@@ -176,10 +178,10 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler imple
     private String password;
     private AuthenticationType type;
     private String authUrl;
-    // default caching to false
-    private boolean doCacheClient = false;
-
+    private boolean doCacheClient;
+    
     private ClassLoader classLoader;
+
 
     protected static PoolingHttpClientConnectionManager connectionManager;
     protected static CloseableHttpClient cachedClient;
@@ -205,20 +207,20 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler imple
      * Used when no authentication is required
      */
     public RESTWorkItemHandler() {
-        logger.debug("REST work item handler will use http client 4.3 api " + HTTP_CLIENT_API_43);
-        this.type = AuthenticationType.NONE;
-        this.classLoader = this.getClass().getClassLoader();
+        this(false);
+        
     }
 
-    /**
+     /**
      * Used when no authentication is required
-     * @param doCacheClient - - true/false setting for client cache
+     * @param doCacheClient true/false setting for client cache
+     * 
      */
     public RESTWorkItemHandler(boolean doCacheClient) {
-        this();
-        this.doCacheClient = doCacheClient;
+        this(doCacheClient,DEFAULT_TOTAL_POOL_CONNECTIONS, DEFAULT_MAX_POOL_CONNECTIONS_PER_ROUTE);
     }
 
+    
     /**
      * Used when no authentication is required
      * @param doCacheClient - - true/false setting for client cache
@@ -228,8 +230,10 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler imple
     public RESTWorkItemHandler(boolean doCacheClient,
                                int totalPoolConnections,
                                int maxPoolConnectionsPerRoute) {
-        this();
-        this.doCacheClient = doCacheClient;
+        logProtocol();
+        this.type = AuthenticationType.NONE;
+        this.classLoader = this.getClass().getClassLoader();
+        this.doCacheClient = doCacheClient;        
         if (doCacheClient) {
             connectionManager.setMaxTotal(totalPoolConnections);
             connectionManager.setDefaultMaxPerRoute(maxPoolConnectionsPerRoute);
@@ -237,10 +241,10 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler imple
     }
 
     /**
-     * Dedicated constructor when BASIC authentication method shall be used
-     * @param username - user name to be used for authentication
-     * @param password - password to be used for authentication
-     */
+    * Dedicated constructor when BASIC authentication method shall be used
+    * @param username - user name to be used for authentication
+    * @param password - password to be used for authentication
+    */
     public RESTWorkItemHandler(String username,
                                String password) {
         this(username,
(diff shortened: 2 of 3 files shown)


tsurdilo:
Please add the title of the jira in your pr title as well, for example

JBPM-1234 - This is the title of the Jira

elguardian:
Add summary to the commit. 
Also explain what it does.
[jira key] - summary

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -156,6 +155,7 @@
     private static final Logger logger = LoggerFactory.getLogger(RESTWorkItemHandler.class);
     private static final int DEFAULT_TOTAL_POOL_CONNECTIONS = 500;
     private static final int DEFAULT_MAX_POOL_CONNECTIONS_PER_ROUTE = 50;
+    private static final boolean DEFAULT_USE_SYSTEM_PROPERTIES = false;
```
elguardian, line 158: usually if you have a proxy what is the common thing is that everything is affecting by the environment and not really at design time. 
You need to ajust this to a system property and read from it. Also would be good if you can set/override the properties like which proxy to use
you have examples here how to do this 
https://github.com/kiegroup/droolsjbpm-integration/blob/master/kie-server-parent/kie-server-api/src/main/java/org/kie/server/api/KieServerConstants.java#L39

fjtirado: The decission to use system properties or not is not base on the value of this private variable. The decision to use system properties or not is based on the value of the parameter passed to the constructor. 
This private constant is there to make easier maintance in case we decide to use system properties by default in the future. Now, for backward compatibility, the default value is false. If in future we decide to use them by default, we just need to change the constant (so this constant was introduce just becasue we have many constructor, if this class was built from scracth, we should use a builder pattern)
If we were going to control this functionality through system properties, changes in the constructor are not needed, but that was not the proposal of the JIRA, not consistent with the current structure of the code

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/BasicAuthRestWorkItemHandlerTest.java
```diff
@@ -104,6 +107,7 @@ public void setClientApiVersion() {
 
     @Test
     public void testGETOperation() {
+        System.setProperty(RESTWorkItemHandler.USE_SYSTEM_PROPERTIES, "true");
```
elguardian, line 110: This is not testing anything. Just covering the line you added. remove this.

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -1312,9 +1315,11 @@ protected HttpClient getHttpClient(Integer readTimeout,
             HttpClientBuilder clientBuilder = HttpClientBuilder.create()
                     .setDefaultRequestConfig(config);
 
-            HttpClient httpClient = clientBuilder.build();
+            if (Boolean.getBoolean(USE_SYSTEM_PROPERTIES))
```
elguardian, line 1318: open/close brackets

elguardian:
two minor things. the rest looks ok.

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/BasicAuthRestWorkItemHandlerTest.java
fjtirado: This is added for coverage and for testing that, if property is enable and there is not any system property added, this method behaves as beforee. 
So, I think is a very productive addition. I vote for keeping it

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
fjtirado: Rather than following the proposal of the JIRA, since the decission to use or not system properties should be taken as execution time, not desing time, it does not make sense to change current constructor, but provide a new system property org.kie.workspace.rest.useSystemProperties (default value false) which will determine if useSystemProperties is invoked or not when creating the http connection

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/BasicAuthRestWorkItemHandlerTest.java
elguardian: No it is not. 
1) You are modifying the behaviour of the test and make it dependent of the environment (so it is out of control)
2) You are not testing properly if the new property is using the system properties or not. e.g. one example would be to set a wrong address default adress (unreachable) and a proxy address setting in the system properties. then see if the system is taking or not what you are doing running a get operation against the proxied server.
This is not a proper test. Please remove and add a proper test.

fjtirado: Ok, I will remove and define a specific case.

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -718,24 +717,28 @@ public RESTWorkItemHandler(String username,
                                String password,
                                ClassLoader classLoader,
                                boolean doCacheClient,
-                               int totalPoolConnections,
+                                int totalPoolConnections,
```
MarianMacik, line 720: Unnecessary space.

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -718,24 +717,28 @@ public RESTWorkItemHandler(String username,
...
                                String handlingStrategy) {
-        this();
+        printProtocol();
         this.username = username;
         this.password = password;
         this.type = AuthenticationType.BASIC;
         this.classLoader = classLoader;
         this.doCacheClient = doCacheClient;
-        if (doCacheClient) {
+         if (doCacheClient) {
```
MarianMacik, line 730: Unnecessary space.

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -718,24 +717,28 @@ public RESTWorkItemHandler(String username,
...
             connectionManager.setMaxTotal(totalPoolConnections);
             connectionManager.setDefaultMaxPerRoute(maxPoolConnectionsPerRoute);
         }
         this.handlingProcessId = handlingProcessId;
         this.handlingStrategy = handlingStrategy;
     }
 
+    private void printProtocol() {
```
MarianMacik, line 738: Please rename the method to something like `logProtocol` so it is obvious that it doesn't print anything on stdout.

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/BasicAuthRestWorkItemHandlerTest.java
elguardian: Hi, you need to try to test this as well. otherwise looks good.

fjtirado: Additional test case using http.proxy system property has been added.

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/RestWorkItemHandlerTest.java
```diff
@@ -133,6 +136,34 @@ public void testGETOperation() {
         assertEquals("request to endpoint " + workItem.getParameter("Url") + " successfully completed OK",
                      responseMsg);
     }
+    
+    @Test
+    public void testGETWithProxyOperation() { 
+        System.setProperty(RESTWorkItemHandler.USE_SYSTEM_PROPERTIES, "true");
+        System.setProperty("http.proxyHost", "127.0.0.1");
```
elguardian, line 143: If I set this ip to an arbitrary one that does not exist. This test does not fail.
Looks to me that this is a faulty test.

fjtirado:
> 
> 
> Please see my comment about the test.

Do we have any proxy ip address to test this? Without it I do not see a way to configure a proper test. Also, do we really want to test JVM proxy functionality?

fjtirado:
Can I test the proxy by using https://github.com/adamfisk/LittleProxy?

elguardian:
@fjtirado https://github.com/kiegroup/jbpm/pull/1671

fjtirado:
Implemented proxy test in a different test class with library https://github.com/mock-server/mockserver

gmunozfe:
jenkins retest this

jbpm-workitems/jbpm-workitems-rest/pom.xml
```diff
@@ -19,6 +19,16 @@
   <properties>
     <java.module.name>org.jbpm.workitem.rest</java.module.name>
   </properties>
+  
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>org.mock-server</groupId>
```
elguardian, line 26: dependencies needs to be managed.

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -1311,10 +1315,18 @@ protected HttpClient getHttpClient(Integer readTimeout,
 
             HttpClientBuilder clientBuilder = HttpClientBuilder.create()
                     .setDefaultRequestConfig(config);
+            
+            
+            if (Boolean.getBoolean(USE_SYSTEM_PROPERTIES)) {
```
elguardian, line 1320: done only when HTTP_CLIENT_API_43 client.... when this is false does not use the proxy.

elguardian:
@fjtirado still lacking. one more push.

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/RestWorkItemHandlerTest.java
fjtirado: Test modified and moved to a new class

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
fjtirado: Changed

jbpm-workitems/jbpm-workitems-rest/pom.xml
fjtirado: Depends on [jbpm#1355](https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1355)

elguardian:
depends on https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/1355

elguardian:
@fjtirado good work. keep up

PR #1657, opened 2020-05-21
[JBPM-9126] Missing "subject" attribute in the UserTaskInstanceDesc

fjtirado:
Adding subject to existing constructor.
Changing query


elguardian:
this on hold till https://github.com/kiegroup/droolsjbpm-integration/pull/2098 pass checks (it should not affect but not sure.

elguardian:
good to go.

PR #1658, opened 2020-05-22
[JBPM-9171] Case traceability from a Task Instance

elguardian:
add correlation key and process type to a specialize class
add new query for retrieving the information


elguardian:
Jenkins retest this please.

PR #1659, opened 2020-05-22
[JBPM-9097] Case variable: "readonly" tag permits changing value after

fjtirado:
Now it is possible to assign a value to a case property when the
process is created, but a ViolationException will be thrown if trying
to update it when reopening.

Implementation note. In order to not change existing interface, which I
believe to be out of scope, the
parameter to decide if ViolationException should be thrown or not is
specified inside the parameters map.

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
@@ -30,6 +30,7 @@ import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
 import org.jbpm.casemgmt.impl.event.CaseEventSupport;
 import org.jbpm.casemgmt.impl.model.instance.CaseFileInstanceImpl;
 import org.jbpm.process.core.context.variable.VariableScope;
+import org.jbpm.process.core.context.variable.VariableViolationException;
 import org.jbpm.process.instance.InternalProcessRuntime;
 import org.jbpm.process.instance.ProcessInstance;
 import org.jbpm.services.api.ProcessService;
@@ -81,6 +82,16 @@ public class AddDataCaseFileInstanceCommand extends CaseCommand<Void> {
         // apply authorization
         authorizationManager.checkDataAuthorization(caseFile.getCaseId(), caseFile, parameters.keySet());
         
+        // check read only variables, due to dependencies between firexxxCaseData methods, we need this logic here, we cannot reuse VariableScopeInstance.setVariable
+        org.jbpm.process.instance.ProcessInstance pi = (org.jbpm.process.instance.ProcessInstance) ksession.getProcessInstance(processInstanceId);
+        VariableScope variableScope = (VariableScope) pi.getContextContainer().getDefaultContext(VariableScope.VARIABLE_SCOPE);
+        for (Map.Entry<String, Object> entry : parameters.entrySet()) {
+            Object oldValue = caseFile.getData(entry.getKey());
+            if (oldValue != null && !oldValue.equals(entry.getValue()) && variableScope.isReadOnly(VariableScope.CASE_FILE_PREFIX + entry.getKey())) {
+                throw new VariableViolationException(pi.getId(), entry.getKey(), "variable is read only and cannot be replaced");
+            }
+        }
+
         FactHandle factHandle = ksession.getFactHandle(caseFile);
         
         CaseEventSupport caseEventSupport = getCaseEventSupport(context);
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
@@ -25,6 +25,8 @@ import org.drools.core.ClassObjectFilter;
 import org.jbpm.casemgmt.api.auth.AuthorizationManager;
 import org.jbpm.casemgmt.api.model.instance.CaseFileInstance;
 import org.jbpm.casemgmt.impl.event.CaseEventSupport;
+import org.jbpm.process.core.context.variable.VariableScope;
+import org.jbpm.process.core.context.variable.VariableViolationException;
 import org.kie.api.runtime.Context;
 import org.kie.api.runtime.KieSession;
 import org.kie.api.runtime.rule.FactHandle;
@@ -40,9 +42,11 @@ public class RemoveDataCaseFileInstanceCommand extends CaseCommand<Void> {
 
     private List<String> variableNames;
     private AuthorizationManager authorizationManager;
+    private Long processInstanceId;
     
-    public RemoveDataCaseFileInstanceCommand(IdentityProvider identityProvider, List<String> variableNames, AuthorizationManager authorizationManager) {                
+    public RemoveDataCaseFileInstanceCommand(Long processInstanceId, IdentityProvider identityProvider, List<String> variableNames, AuthorizationManager authorizationManager) {                
         super(identityProvider);
+        this.processInstanceId = processInstanceId;
         this.variableNames = variableNames;        
         this.authorizationManager = authorizationManager;
     }
@@ -59,7 +63,15 @@ public class RemoveDataCaseFileInstanceCommand extends CaseCommand<Void> {
         
         // apply authorization
         authorizationManager.checkDataAuthorization(caseFile.getCaseId(), caseFile, variableNames);
-        
+          
+        org.jbpm.process.instance.ProcessInstance pi = (org.jbpm.process.instance.ProcessInstance) ksession.getProcessInstance(processInstanceId);
+        VariableScope variableScope = (VariableScope) pi.getContextContainer().getDefaultContext(VariableScope.VARIABLE_SCOPE);
+        for (String name: variableNames) {
+            if (caseFile.getData(name) != null && variableScope.isReadOnly(VariableScope.CASE_FILE_PREFIX+name)) {
+                throw new VariableViolationException(pi.getId(), name,"Variable '" + name + "' is already set and is marked as read only");
+            }
+        }
+         
         FactHandle factHandle = ksession.getFactHandle(caseFile);
         
         Map<String, Object> remove = new HashMap<>();        
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
@@ -78,16 +78,12 @@ public class ReopenCaseCommand extends CaseCommand<Void> {
         
         KieSession ksession = ((RegistryContext) context).lookup( KieSession.class );
                                
-        CaseFileInstance caseFile = getCaseFile(ksession, caseId);                        
+        CaseFileInstance caseFile = getCaseFile(ksession, caseId);     
         
         caseEventSupport.fireBeforeCaseReopened(caseId, caseFile, deploymentId, caseDefinitionId, data);
         
         logger.debug("Updating case file in working memory");
         FactHandle factHandle = ksession.getFactHandle(caseFile);
-        ((CaseFileInstanceImpl)caseFile).setCaseReopenDate(new Date());
-        if (data != null && !data.isEmpty()) {
-            caseFile.addAll(data);
-        }
         ksession.update(factHandle, caseFile);
         
         logger.debug("Starting process instance for case {} and case definition {}", caseId, caseDefinitionId);
@@ -95,12 +91,30 @@ public class ReopenCaseCommand extends CaseCommand<Void> {
         Map<String, Object> params = new HashMap<>();
         // set case id to allow it to use CaseContext when creating runtime engine
         params.put(EnvironmentName.CASE_ID, caseId);
+        final Map<String, Object> caseData = caseFile.getData();
+        
+        
+        for (Map.Entry<String, Object> entry : caseData.entrySet()) {
+            params.put(VariableScope.CASE_FILE_PREFIX + entry.getKey(), entry.getValue());
+        }
+        
+        if (data != null)   {
+            for (Map.Entry<String, Object> entry : data.entrySet()) {
+                params.put(VariableScope.CASE_FILE_PREFIX + entry.getKey(), entry.getValue());
+            }
+        }
+
         long processInstanceId = processService.startProcess(deploymentId, caseDefinitionId, correlationKey, params);
+        
+        ((CaseFileInstanceImpl)caseFile).setCaseReopenDate(new Date());
+        if (data != null) {
+            caseFile.addAll(data);
+        }
+        
         logger.debug("Removing case file from working memory to allow refiring of rules...");
         ksession.delete(factHandle);
         ksession.insert(caseFile);
-        final Map<String, Object> caseData = caseFile.getData();
-        if (caseData != null && !caseData.isEmpty()) {
+        if (!caseData.isEmpty()) {
             processService.execute(deploymentId, CaseContext.get(caseId), new ExecutableCommand<Void>() {
     
                 private static final long serialVersionUID = -7093369406457484236L;
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
@@ -112,10 +112,8 @@ public class StartCaseCommand extends CaseCommand<Void> {
         });
         commands.add(commandsFactory.newInsert(caseFile));
         commands.add(commandsFactory.newFireAllRules());
-
         BatchExecutionCommand batch = commandsFactory.newBatchExecution(commands);
         processService.execute(deploymentId, CaseContext.get(caseId), batch);
-
         logger.debug("Starting process instance for case {} and case definition {}", caseId, caseDefinitionId);
         CorrelationKey correlationKey = correlationKeyFactory.newCorrelationKey(caseId);
         Map<String, Object> params = new HashMap<>();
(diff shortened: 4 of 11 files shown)


jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
```diff
@@ -73,6 +74,15 @@ public Void execute(Context context) {
         
         KieSession ksession = ((RegistryContext) context).lookup( KieSession.class );
         
+        org.jbpm.process.instance.ProcessInstance pi = (org.jbpm.process.instance.ProcessInstance) ksession.getProcessInstance(processInstanceId);
+        VariableScope variableScope = (VariableScope) pi.getContextContainer().getDefaultContext(VariableScope.VARIABLE_SCOPE);
+        
+        for (String name: parameters.keySet())
+        {
+            if (variableScope.isReadOnly(VariableScope.CASE_FILE_PREFIX+name))
```
gmunozfe, line 82: It should control also that variable has a previous value assigned which tries to be overwritten

fjtirado: Done

gmunozfe:
Though it's not part of the jira, `removeDataFromCaseFile` should throw `VariableViolationException` when the variable to remove is a readonly variable which already has assigned value.

gmunozfe:
jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
```diff
@@ -125,6 +123,7 @@ public void matchCreated(MatchCreatedEvent event) {
         }
         // set case id to allow it to use CaseContext when creating runtime engine
         params.put(EnvironmentName.CASE_ID, caseId);
+        params.put(VariableScope.FORCE_UPDATE, Boolean.TRUE);
```
elguardian, line 126: I am not very convince about this.
If you are starting a case o restarting one to avoid any problem with variable you just need to check whether the oldValue and the newOne are the same instead of this. This will fix the issue regarding using the newCaseFileInstance + startCase... reopening will be more tricky as drools and facts are involved.

fjtirado: Additional parameter has been removed. 
assuming assigment of equal object is not considered assigment, we can avoid it

gmunozfe:
Test failures related

fjtirado:
jenkins retest this please

gmunozfe:
jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
```diff
@@ -85,6 +87,16 @@ public Void execute(Context context) {
         
         CaseEventSupport caseEventSupport = getCaseEventSupport(context);
         caseEventSupport.fireBeforeCaseDataAdded(caseFile.getCaseId(), caseFile, caseFile.getDefinitionId(), parameters);
+        
+        org.jbpm.process.instance.ProcessInstance pi = (org.jbpm.process.instance.ProcessInstance) ksession.getProcessInstance(processInstanceId);
```
elguardian, line 91: logic spreaded between VariableScopeInstance and here.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
```diff
@@ -60,6 +64,17 @@ public Void execute(Context context) {
         // apply authorization
         authorizationManager.checkDataAuthorization(caseFile.getCaseId(), caseFile, variableNames);
         
+        
+        org.jbpm.process.instance.ProcessInstance pi = (org.jbpm.process.instance.ProcessInstance) ksession.getProcessInstance(processInstanceId);
```
elguardian, line 68: logic spreaded between VariableScopeInstance and here.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
```diff
@@ -78,29 +78,43 @@ public Void execute(Context context) {
...
+        if (data != null) {
+            caseFile.addAll(data);
+        }
+        
         logger.debug("Removing case file from working memory to allow refiring of rules...");
         ksession.delete(factHandle);
         ksession.insert(caseFile);
-        final Map<String, Object> caseData = caseFile.getData();
-        if (caseData != null && !caseData.isEmpty()) {
+        if (!caseData.isEmpty()) {
```
elguardian, line 117: why removed the check of null ?

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
```diff
@@ -112,10 +111,7 @@ public void matchCreated(MatchCreatedEvent event) {
         });
         commands.add(commandsFactory.newInsert(caseFile));
         commands.add(commandsFactory.newFireAllRules());
-
-        BatchExecutionCommand batch = commandsFactory.newBatchExecution(commands);
-        processService.execute(deploymentId, CaseContext.get(caseId), batch);
-
+        processService.execute(deploymentId, CaseContext.get(caseId), commandsFactory.newBatchExecution(commands));
```
elguardian, line 114: unneeded

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
```diff
@@ -113,7 +113,7 @@ public Date getCaseEndDate() {
 
     @Override
     public Map<String, Object> getData() {
-        return this.data;
+        return Collections.unmodifiableMap(this.data);
```
elguardian, line 116: unneeded

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
```diff
@@ -40,9 +42,11 @@
 
     private List<String> variableNames;
     private AuthorizationManager authorizationManager;
+    private Long processInstanceId;
     
-    public RemoveDataCaseFileInstanceCommand(IdentityProvider identityProvider, List<String> variableNames, AuthorizationManager authorizationManager) {                
+    public RemoveDataCaseFileInstanceCommand(Long processInstanceId, IdentityProvider identityProvider, List<String> variableNames, AuthorizationManager authorizationManager) {
```
elguardian, line 47: this will required a full downstream compilation at least. Commands are sensitive. @gmunozfe

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/cases/UserTaskCaseRestrictedCaseFileItem.bpmn2
```diff
@@ -18,6 +18,14 @@
         </tns:metaData>
       </bpmn2:extensionElements>
     </bpmn2:property>
+    
+    <bpmn2:property id="caseFile_r" itemSubjectRef="_sItem" name="caseFile_r">
```
elguardian, line 22: generate new file for these test.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/ReopenCaseCommand.java
fjtirado: because it was unneded and reported as a code smell

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
fjtirado: this is to prevent a backdoor update of the intenal map, it should stay to make the solution more robust

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
fjtirado: this is the only way I found to access variable readonly definition in RemoveDataCaseFileInstace. Lets do the full downstream compilation

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/cases/UserTaskCaseRestrictedCaseFileItem.bpmn2
fjtirado: Duplicating bpmn definitions (as duplicating code through copy paste), does not seem a good idea. The modified file was initially created to check required constraint. It makes full sense to reuse it to check readonly constraint.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
fjtirado: VariableScopeInstance is not really used in this scenario. Please check getVariable, which relies on CaseFileInnstan

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
fjtirado: VariableScopeInstance is not really used in this scenario. Please check getVariable, which relies on CaseFileInnstan

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/StartCaseCommand.java
fjtirado: what was unnededed is the definition of a variable which is used in one place, this is an improvement of previous code and should stay

fjtirado: Change rolled back

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/model/instance/CaseFileInstanceImpl.java
fjtirado: Change rolled back

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
fjtirado: Code changed to reuse setVariable on VariableScopeInstance

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
fjtirado: Remove logic is very particular of CaseFileInstance (the events published are specific), so there is not really logic spread besided the call to variableScopeInstance.isReadOnly, which I believe is probably fine.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/java/org/jbpm/casemgmt/impl/CaseServiceImplTest.java
```diff
@@ -146,6 +146,7 @@
         processes.add("cases/UserTaskCaseRequiredCaseFileItem.bpmn2");
         processes.add("cases/UserTaskCaseRestrictedCaseFileItem.bpmn2");
         processes.add("cases/UserTaskCaseRequiredRestrictedCaseFileItem.bpmn2");
+        processes.add("cases/UserTaskCaseReadOnlyCaseFileItem.bpmn2");
```
gmunozfe, line 149: It doesn't match with the file name (case sensitive)
```suggestion
        processes.add("cases/UserTaskCaseReadonlyCaseFileItem.bpmn2");
```
or change the file name.

fjtirado: file name changed

elguardian:
@MarianMacik Can you check. I want another couple of eyes on this.

fjtirado:
jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
fjtirado: After some failed tests, I think the original code is good there, reusing VariableScopeInstance.setVariable cause case duplication because of the way the listeners are handled there. 
I do not think it is wise to change it more deeply.

fjtirado:
jenkins retest this

gmunozfe:
jenkins retest this

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
MarianMacik, line 47: In theory somebody can be already using that via `ProcessService` to be run as a separate command, so we can break backward compatibility, but if there is no other way, it is ok from my side.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
```diff
@@ -59,7 +63,15 @@ public Void execute(Context context) {
         
         // apply authorization
         authorizationManager.checkDataAuthorization(caseFile.getCaseId(), caseFile, variableNames);
-        
+          
+        org.jbpm.process.instance.ProcessInstance pi = (org.jbpm.process.instance.ProcessInstance) ksession.getProcessInstance(processInstanceId);
+        VariableScope variableScope = (VariableScope) pi.getContextContainer().getDefaultContext(VariableScope.VARIABLE_SCOPE);
+        for (String name: variableNames) {
+            if (caseFile.getData(name) != null && variableScope.isReadOnly(VariableScope.CASE_FILE_PREFIX+name)) {
+                throw new VariableViolationException(pi.getId(), name,"variable is read only and cannot be removed");
```
MarianMacik, line 71: Maybe use the same message as in `VariableScopeInstance`?
`throw new VariableViolationException(getProcessInstance().getId(), name, "Variable '" + name + "' is already set and is marked as read only");`

jbpm-test-coverage/src/main/java/org/jbpm/test/entity/MedicalRecord.java
```diff
@@ -110,9 +110,6 @@ public boolean equals(Object obj) {
         if ((this.description == null) ? (other.description != null) : !this.description.equals(other.description)) {
             return false;
         }
-        if (this.patient != other.patient && (this.patient == null || !this.patient.equals(other.patient))) {
-            return false;
-        }
```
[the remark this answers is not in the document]
fjtirado: This was changed to avoid a circular dependency between equals that causes StackOverflowException

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/AddDataCaseFileInstanceCommand.java
```diff
@@ -81,6 +82,16 @@ public Void execute(Context context) {
         // apply authorization
         authorizationManager.checkDataAuthorization(caseFile.getCaseId(), caseFile, parameters.keySet());
         
+        // check read only variables, due to dependencies between firexxxCaseData methods, we need this logic here, we cannot reuse VariableScopeInstance.setVariable
```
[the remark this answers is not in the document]
fjtirado: Yes, after trying to modify original code to call setVariable instance and centralize some logic, some test were failing because the trigger call needs to be invoked at a very specific moment of time (or the case will be added twice), so I was adding that comment to avoid the temptation to do the same change in the future

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
fjtirado: Im aware of the potential backward compatiiliry problem, but I cannot find other way, processInstanceId is needed to check the readonly flag (and now both command constructors, add and remove data case are symetrical)

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/command/RemoveDataCaseFileInstanceCommand.java
fjtirado: Changed

fjtirado:
jenkins retest this

elguardian:
@MarianMacik thx for the feedback

PR #1660, opened 2020-05-25
[RHPAM-2960] Case instance update fails with java.io.NotSerializableE…

elguardian:
…xception: org.jbpm.services.task.persistence.JPATaskModelFactory for cases containing multiple SLAs

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/instance/timer/TimerManager.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/instance/timer/TimerManager.java
@@ -16,6 +16,13 @@
 
 package org.jbpm.process.instance.timer;
 
+import java.io.Serializable;
+import java.util.Collection;
+import java.util.Date;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
 import org.drools.core.common.InternalKnowledgeRuntime;
 import org.drools.core.common.InternalWorkingMemory;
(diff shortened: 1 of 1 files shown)


elguardian:
depends on https://github.com/kiegroup/drools/pull/2916

elguardian:
Jenkins retest this please.

jbpm-flow/src/main/java/org/jbpm/process/instance/timer/TimerManager.java
```diff
@@ -261,9 +261,10 @@ public void deserialize(MarshallerReaderContext inCtx, Timer timer) throws Class
                     long now = pctx.getKnowledgeRuntime().getSessionClock().getCurrentTime();
                     // overdue timer                    
                     if (then < now) {
-                        trigger = new OverdueTrigger(trigger, pctx.getKnowledgeRuntime());
+                        trigger = new OverdueTrigger(trigger);
                     }
                 }
+                trigger.initialize(pctx.getWorkingMemory().getKnowledgeRuntime());
```
afalhambra, line 267: Current version is getting `kruntime` from `pctx.getKnowledgeRuntime()` method. So not really sure why we are now retrieving in a different way by using `pctx.getWorkingMemory().getKnowledgeRuntime()`?
If `kruntime` is not an instance of `InternalWorkingMemory` this will return a null and thus it might fail later on - this wasn't previous/current behavior.
Can you please clarify?

PR #1653, opened 2020-05-19
[RHPAM-2934] Performance bottleneck search variables in Postgres

elguardian:

PR #1661, opened 2020-05-25
PLANNER-1985:  [Tasks Assignment] Add task assignment tables to the DB scripts

wmedvede:


rsynek:
@wmedvede (not required for merging) I remember the figures about task assigning performance you shared showed a significant portion of time is spent by doing queries. I wonder if adding a couple of indexes based on the queries the extension do would help?

wmedvede:
@rsynek this PR is also related with https://github.com/kiegroup/droolsjbpm-integration/pull/2104

wmedvede:
jenkins retest this

wmedvede:

wmedvede:
Hi  @rsynek again the full downstream is failing.

BUT considering that this PR defines only some .sql scripts there's no way that this PR can break the build. If you could finish the integration tests I strongly suggest moving this forward.

Thanks.

jbpm-db-scripts/src/main/resources/db/ddl-scripts/sybase/task_assigning_tables_sybase.sql
```diff
@@ -0,0 +1,15 @@
+    create table PlanningTask (
+        taskId numeric(19,0) not null,
+        assignedUser varchar(255),
+        taskIndex int not null,
+        lastModificationDate datetime,
+        published smallint not null,
+        primary key (taskId)
+    )
```
rsynek, line 8: The jBPM sql script uses "lock datarows" for every table it creates. Is there any motivation for keeping the default "page locking" of sybase [1]?

[1] http://infocenter.sybase.com/help/index.jsp?topic=/com.sybase.dc20021_1251/html/locking/locking9.htm

gmunozfe:
hi @wmedvede , as a new table is defined, upgrade scripts should be also created/updated. Will them be covered in other PR?

rsynek:
> hi @wmedvede , as a new table is defined, upgrade scripts should be also created/updated. Will they be covered in other PR?

This PR introduces the first version of DDL for task assigning; there is nothing to upgrade from yet. I think the update script needs to be provided only if we change the DDL in the future.

elguardian:
@rsynek there is something to upgrade from people working with older versions of the product.

rsynek:
@elguardian Do I understand it correctly, that users of the previous version of jBPM are supposed to always ignore the `create` scripts and use the `update` scripts?

Does it apply even for this situation:
- task assigning is an optional feature, so its DB schema is added on top of jBPM DB schema
- this is the first time the DDL scripts for task assigning are introduced
?

jbpm-db-scripts/src/main/resources/db/ddl-scripts/sybase/task_assigning_tables_sybase.sql
wmedvede: good catch, no there's no reason. Just forgot to add it, I'll fix it.

wmedvede:
> requires upgrade scripts.

@elguardian I see, thanks. I'll add them, just quick query for clarifying, so given the sequence of change scripts we have now:
.....
....
jbpm-7.6-to-7.7.sql
jbpm-7.7-to-7.11.sql

and that we are in 7.39.0-SNAPSHOT

this new update scripts should be:

jbpm-7.11-to-7.39.sql

right?

gmunozfe:
@wmedvede there'll be new upgrade scripts bpms/jbpm-7.38-to-7.39.sql after this [jbpm#1653](https://github.com/kiegroup/jbpm/pull/1653)

wmedvede:
> @wmedvede there'll be new upgrade scripts bpms/jbpm-7.38-to-7.39.sql after this [jbpm#1653](https://github.com/kiegroup/jbpm/pull/1653)

ok, I'll add my stuff 

> @wmedvede there'll be new upgrade scripts bpms/jbpm-7.38-to-7.39.sql after this [jbpm#1653](https://github.com/kiegroup/jbpm/pull/1653)

@rsynek @elguardian Important thing I was thinking here, this task assigning development is an optional feature that people can plug if they want. And this is why the create schema scrips are in a separate script in the same way that we have he quartz.sql scrips. So the version update scripts must be separate so, note that we have those update scripts for quartz too. So my update scripts shouldn't go in the jbpm-X.XX-to-Y.YY.sql files

wmedvede:
@rsynek suggested index on assignedUser field where added in all DBMSs and sybase script fixed too.

wmedvede:
@rsynek @elguardian regarding the thread about the update scripts as commented above, this is an optional feature that has separate scrips the same as the quartz scripts. So main jbpm scripts won't include the specifics for this feature.

rsynek:
@elguardian @wmedvede we need to merge this PR quickly, as there are integration tests depending on it. If there are still doubts about the necessity of update scripts, let's solve it on the chat. Thanks!

elguardian:
@wmedvede you can go ahead and merge... but the scripting stuff being optional or not must be set in place. We also check quartz scripts because there are customers using those.
@mareknovotny

rsynek:
@elguardian Thanks, merging. If there are still any doubts, let's discuss it in the chat.

PR #1665, opened 2020-05-28
Register dialect for validating Quartz-mariaDB

gmunozfe:
Previously, dbAllocator was instantiating MariaDB instances with "MySQL" db name, but now Quartz schema validation is failing because the database name was missing.

### DIFF
--- a/jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/quartzdialects/QuartzDialectResolver.java
+++ b/jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/quartzdialects/QuartzDialectResolver.java
@@ -30,7 +30,7 @@ public class QuartzDialectResolver implements DialectResolver {
     private static Map<String, Dialect> DIALECT_BY_NAME = new HashMap<>();
 
     {
-        registerDialect(new MySQLCustomDialect(), "MySQL");
+        registerDialect(new MySQLCustomDialect(), "MySQL", "MariaDB");
         registerDialect(new DB2CustomDialect(), "DB2/LINUXX8664");
         registerDialect(new OracleCustomDialect(), "Oracle");
         registerDialect(new PostgreSQLCustomDialect(), "PostgreSQL", "EnterpriseDB");


jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/quartzdialects/QuartzDialectResolver.java
```diff
@@ -30,6 +30,7 @@
     private static Map<String, Dialect> DIALECT_BY_NAME = new HashMap<>();
 
     {
+        registerDialect(new MySQLCustomDialect(), "MariaDB");
```
afalhambra, line 33: Wouldn't be clearer to use (or create) a new custom dialect for MariaDB instead of using `MySQLCustomDialect` to decoupling?

gmunozfe: No, it's not worth it. MariaDB is 100% compatible with mysql and currently we have only scripts for mysql

jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/quartzdialects/QuartzDialectResolver.java
```diff
@@ -30,6 +30,7 @@
...
         registerDialect(new MySQLCustomDialect(), "MySQL");
```
afalhambra, line 34: Just a suggestion then as both are sharing some custom dialect, but either ways are fine to me. Up to you.
```suggestion
        registerDialect(new MySQLCustomDialect(), "MySQL", "MariaDB");
```

gmunozfe: Thanks! done

gmunozfe:
jenkins retest this

gmunozfe:
@kiegroup/gatekeepers could you merge it?

PR #1666, opened 2020-06-01
[RHPAM-2591] Task identifers are recorded on the DB with negative numbers

elguardian:
correction in postgresql scripts for springboot apps


elguardian:
> Just a couple of comments/questions:
> 
> * I would not include oracle script changes in this PR. This is supposed to be only for Postgresql.

I realize the problem with Oracle script (lack of semicolon at the end of sql sentence)

> * Why adding two new files for fixing this seq issue? Why not amend the ones already in place? I think this may lead to confusion/misunderstanding.

because they are not working in the same way. One is for our kie-server. The other one for spring boot

> * What about upgrade scripts? Should this fix not be included in those scripts as well?

Not really. The problem is during DDL definition not really any problem with an old version. So there is no need for amend.

afalhambra:
> > Just a couple of comments/questions:
> > 
> > * I would not include oracle script changes in this PR. This is supposed to be only for Postgresql.
> 
> I realize the problem with Oracle script (lack of semicolon at the end of sql sentence)
> 
> > * Why adding two new files for fixing this seq issue? Why not amend the ones already in place? I think this may lead to confusion/misunderstanding.
> 
> because they are not working in the same way. One is for our kie-server. The other one for spring boot

I think problem with sequences should be fixed in the current sql file, instead of duplicating/adding a new sql file. If we fix the current sql script (amending sequences only) we will ensure this will work for both kie-server and for using spring-boot and there won't be any need to add any additional sql file. 

> 
> > * What about upgrade scripts? Should this fix not be included in those scripts as well?
> 
> Not really. The problem is during DDL definition not really any problem with an old version. So there is no need for amend.

But let's say someone is already using a previous version, they might want to upgrade to a new version to fix this issue because let's say they start setting this hibernate property to true in a spring-boot project.

elguardian:
>I think problem with sequences should be fixed in the current sql file, instead of duplicating/adding a new sql file. If we fix the current sql script (amending sequences only) we will ensure this will work for both kie-server and for using spring-boot and there won't be any need to add any additional sql file.

Modifiying the DDL would be ok but I need to touch the code and have another flag which is far worse than having another ddl. Also requires upgrade scripts. It is error prone, it is more difficult to asses as you would have users with/without the new mapping. This is really a design problem how jbpm was design to work with spring boot.

> But let's say someone is already using a previous version, they might want to upgrade to a new version to fix this issue because let's say they start setting this hibernate property to true in a spring-boot project.

The jira is about DDL not really migration. I am not broading the scope of the jira.

PR #1668, opened 2020-06-02
JBPM-9151 API return 200 if containerId does not exist

fjtirado:
Check if deployment id is available in the in memory list and throw
proper exception if not.

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/BPMN2DataServicesTest.java
+++ b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/BPMN2DataServicesTest.java
@@ -347,6 +347,7 @@ public class BPMN2DataServicesTest extends AbstractKieServicesBaseTest {
 
         // now let's undeploy the unit
         deploymentService.undeploy(deploymentUnit);
+        units.remove(deploymentUnit);
 
         try {
             bpmn2Service.getProcessDefinition(deploymentUnit.getIdentifier(), processId);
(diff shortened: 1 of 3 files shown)


jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/BPMN2DataServicesTest.java
```diff
@@ -347,6 +347,7 @@ public void testHumanTaskProcessBeforeAndAfterUndeploy() {
 
         // now let's undeploy the unit
         deploymentService.undeploy(deploymentUnit);
+        units.remove(deploymentUnit);
```
elguardian, line 350: why this ? the change to RuntimeDataServiceImpl.java looks to me like completely compatible regarding this.

elguardian:
some minor question.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/BPMN2DataServicesTest.java
fjtirado: This change was suggested by Gonzalo. The target is to keep client units array consistent with the server.

gmunozfe: Yes, that's the reason, it's the same in other tests:
https://github.com/kiegroup/jbpm/blob/74b525d1454e80de85ae02a8a5ffe77bf85706e7/jbpm-container-test/jbpm-in-container-test/jbpm-container-test-suite/src/test/java/org/jbpm/test/container/archive/EJBService.java#L261-L263

PR #1672, opened 2020-06-04
JBPM-6574 Timer name to be added to job_name for triggers/jobs

fjtirado:
Adding name to Timer and TimerInstance. Name is lated user to generate
quartz job name


fjtirado:
jenkins retest this

jbpm-flow/src/main/java/org/jbpm/process/core/timer/Timer.java
```diff
@@ -32,7 +32,16 @@
     private String period;
     private String date;
     private int timeType;
+    private String name;
```
gmunozfe, line 35: Add name to the `toString` of this class

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/timer/GlobalQuartzDBTimerServiceTest.java
```diff
@@ -342,10 +342,12 @@ public void testTimerRequiresRecoveryFlagSet() throws Exception {
             connection = ((DataSource)InitialContext.doLookup("jdbc/jbpm-ds")).getConnection();
             stmt = connection.createStatement();
 
-            ResultSet resultSet = stmt.executeQuery("select REQUESTS_RECOVERY from QRTZ_JOB_DETAILS");
+            ResultSet resultSet = stmt.executeQuery("select REQUESTS_RECOVERY, JOB_NAME from QRTZ_JOB_DETAILS");
             while(resultSet.next()) {
                 boolean requestsRecovery = resultSet.getBoolean(1);
                 assertEquals("Requests recovery must be set to true", true, requestsRecovery);
+                String jobName = resultSet.getString(2);
+                assertTrue(jobName + " does not contain ticket name", jobName.contains("Boundary Event"));
```
gmunozfe, line 350: typo "timer", not ticket

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/ThreadPoolSchedulerService.java
```diff
@@ -93,18 +92,11 @@ public JobHandle scheduleJob(Job job, JobContext ctx, Trigger trigger) {
...
-                if (processCtx instanceof StartProcessJobContext) {
-                    jobname = "StartProcess-"+((StartProcessJobContext) processCtx).getProcessId()+ "-" + processCtx.getTimer().getId();
-                }
-                if (activeTimer.containsKey(jobname)) {
-                    return activeTimer.get(jobname);
-                }
-            
+            String jobname = QuartzNameHelper.getJobName(ctx);
+            if (activeTimer.containsKey(jobname)) {
+                return activeTimer.get(jobname);
```
gmunozfe, line 97: This replacement is not exactly as it was before, as only has to return the activeTime if ctx is instance of ProcessJobContext (missing this condition for all the block, though it's evaluated also inside helper class), wdyt?

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzNameHelper.java
```diff
@@ -0,0 +1,64 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
gmunozfe, line 2: update to 2020

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzNameHelper.java
```diff
@@ -0,0 +1,64 @@
...
+ */
+package org.jbpm.process.core.timer.impl;
+
+import org.drools.core.time.JobContext;
+import org.jbpm.process.core.timer.NamedJobContext;
+import org.jbpm.process.instance.timer.TimerManager.ProcessJobContext;
+import org.jbpm.process.instance.timer.TimerManager.StartProcessJobContext;
+import org.kie.api.runtime.EnvironmentName;
+
+class QuartzNameHelper {
```
gmunozfe, line 24: Shouldn't it be "public"?

fjtirado:
> 
> 
> Shouldn't be useful to compose the jobName in a similar way for EJBTimers?

Yes, QuartzNameHelper has been rename to JobNameHelper and used for EJBTimers too. JdkTimers do not use job name.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzNameHelper.java
fjtirado: Once this class is used also for EJbTimer, yes, it should be public, before was package becuase it was only used for Quartz ones

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/QuartzNameHelper.java
fjtirado: Done

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/ThreadPoolSchedulerService.java
fjtirado: I think regardless of the job type, it should not be created again if already exist a unique job name for it

jbpm-flow/src/main/java/org/jbpm/process/core/timer/Timer.java
fjtirado: done

fjtirado:
> 
> Modify JBPMMessages.java is not an option.
The change in JBPMMessages (Protobuf serialization) is backward compatible.  Is there any reason why this should not be changed?
> Name of the node is not proper information as a timer is more generic entity and does not affect only to a specific node. Also usually is null.
Which is the alternative to use node name? 

Let me justify the current proposal. 
The JIRA ask for using timer name in job name. Before this PR, the only place I have seen there is a timer name, that name is taken from the node name (ProcessAdminServiceBase.convertTimerIntance method in the service layer) 
But the TimerInstance object that is used for creating Job Name is the one associated to the session (org.jbpm.process.instance.timer.TimerInstance) which does not have any name, so it feels natural to add a name field to that one in order to use it in different timer services when generating the job name. 
Once this path is taken, the modification of JBPMMessages is a requirement to make it work, because since that timer is associated to the session, it is persisted along with other session data using protobuf. But, again, this change is backward compatible, so there should be a strong reason to not do that (since the alternative look worse to me, to traverse the node tree in order to search the name of the node associated to that timer, as it is currently done in the service layer)  
Finally, the timer name value is taken from the BPMN process (the proper node depending on the type of timer), which is the only source I can think of. If there is a different source for the timer name available, the current proposal does not make sense and there is no point in discussing the protobuf question. 

So, on summary, two questions, one subordinated to the other, should be asked
1) From where we are going to take the timer name?
2) if the answer to 1) is from BPMN node, then the question is, what is a cleaner approach? To add the name to the timer object associated to the session and created when the bpmn is parsed? (in which case, protobuf needs to be manipulated) Or search for the name in the BPMN note every time the timer name is needed? Since I consider the former a cleaner approach, unless there is a reason Im unaware to not change the protobuf, since it has been already done and everything is working, I need to understand the reason to rollback that change.

elguardian:
@fjtirado please clean up the failure tests.

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
Reverting back to protofuf implementation since accessing the node instance tree while registering timer creates multithreading issues and unexpected exception while testing

gmunozfe:
Looks good to me, great work @fjtirado !

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/ThreadPoolSchedulerService.java
```diff
@@ -91,21 +91,15 @@ public void shutdown() {
...
-                    jobname = "StartProcess-"+((StartProcessJobContext) processCtx).getProcessId()+ "-" + processCtx.getTimer().getId();
-                }
-                if (activeTimer.containsKey(jobname)) {
-                    return activeTimer.get(jobname);
-                }
-            
+            String jobname = JobNameHelper.getJobName(ctx, id);
+            if (activeTimer.containsKey(jobname)) {
+                return activeTimer.get(jobname);
             }
```
MarianMacik, line 100: @mswiderski So until now if the `JobContext` was other than  `ProcessJobContext` the job name was left null and such job wasn't put in the map of active timers, right? So for example for `TaskDeadlineJobContext` (instance of NamedJobContext) we might have ended up with the same timer scheduled more than once? So was that actually a bug?

As right now we will create a job name for every possible JobContext using `JobNameHelper` and that also means all timers will be remembered.

jbpm-flow/src/main/java/org/jbpm/process/core/timer/JobNameHelper.java
```diff
@@ -0,0 +1,67 @@
...
+
+    public static String getJobName(JobContext ctx, long id) {
+        return getJobName(ctx, getGroupName(ctx), id);
+    }
+
+    public static String getJobName(JobContext ctx, String groupName, long id) {
+        String jobName;
+        if (ctx instanceof ProcessJobContext) {
+            ProcessJobContext processCtx = (ProcessJobContext) ctx;
+            final String timerName = "-" + (processCtx.getTimer().getName() != null && !processCtx.getTimer().getName().isEmpty() ? processCtx.getTimer().getName() + "-" : "") + processCtx.getTimer().getId();
```
MarianMacik, line 35: Just a readability thing `processCtx.getTimer().getName()` could be extracted on a new line so the whole line is shorter and easier to read.

PR #1679, opened 2020-06-16
[BAPL-1623] Case traceability from a Task Instance

fjtirado:
CorrelationKey and ProcessType added to UserTaskDescription.

Existing APIs(getTaskById, getTaskByWorkItem and getTaskByProcess) are
reused, since I believe is preferable to assume the performance cost of
an extra join rather than having to maintain a new API for every
attribute combination (specially when there is already a join because of
the introduction of subject attribute)

### DIFF
--- a/jbpm-services/jbpm-services-api/src/build/revapi-config.json
+++ b/jbpm-services/jbpm-services-api/src/build/revapi-config.json
@@ -89,26 +89,61 @@
                     "methodName": "startProcessFromNodeIds",
                     "elementKind": "method",
                     "justification": "start nodes from an arbitray node"
-                }
-                ,
+                },
                 {
-				   "code": "java.method.addedToInterface",
-  					"new": "method java.lang.String org.jbpm.services.api.model.UserTaskInstanceDesc::getSubject()",
-   					"package": "org.jbpm.services.api.model",
-   					"classSimpleName": "UserTaskInstanceDesc",
-   					"methodName": "getSubject",
-   					"elementKind": "method",
-   					"justification": "subject should be returned by REST service"
- 				},
- 				{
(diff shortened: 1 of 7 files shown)


jbpm-services/jbpm-services-api/src/build/revapi-config.json
```diff
@@ -108,6 +108,42 @@
...
+   					"new": "method void org.jbpm.services.api.model.UserTaskInstanceDesc::setCorrelationKey(java.lang.String)",
+   					"package": "org.jbpm.services.api.model",
+   					"classSimpleName": "UserTaskInstanceDesc",
+   					"methodName": "setCorrelationKey",
+   					"elementKind": "method",
+   					"justification": "https://issues.redhat.com/browse/BAPL-1623"
+				},
+				{
+				   "code": "java.method.addedToInterface",
+  					"new": "method java.lang.String org.jbpm.services.api.model.UserTaskInstanceDesc::getProcessType()",
```
gmunozfe, line 132: Switch return type to java.lang.Integer

elguardian:
This is more or less ok... but where is the integration part ?

fjtirado:
> 
> 
> This is more or less ok... but where is the integration part ?

Do we need to make changes in TaskInstance or just kept them at RuntimeDataService level? I understood the latter.

elguardian:
@fjtirado this is uses by the rest point so the information should be available in there.

elguardian:
https://issues.redhat.com/browse/BAPL-1623?focusedCommentId=14115891&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14115891

PR #1686, opened 2020-06-24
[BAQE-1454] Stabilize flaky test: AsyncMultisubprocessTest

gmunozfe:
Sometimes this test **AsyncMultisubprocessTest** fails, depending on the CPU processing time.

The issue is because 4 iterations (_number_reviews_ variable) for multisubprocess are asynchronously launched, and when two of them (_approvalsRequired_ variable) are completed, the outcome variable switches to true, so processes finishes and uncompleted tasks change to `Exited` status, which makes them fail if they are still trying to transition.

_org.jbpm.services.task.exception.PermissionDeniedException: User '[UserImpl:'rhpamAdmin']' was unable to execute operation 'Claim' on task id 3 due to a no 'current status' match_

I think the easier fix is to force that _approvalsRequired_ are the same as _number_reviews_, so completion condition cannot be achieved while transitioning other human tasks with claim/start/complete actions.


gmunozfe:

PR #1698, opened 2020-07-09
fixed wildfly18 left-overs

mbiarnes:
there were some files still pointing to wildfly18 - or some files with wildlfy18 in its name -
this corrupted the jbpm-installer

PR #1700, opened 2020-07-13
upgraded to JEAP 7.3.0

mbiarnes:


mareknovotny:
this is a followup of #1698

PR #1680, opened 2020-06-17
[JBPM-9195] Case traceability from a Task Instance

fjtirado:
CorrelationKey and ProcessType added to UserTaskDescription.

Depends on [droolsjbpm-knowledge:447](https://github.com/kiegroup/droolsjbpm-knowledge/pull/447) and merge together with https://github.com/kiegroup/droolsjbpm-integration/pull/2136

Existing APIs(getTaskById, getTaskByWorkItem and getTaskByProcess) are
reused, since I believe is preferable to assume the performance cost of
an extra join rather than having to maintain a new API for every
attribute combination (specially when there is already a join because of
the introduction of subject attribute)

PR #1702, opened 2020-07-20
[JBPM-9248] RuntimeDataServiceCDIImplTest must abort process before deleting tables at after

gmunozfe:


jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java
```diff
@@ -1038,7 +1038,6 @@ public void testGetTaskByIdWithCorrelation() {
         assertNotNull(userTask);
         assertCorrelationAndProcess(userTask, correlationKey);
         assertEquals(processInstanceId, userTask.getProcessInstanceId());
-        processService.abortProcessInstance(processInstanceId);
```
afalhambra, line 1041: not sure why do we need to remove this line? Don't we have to abort instances before removing data from the tables?
Also, same method will be called in the `@after` method cleanup once this test is finished.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java
```diff
@@ -1152,6 +1151,7 @@ public void testGetTaskAssignedAsBusinessAdminByStatus() {
         List<TaskSummary> tasks = runtimeDataService.getTasksAssignedAsBusinessAdministratorByStatus("Administrator", statuses, new QueryFilter(0, 5));
         assertNotNull(tasks);
         assertEquals(5, tasks.size());
+        assertTasks(tasks, 5, pids);
```
afalhambra, line 1154: this method is checking for the correlationKey rather than pids. Probably in this case correlationKey = pid always? would it not make sense to modify `assertTasks` method to fetch pids instead of correlationKey? all the calls to this method are passing pids. What do you think?

jbpm-services/jbpm-services-cdi/src/test/java/org/jbpm/services/cdi/test/RuntimeDataServiceCDIImplTest.java
```diff
@@ -173,7 +173,9 @@ public void setIdentityProvider(TestIdentityProvider identityProvider) {
 	
 	@After
 	public void removeAllData() {
-		int deleted = 0;
+        //neeed to abort all processes before deleting all the tables
```
afalhambra, line 176: Small typo

```suggestion
        // need to abort all processes before deleting all the tables
```

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java
gmunozfe: In the after method, it will be aborted as other tests, so it's not strictly needed. Before this fixing, it was necessary for CDI tests to avoid failing and it was the main clue for understanding what was happening.

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/RuntimeDataServiceImplTest.java
gmunozfe: the goal for this assertTasks method is to verify that correlationKey and processType have the expected value (_pids_ and _process_ respectively), as it was introduced in the [jbpm#1680](https://github.com/kiegroup/jbpm/pull/1680)

jbpm-services/jbpm-services-cdi/src/test/java/org/jbpm/services/cdi/test/RuntimeDataServiceCDIImplTest.java
gmunozfe: done

gmunozfe:
jenkins retest this please

fjtirado (approved):

gmunozfe:

PR #1703, opened 2020-07-21
[JBPM-9249] Process upgrade fails in RHPAM 7.7 if process instance ha…

abhijithumbe:
…s active timer and its migrated from BPMS 6.4 release

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
@@ -601,7 +601,7 @@ public class MigrationManager {
                                 List<TimerInstance> collected = new ArrayList<>();
                                 for (Long timerId : timers) {
                                     TimerInstance timerInstance = timerManager.getTimerMap().get(timerId);
-                                    if(timerInstance==null){
+                                    if (timerInstance==null) {
                                         report.addEntry(Type.WARN, "Could not find any timeer instance to cancel.");
                                     	break;
                                     }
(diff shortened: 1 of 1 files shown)


gmunozfe:
jenkins retest this please

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
```diff
@@ -601,7 +601,10 @@ protected TimerManager getTimerManager(KieSession ksession) {
                                 List<TimerInstance> collected = new ArrayList<>();
                                 for (Long timerId : timers) {
                                     TimerInstance timerInstance = timerManager.getTimerMap().get(timerId);
-
+                                    if(timerInstance==null){
```
gmunozfe, line 604: Format: add blanks around the parenthesis:
```suggestion
                                    if (timerInstance==null) {
```

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/migration/MigrationManager.java
```diff
@@ -601,7 +601,10 @@ protected TimerManager getTimerManager(KieSession ksession) {
...
+                                        report.addEntry(Type.WARN, "Could not find any timeer instance to cancel.");
```
gmunozfe, line 605: TimerId may be added also to the report (fix a typo at timer also):
```suggestion
                                        report.addEntry(Type.WARN, "Could not find timer instance with id "+timerId+" to cancel.");
```

abhijithumbe:
@gmunozfe this issue occurs only when process instance is migrated from BPMS 6.4 release so its bit difficult to cover this scenario in a test. Any thoughts about how can we cover this scenario in test

gmunozfe:
jenkins retest this please

elguardian:
The root cause was fixed here. test. https://issues.redhat.com/browse/JBPM-8896
I don't think this is on 6.5.x branch. Already a test in there so there is no need for any additional test cases.
This patch will cover this situation. This is ok to go. @gmunozfe

PR #1705, opened 2020-07-22
ProcessInstanceInfo set start date when constructed

hmatt1:
This is my first PR against jbpm :smile: Please let me know if I should open a Jira ticket [here](https://issues.redhat.com/projects/JBPM/issues/JBPM-9137?filter=allopenissues) for this fix.

----

In the `PrometheusProcessEventListener`, the field `startDate` from a `processInstance` is used to time the process instance duration metric.
https://github.com/kiegroup/droolsjbpm-integration/blob/4fb9e4c239dfc21ea99ea5b701877cafeafcbe1d/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java#L96

However, the start date was always null because it is being set when `getProcessInstance` is first called.
Since `getProcessInstance` is not called before we trigger the event listener, the start date was never set.
The `ProcessInstance` start date should be set when the `ProcessInstanceInfo` object is first created.

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
@@ -103,10 +103,10 @@ public class ProcessInstanceInfo implements PersistentProcessInstance {
         this.processId = processInstance.getProcessId();
 
         if (((WorkflowProcessInstanceImpl) this.processInstance).getStartDate() == null) {
-            startDate = new Date();
+            this.startDate = new Date();
             ((WorkflowProcessInstanceImpl) processInstance).internalSetStartDate(this.startDate);
         } else {
-            startDate = ((WorkflowProcessInstanceImpl) this.processInstance).getStartDate();
+            this.startDate = ((WorkflowProcessInstanceImpl) this.processInstance).getStartDate();
         }
     }
 
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
@@ -55,7 +55,6 @@ import org.jbpm.marshalling.impl.ProcessMarshallerRegistry;
 import org.jbpm.marshalling.impl.ProtobufRuleFlowProcessInstanceMarshaller;
 import org.jbpm.persistence.api.PersistentProcessInstance;
 import org.jbpm.process.instance.impl.ProcessInstanceImpl;
-import org.jbpm.workflow.instance.WorkflowProcessInstance;
 import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
 import org.kie.api.runtime.Environment;
 import org.kie.api.runtime.process.ProcessInstance;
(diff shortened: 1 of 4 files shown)


hmatt1:
ok to test

gmunozfe:
jenkins retest this please

bxf12315:
I think you also can set it on afterProcessStarted https://github.com/kiegroup/droolsjbpm-integration/blob/4fb9e4c239dfc21ea99ea5b701877cafeafcbe1d/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java#L81.

deekshajohari:
> I think you also can set it on afterProcessStarted https://github.com/kiegroup/droolsjbpm-integration/blob/4fb9e4c239dfc21ea99ea5b701877cafeafcbe1d/kie-server-parent/kie-server-services/kie-server-services-prometheus/src/main/java/org/kie/server/services/prometheus/PrometheusProcessEventListener.java#L81.

Yes  this will work. However I still believe that more appropriate way will be to have start date in the object all the time :-)

gmunozfe:
jenkins retest this please

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
```diff
@@ -103,6 +104,16 @@ public ProcessInstanceInfo(ProcessInstance processInstance) {
         this.processInstance = processInstance;
         this.processId = processInstance.getProcessId();
         startDate = new Date();
+
+        // If we are creating a second Process Instance Info for the same process instance,
+        // it should not generate a new start date
+        if (this.processInstance != null) {
```
gmunozfe, line 110: You can remove this check, as it will always be true (notice that a null pointer exception will be thrown 2 lines above if processInstance were null)

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
```diff
@@ -103,6 +104,16 @@ public ProcessInstanceInfo(ProcessInstance processInstance) {
         this.processInstance = processInstance;
         this.processId = processInstance.getProcessId();
         startDate = new Date();
```
gmunozfe, line 106: You can move this line inside the first condition, for clarity and not overwriting when getStartDate is not null

hmatt1:
Thank you @gmunozfe @deekshajohari and @bxf12315 for the review comments! I updated the changes based on @gmunozfe's comments. I'm happy to refactor again if we want to go with what @deekshajohari suggested, to move initializing the `startDate` to the `ProcessInstance` instead if you feel we should go that route. That would be a slight change in behavior, since the `startDate` is currently initialized within the `ProcessInstanceInfo`.

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/ProcessInstanceInfo.java
```diff
@@ -55,6 +55,7 @@
 import org.jbpm.marshalling.impl.ProtobufRuleFlowProcessInstanceMarshaller;
 import org.jbpm.persistence.api.PersistentProcessInstance;
 import org.jbpm.process.instance.impl.ProcessInstanceImpl;
+import org.jbpm.workflow.instance.WorkflowProcessInstance;
```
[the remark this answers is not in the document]
hmatt1: I found that `import org.drools.persistence.api.Transformable;` was unused as well and removed it

hmatt1:
> @hmatt1 thanks a lot for your contribution!
> Some comments regarding some stuff that can be removed or moved to a better place, but looks good.
> Could you add also a test to check startDates are the same when creating a second ProcessInfo for the same process instance?

I removed [these comments](https://github.com/kiegroup/jbpm/pull/1705/commits/5971786f7797b13509cb9d4a92050acfb23d59f5#diff-8d19e3c8a29704a1927512353216e0bcR108) since we now have a test for it.

gmunozfe:
jenkins retest this please

hmatt1:
Thanks @elguardian for the review! This approach is much simpler :smile: Thanks for the background about process instance info too

PR #1707, opened 2020-07-22
[BAQE-1444] - Switch from tcp to mem H2 in jbpm-bpmn2 module

afalhambra:
- Due to this change, there might be some tests failing as records are inserted in the in memory DB much faster (some records in DB might have same date, including milliseconds).

Full detail can be found on: https://issues.redhat.com/browse/BAQE-1444


afalhambra:
> Could you check if it also will work with just 5 ms delay?

Done

afalhambra:
@MarianMacik do you mind have a review of this PR? thanks!

elguardian:
@gmunozfe @afalhambra If there is a problem how correlation keys are generating adding a dealy is only going to hide the problem and not fixing it. I think we should address the problem gonzalo pointed out

https://github.com/kiegroup/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/SubProcessNodeInstance.java#L214

And using another way to generate a unique id instead time millis which is pretty weak.

wdyt ?

afalhambra:
> @gmunozfe @afalhambra If there is a problem how correlation keys are generating adding a dealy is only going to hide the problem and not fixing it. I think we should address the problem gonzalo pointed out
> 
> https://github.com/kiegroup/jbpm/blob/master/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/SubProcessNodeInstance.java#L214
> 
> And using another way to generate a unique id instead time millis which is pretty weak.
> 
> wdyt ?

I do agree. Was wondering whether it would be worth that change since I guess it might be very unlikely to happen, as this problem may only arise when parallel subprocesses are created in the same milisecond. Another thing for this change, is that it will be almost impossible to test.
But again, I'm up for that change.

afalhambra:
Just modified correlationKey to be created using the following:

- Custom correlation key + parent process id + UUID

Though now, UUID generates a 36 characters length field, instead of 13 already when using current system millis. Problem with this approach is that now this name is longer. Possible problem out of this is when the following combination is taken into account:

- Process chaining calls + setting up `org.jbpm.correlationkey.length` to a low value. For instance a very long correlation key name in a chain process, restricted to a limit, might end up colliding with an existing process. Though this scenario could arise already, when setting this limit length to a very low value and having a long key correlation name defined.

Take a look at below test for a better understanding of this issue: 
org.jbpm.bpmn2.ActivityTest#testCallActivityChain

afalhambra:
@MarianMacik, @elguardian would you please have a look? and let me know what you think? thanks.

elguardian:
Jenkins restest this please.

elguardian:
Jenkins retest this please.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/SubProcessNodeInstance.java
```diff
@@ -211,7 +212,7 @@ public void internalTrigger(final NodeInstance from, String type) {
                 List<String> businessKeys = new ArrayList<String>();
```
MarianMacik, line 212: Comment on the line 211 no longer holds (timestamp -> UUID).

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/DynamicUtils.java
```diff
@@ -275,7 +276,7 @@ private static long executeSubProcess(StatefulKnowledgeSessionImpl ksession,
                 List<String> businessKeys = new ArrayList<>();
                 businessKeys.add(((WorkflowProcessInstanceImpl) processInstance).getCorrelationKey());
                 businessKeys.add(processId);
-                businessKeys.add(String.valueOf(System.currentTimeMillis()));
+                businessKeys.add(UUID.randomUUID().toString());
```
MarianMacik, line 279: So I guess issue was in case there were 2 subprocesses created in a succession, for example with multi-instance/loop subprocess, right? And at least 2 of them had the same timestamp... That's really quick!

afalhambra: Correct. Two parallel subprocesses from same parent created in the same millisecond... specially now having H2 pure in memory did the trick.

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/SubProcessNodeInstance.java
afalhambra: done

afalhambra:
@kiegroup/gatekeepers please merge

PR #1712, opened 2020-07-29
[JBPM-9249] Process upgrade fails in RHPAM 7.7 if process instance has active timer and its migrated from BPMS 6.4 release

elguardian:


elguardian:
depends on https://github.com/kiegroup/jbpm/pull/1713

PR #1651, opened 2020-05-15
[RHPAM-2591] Task identifers are recorded on the DB with negative numbers

elguardian:
changed oracle scripts to reflect properly the increment by = allocation size in entities.

PR #1714, opened 2020-07-29
[JBPM-9259] Create a test to validate new springboot DDL scripts

gmunozfe:
With [jbpm#1651](https://github.com/kiegroup/jbpm/pull/1651) new DDL scripts (named with prefix _springboot_) were added for postgresql and oracle, generated with the property _**hibernate.id.new_generator_mappings**_ set to _true_.

These scripts have to be validated against definition in order to keep them always updated.

### DIFF
--- a/jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/util/TestsUtil.java
+++ b/jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/util/TestsUtil.java
@@ -17,11 +17,11 @@
 package org.jbpm.test.persistence.scripts.util;
 
 import java.io.File;
-import java.io.FilenameFilter;
 import java.util.Arrays;
 import java.util.Comparator;
 import java.util.Properties;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.jbpm.test.persistence.scripts.DatabaseType;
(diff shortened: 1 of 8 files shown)


jbpm-test-util/src/main/java/org/jbpm/test/persistence/scripts/util/TestsUtil.java
```diff
@@ -41,35 +41,36 @@
...
-        } else {
+        
+        if (!folderWithScripts.exists()) {
+            logger.warn("Folder with DDLs doesn't exist {}", folderWithDDLs);
             return new File[0];
         }
+        
+        Predicate<File> filterExtension = (name) -> name.getName().toLowerCase().endsWith(".sql");
+        Predicate<File> filterSpringboot = (name) -> name.getName().toLowerCase().contains("springboot");
+        Predicate<File> filterName = (name) -> name.getName().contains("drop");
```
MarianMacik, line 59: Just really minor thing that this parameter is actually a file and not a name, but nothing major :)

gmunozfe: Thanks for your review, updated.

elguardian:
LGTM

PR #1724, opened 2020-08-07
[JBPM-9278] Tests for Email deadline notifications with DBUserInfoImpl

gmunozfe:
Changed also the base test to make it more clear: the email defined in the mvel rule is not needed to be populated before nor at userinfo.properties (for DefaultUserInfoImpl) neither at database (for DBUserInfoImpl)

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.services.task;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.util.Properties;
+import javax.persistence.EntityManagerFactory;
+import javax.persistence.Persistence;
+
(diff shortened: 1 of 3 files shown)


gmunozfe:
@elguardian @afalhambra could you review?

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java
```diff
@@ -0,0 +1,134 @@
...
+            st.execute();
+
+            st.close();
+            conn.close();
+
+        } catch (Exception e) {
+            e.printStackTrace();
+
+        } finally {
+            pds.close();
```
afalhambra, line 129: I would leave `pds` to be closed outside this method since it's being already closed at `clean` method. 

https://github.com/kiegroup/jbpm/blob/26921f4d212e561453888444861987322e453264/jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java#L106

I would though move both `st` and `conn` variables closure to this `finally` block cause at the moment in case of any exception these won't be closed.

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+        }
+    }
+
+    @After
+    public void clean() {
+        TaskDeadlinesServiceImpl.reset();
+        super.tearDown();
+        cleanDb();
+        if (pds != null) {
```
afalhambra, line 105: `emf` should also be closed

afalhambra (changes requested):
Just a few comments, nothing major, otherwise looks great!

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java
gmunozfe: done

jbpm-human-task/jbpm-human-task-core/src/test/java/org/jbpm/services/task/EmailDeadlinesDBTest.java
gmunozfe: I've refactored to use try-with-resources for taking advantage of autocloseable and creating auxiliary methods for enhancing readability. Thanks for your review @afalhambra !

afalhambra (approved):
Nice job @gmunozfe!

gmunozfe:

PR #1732, opened 2020-08-19
[JBPM-9298] Serialized session id is not written to .ser file in windows

fjtirado:
Here we are facing an OS related issue. Windows does not support ':' as part of 
file name, so file writing always fails.
In order to fix it, ':' is replaced by '_' if the file system use a
separator char different than '/'
I prefer that approach rather than using os.name property, which is error
prompt, check for example this thread
https://stackoverflow.com/questions/31909107/javas-os-name-for-windows-10/31909887#31909887

**JIRA**: 

[JBPM-9298](https://issues.redhat.com/browse/JBPM-9298)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/SingletonRuntimeManager.java
```diff
@@ -237,14 +237,19 @@ public void close() {
         this.singleton = null;   
     }
     
+    private File getPersistedFile(String location, String identifier) {
+        // if windows, replace ':' by '_'
+        return new File(location + File.separatorChar +
+                        (File.separatorChar == '/' ? identifier : identifier.replace(':', '_')) + "-jbpmSessionId.ser");
```
elguardian, line 243: we have  commons lang3 in kie-server distribution. I would feel more comfortable using something like

https://commons.apache.org/proper/commons-lang/javadocs/api-3.4/src-html/org/apache/commons/lang3/SystemUtils.html#line.1273

wdyt ?

fjtirado: yes, that seems safer. Thanks for pointing out that commons-lang was available.

elguardian (approved):
LGTM

PR #1740, opened 2020-08-26
[JBPM-9306] Raising exception when user is not defined

fjtirado:
Currently If user does not exist in callback, the organizational entity is not created, but no exception is thrown, so later a constraint violation is thrown, as described in the JIRA.

This change will make error message more descriptive  (the operation itself cannot suceed if the user is not in the context, we cannot allow arbitrary users to be added to organizationentity)

**JIRA**:

[JBPM-9306](https://issues.redhat.com/browse/JBPM-9306)

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/commands/UserGroupCallbackTaskCommand.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/commands/UserGroupCallbackTaskCommand.java
@@ -413,7 +413,6 @@ public class UserGroupCallbackTaskCommand<T> extends TaskCommand<T> {
             if (escalations != null) {
                 for (Escalation escalation : escalations) {
                     List<? extends Notification> notifications = escalation.getNotifications();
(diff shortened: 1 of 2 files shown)


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/commands/UserGroupCallbackTaskCommand.java
```diff
@@ -441,13 +440,14 @@ private void doCallbackDeadlines(List<? extends Deadline> deadlines, TaskContext
                             }
                         }
                     }
+                    List<? extends Reassignment> ressignments = escalation.getReassignments();
```
afalhambra, line 443: typo

```suggestion
                    List<? extends Reassignment> reassignments = escalation.getReassignments();
```

afalhambra (changes requested):
Just one typo.

I would also add a test to cover this scenario, where an exception is expected.

elguardian (approved):
LGTM

afalhambra (approved):
Nice one. Good job

fjtirado:
@kiegroup/gatekeepers please merge this

PR #1750, opened 2020-09-07
[JBPM-9261] unclear validation message for link in different subprocess

fjtirado:
Actually, the way engine works, links are included as metadata
information of the parent process or every container node, so there is
not vissibility between the links of a subprocess and links of the
parent process, meaning that, at all effects, a no link scenario is
similar to a link in different subprocess.

Error Message was already improved as part of RPHAM-3132, so message
when there is no association is

"There is not connection from any throw link to these catch links {list
of ids}" or "There is not connection from any catch link to these throw
links {list of ids}"

Since now the ids of the missing connections are logged, although the
same message is printed regardless there is no link or there is a link
in a different process,  the user can easily tie knots, avoiding the
confusion.

A test case was added to verify message is printed

**JIRA**:

[JBPM-9261](https://issues.redhat.com/browse/JBPM-9261)

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
@@ -86,10 +86,12 @@ import org.jbpm.workflow.core.node.Trigger;
 import org.jbpm.workflow.core.node.WorkItemNode;
 import org.kie.api.definition.process.Node;
 import org.kie.api.definition.process.NodeContainer;
+import org.kie.api.definition.process.Process;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.xml.sax.Attributes;
 import org.xml.sax.SAXException;
+
 
 public class ProcessHandler extends BaseAbstractHandler implements Handler {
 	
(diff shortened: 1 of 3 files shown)


elguardian (approved):
LGTM

gmunozfe (approved):
Looks good to me, great work @fjtirado !

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
```diff
@@ -258,28 +260,34 @@ else if (link.isThrowLink()) {
...
         if (errors.length() > 0) {
             throw new IllegalArgumentException (errors.toString());
         }
         
     }
     
-    private static void formatError (StringBuilder errors, String message, Stream<IntermediateLink> stream) {
+    private static void formatError (StringBuilder errors, String message, Stream<IntermediateLink> stream, NodeContainer container) {
         errors.append(message).append(stream.map(IntermediateLink::getUniqueId).collect(Collectors.joining(", ","{","}")));
+        if (container instanceof Process) {
```
afalhambra, line 285: Maybe we can use `org.kie.api.definition.process.Process` instead, just to make it a little more abstract? just like the one used below which is an instance of `org.kie.api.definition.process.Node`? - wdyt?

afalhambra (commented):
Looks very good @fjtirado. Just only one comment, but up to you to change, not an issue.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/ProcessHandler.java
fjtirado: you are right, changed

afalhambra (approved):
Nicely done @fjtirado

fjtirado:
@kiegroup/gatekeepers please merge this

PR #1752, opened 2020-09-11
[JBPM-9342] disposeRuntimeEngine shouldn't throw exception if already closed

gmunozfe:
**JIRA**: 

[JBPM-9342](https://issues.redhat.com/browse/JBPM-9342)

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerRequestRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerRequestRuntimeManager.java
@@ -37,6 +37,8 @@ import org.kie.internal.runtime.manager.TaskServiceFactory;
 import org.kie.internal.runtime.manager.context.EmptyContext;
 import org.kie.internal.task.api.ContentMarshallerContext;
 import org.kie.internal.task.api.InternalTaskService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * A RuntimeManager implementation that is backed by the "Per Request" strategy. This means that for every call to 
@@ -50,6 +52,8 @@ import org.kie.internal.task.api.InternalTaskService;
  *
  */
 public class PerRequestRuntimeManager extends AbstractRuntimeManager {
+
+    private static final Logger logger = LoggerFactory.getLogger(PerRequestRuntimeManager.class);
 
     private SessionFactory factory;
     private TaskServiceFactory taskServiceFactory;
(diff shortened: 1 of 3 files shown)


elguardian (approved):
fair enough. LGTM

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerRequestRuntimeManager.java
```diff
@@ -131,7 +131,8 @@ public void validate(KieSession ksession, Context<?> context) throws IllegalStat
     @Override
     public void disposeRuntimeEngine(RuntimeEngine runtime) {
     	if (isClosed()) {
-    		throw new IllegalStateException("Runtime manager " + identifier + " is already closed");
+            logger.warn("Runtime manager {} is already closed", identifier);
```
afalhambra, line 134: A new `logger` member should be created in this class belonging to the same - using `logger` from AbstractRuntimeManager might lead to confusion in logs, and might not be displayed in case a different logger level is set for that class.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -87,7 +87,7 @@
  */
 public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
     
-    private static final Logger logger = LoggerFactory.getLogger(AbstractRuntimeManager.class);
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractRuntimeManager.class);
```
afalhambra, line 90: I think this should be kept as it was - `logger` for PerRequestRuntimeManager should be created in its own class and not inherit this to avoid any issues when configuring logging.

afalhambra (changes requested):
Only one change requested with regards to the logger, other than that it looks very good to me!

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerRequestRuntimeManager.java
gmunozfe: Done, thanks for your review @afalhambra !

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
gmunozfe: Kept it back to the previous qualifier, thanks

afalhambra (approved):
Sweet one @gmunozfe!

gmunozfe:

mareknovotny:
Jenkins do fdb

PR #1753, opened 2020-09-14
[JBPM-9344] Maven project loader requires solving pom with immutable spring boot images.

elguardian:
If the project is classpath there is no need to resolve artifacts. They should be already in the class loader.
https://issues.redhat.com/browse/JBPM-9344


elguardian:
depends on 
https://github.com/kiegroup/kie-soup/pull/155

afalhambra (approved):
Nice job @elguardian!

PR #1754, opened 2020-09-14
[JBPM-9341] - Detached WorkItemInfo entities error when aborting cases containing active/running subcases

afalhambra:
**[JBPM-9341](https://issues.redhat.com/browse/JBPM-9341)**

### DIFF
--- a/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/NotifyParentCaseEventListener.java
+++ b/jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/NotifyParentCaseEventListener.java
@@ -142,7 +142,10 @@ public class NotifyParentCaseEventListener extends DefaultProcessEventListener i
             
             Map<String, Object> results = new HashMap<>(caseFileInstance.getData());
             results.put("CaseId", event.getCaseId());
-            processService.completeWorkItem(caseFileInstance.getParentWorkItemId(), results);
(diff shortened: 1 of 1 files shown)


afalhambra:
Jenkins execute FDB

krisv:
@afalhambra I like the approach of the other PR more, as simply checking whether the parent process instance is aborted might work in this case, but there are many other cases that could lead to a subcase node to be cancelled (e.g. if the subcase is in a sub-process that is being aborted) and it would be difficult to detect all of these.

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/main/java/org/jbpm/casemgmt/impl/wih/NotifyParentCaseEventListener.java
```diff
@@ -142,7 +142,10 @@ protected void notifyParentOnCompletion(CaseEvent event) {
             
             Map<String, Object> results = new HashMap<>(caseFileInstance.getData());
             results.put("CaseId", event.getCaseId());
-            processService.completeWorkItem(caseFileInstance.getParentWorkItemId(), results);
+            org.kie.api.runtime.process.ProcessInstance parentProcessInstance = processService.getProcessInstance(caseFileInstance.getParentInstanceId());
+            if (parentProcessInstance.getState() != ProcessInstance.STATE_ABORTED) {
+                processService.completeWorkItem(caseFileInstance.getParentWorkItemId(), results);
+            }
```
elguardian, line 148: this change is not backward compatible.

afalhambra:
Closing this one, as preferred approach is https://github.com/kiegroup/drools/pull/3092

PR #1761, opened 2020-09-25
RHPAM-3188: Exception after dynamically adding new node to process

krisv:


elguardian:
yep. This would work but the problem is that the dumper is not dumping all the data. For instance the process provided would not work.

elguardian:
[~krisv] added the default to the dumper. Let see how CI works.

elguardian (approved):
LGTM.

PR #1780, opened 2020-10-14
[JBPM-9427] IllegalArgumentException when task is skipped in afterTas…

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9427


gmunozfe (approved):
It works fine with the reproducer in wildfly/EAP.
Good work @elguardian !

PR #1782, opened 2020-10-18
[JBPM-9434] REST endpoint authetication fails if Authorization header…

abhijithumbe:
… has == in encoded key in REST WIH

**JIRA**: _(https://issues.redhat.com/browse/JBPM-9434)_

### DIFF
--- a/jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
+++ b/jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
@@ -111,7 +111,7 @@ import org.apache.http.conn.routing.HttpRoutePlanner;
  * <li>ResultClass - fully qualified class name of the class that response should be transformed to,
  * if not given string format will be returned</li>
  * <li>AcceptHeader - accept header value</li>
- * <li>Headers - additional HTTP Headers in format: header1:val1;header2:val2</li>
+ * <li>Headers - additional HTTP Headers in format: header1=val1;header2=val2</li>
  * </ul>
  */
 
@@ -905,7 +905,7 @@ public class RESTWorkItemHandler extends AbstractLogOrThrowWorkItemHandler imple
     protected void addCustomHeaders(String headers,
             BiConsumer<String, String> consumer) {
         for(String h : headers.split(";")) {
-             String[] headerParts = h.split(":");
+             String[] headerParts = h.split("=",2);
              if(headerParts.length == 2) {
                  consumer.accept(headerParts[0], headerParts[1]);
              } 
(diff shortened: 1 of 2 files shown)


jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
```diff
@@ -905,7 +905,7 @@ protected void addAcceptHeader(HttpRequestBase theMethod,
     protected void addCustomHeaders(String headers,
             BiConsumer<String, String> consumer) {
         for(String h : headers.split(";")) {
-             String[] headerParts = h.split("=");
+             String[] headerParts = h.split(":");
```
gmunozfe, line 908: We can keep previous separator, but forcing the length of the resulting array, just to split with the first separator:
```suggestion
             String[] headerParts = h.split("=", 2);
```

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/RestWorkItemHandlerTest.java
```diff
@@ -1069,7 +1069,29 @@ public void testHeadersSingleValue() {
...
+        workItem.setParameter("Method",
+                              "GET");
+        workItem.setParameter(PARAM_HEADERS, headers);
+        
+        WorkItemManager manager = new TestWorkItemManager();
+        handler.executeWorkItem(workItem,
+                                manager);
+        Map<String, Object> results = ((TestWorkItemManager) manager).getResults(workItem.getId());
+        String result = (String) results.get(PARAM_RESULT);
+        assertEquals(result, headerValue);
```
gmunozfe, line 1086: It should be the other way around (expected, current):
```suggestion
        assertEquals(headerValue, result);
```

gmunozfe (changes requested):
hi @abhijithumbe, thanks for pointing this out.
If the issue is because the separator appears more than one time, and only the first one is relevant, I think we could use the `split("=", 2) ` to get just 2 elements as expected. Therefore, we don't break backward compatibility, by using the same separator, wdyt?
You can keep the test that you have added, just one comment there

jbpm-workitems/jbpm-workitems-rest/src/main/java/org/jbpm/process/workitem/rest/RESTWorkItemHandler.java
abhijithumbe: @gmunozfe Thank you for pointing it out. I should have thought about backward compatibility.

jbpm-workitems/jbpm-workitems-rest/src/test/java/org/jbpm/process/workitem/rest/RestWorkItemHandlerTest.java
abhijithumbe: @gmunozfe Made suggested changes in this test case and in other test cases as well.

gmunozfe (approved):
Looks good to me, well done @abhijithumbe !

abhijithumbe:
hi @kiegroup/gatekeepers, could you merge?

gmunozfe:
@kiegroup/gatekeepers could you merge please?

mareknovotny:
jenkins retest this please

abhijithumbe:
@mareknovotny can this PR be merged or need any modification ?

PR #1786, opened 2020-10-26
[JBPM-9447] No session found for context in case SLA on user task is violated

elguardian:
Jira:
https://issues.redhat.com/browse/JBPM-9447


sutaakar (approved):
Looks good. Works as expected.
No exceptions are thrown, WebLogic tests pass.

PR #1794, opened 2020-11-04
[JBPM-4957] Kafka Event Emitter should reuse the same Producer

fjtirado:
**JIRA**

[JBPM-9457](https://issues.redhat.com/browse/JBPM-9457)

Using one producer to publish all kafka events. 
Since the events are published at the end of transaction there is no chance for the appearance of "ghost events"
The main caveat againts the previous implementation is that there exist a minimal chance (if Kafka broker goes down or out of reach during publishing of events) that not all events associate to the transaction are published if Kafka is avaialble (normally there is only one event associated to the transaction, so 99% of the cases this error scenario is equivalent to the transacted one, where the options were none vs all published).

### DIFF
--- a/jbpm-event-emitters/jbpm-event-emitters-kafka/src/main/java/org/jbpm/event/emitters/kafka/KafkaEventEmitter.java
+++ b/jbpm-event-emitters/jbpm-event-emitters-kafka/src/main/java/org/jbpm/event/emitters/kafka/KafkaEventEmitter.java
@@ -15,12 +15,10 @@
  */
 package org.jbpm.event.emitters.kafka;
 
-import java.io.IOException;
 import java.text.SimpleDateFormat;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
-import java.util.UUID;
 
 import com.fasterxml.jackson.databind.MapperFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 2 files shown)


gmunozfe (approved):
Looks good to me, wonderful work @fjtirado !

jbpm-event-emitters/jbpm-event-emitters-kafka/src/main/java/org/jbpm/event/emitters/kafka/KafkaEventEmitter.java
```diff
@@ -110,55 +112,40 @@ public void deliver(Collection<InstanceView<?>> data) {
                 producer.send(new ProducerRecord<>(getTopic(topic), mapper.writeValueAsBytes(
```
elguardian, line 112: you need to add a callback here to know if the message was delivered or not.

https://kafka.apache.org/10/javadoc/org/apache/kafka/clients/producer/KafkaProducer.html#send-org.apache.kafka.clients.producer.ProducerRecord-org.apache.kafka.clients.producer.Callback-

in case the callback is a exception the data needs to be logged entirely

elguardian (changes requested):

elguardian:
overall looks ok. just need a log in case the delivery fails.

jbpm-event-emitters/jbpm-event-emitters-kafka/src/main/java/org/jbpm/event/emitters/kafka/KafkaEventEmitter.java
fjtirado: Thanks for pointing it out, done

elguardian (approved):

PR #1805, opened 2020-11-16
JBPM-9469 - Task inputs are not updated in TaskVariableImpl table aft…

MarianMacik:
…er the task is updated

[JBPM-9469](https://issues.redhat.com/browse/JBPM-9469)

Currently supports addition of the new inputs. If the input is overwritten, it adds another record to the table currently. For deletion of the old/updated inputs it would take more effort as the deletion of the old inputs in TaskVariableImpl was removed to have good performance when adding a new task [1]. In short, at that time it was considered that task inputs are not changeable. But since we already have the `UpdateTaskCommand` available as a public API, this fix at least allows proper addition of the input, so better than not reflecting changes at all I think. But your call guys, I am OK with closing this if we decide we either want all or nothing to be reflected in the TaskVariableImpl table in case the inputs are updated.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1322398

<details>
<summary>
How to retest this PR or trigger a specific build:
</summary>

* <b>a pull request</b> please add comment: <b>Jenkins retest this</b>
 
* <b>a full downstream build</b> please add comment: <b>Jenkins run fdb</b>
  
* <b>a compile downstream build</b> please  add comment: <b>Jenkins run cdb</b>

* <b>a full production downstream build</b> please add comment: <b>Jenkins execute product fdb</b>

* <b>an upstream build</b> please add comment: <b>Jenkins run upstream</b>
</details>

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/UpdateTaskCommand.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/UpdateTaskCommand.java
@@ -116,6 +116,7 @@ public class UpdateTaskCommand extends UserGroupCallbackTaskCommand<Void> {
                 persistenceContext.persistContent(inputContent);
             }
             ((InternalTaskData)task.getTaskData()).setTaskInputVariables(mergedContent);
+            taskEventSupport.fireAfterTaskInputVariablesChanged(task, context, inputs);
         }
         
         if (outputs != null) {
--- a/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskServiceImplTest.java
+++ b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskServiceImplTest.java
@@ -24,6 +24,7 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.drools.compiler.kie.builder.impl.InternalKieModule;
 import org.drools.core.util.StringUtils;
@@ -34,6 +35,9 @@ import org.jbpm.services.api.ProcessInstanceNotFoundException;
 import org.jbpm.services.api.TaskNotFoundException;
 import org.jbpm.services.api.model.DeploymentUnit;
 import org.jbpm.services.api.model.UserTaskInstanceDesc;
+import org.jbpm.services.task.HumanTaskServiceFactory;
+import org.jbpm.services.task.audit.TaskAuditServiceFactory;
+import org.jbpm.services.task.audit.service.TaskAuditService;
 import org.jbpm.services.task.commands.GetTaskCommand;
 import org.jbpm.services.task.exception.PermissionDeniedException;
 import org.junit.After;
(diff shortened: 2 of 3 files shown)


elguardian:
@MarianMacik failure related.

elguardian (changes requested):
null pointer. requires assesment

afalhambra (commented):
Very good job @MarianMacik . Thanks for this! 
Just only two comments:

- Shouldn't we add (or modify) a test to update an input task variable twice? so we make sure it gets updated properly? - wdyt?
- The only other issue that comes to my mind is about any potential performance downgrade when it comes to using the new api for advanced queries (only when updating input task vars). Though... at the same time, performance wise should now be consistent with output task variables use

MarianMacik:
>     * Shouldn't we add (or modify) a test to update an input task variable twice? so we make sure it gets updated properly? - wdyt?
As I said in the description, in case an already-existing input is overwritten (updated twice), there will be another task variable impl record in the table, I guess we may consider enhancing it in a separate PR after we test any possible performance degradation caused by one more delete query (as the same piece of code is used when the task is added for example, that would impact pretty much all operations then).

>     * The only other issue that comes to my mind is about any potential performance downgrade when it comes to using the new api for advanced queries (only when updating input task vars). Though... at the same time, performance wise should now be consistent with output task variables use
My understanding was that it was a bug as newly-added inputs or overwritten inputs weren't even queried as they were not visible in the database. So in that case it was a bug I would say.

MarianMacik:
Jenkins retest this

afalhambra:
> > ```
> > * Shouldn't we add (or modify) a test to update an input task variable twice? so we make sure it gets updated properly? - wdyt?
> > ```
> 
> As I said in the description, in case an already-existing input is overwritten (updated twice), there will be another task variable impl record in the table, I guess we may consider enhancing it in a separate PR after we test any possible performance degradation caused by one more delete query (as the same piece of code is used when the task is added for example, that would impact pretty much all operations then).
> 

Regardless of performance, for this particular PR, I was more thinking about a particular test case for updating the input variable twice and afterwards fetching the input tasks and making sure the value retrieved is the latest one - I believe this scenario is not covered in any of the tests? or am I wrong?

MarianMacik:
So you meant to update the input twice and just check it is overwritten when `userTaskService.getTaskInputContentByTaskId(taskId);` is called? I can do that though it works as inputs are map-based and new inputs are just added there via put. Though as I said, in task variable impl table there will be both the previous and the current value present.

MarianMacik:
@afalhambra It is green now. Do you want me to add another test case? But I personally don't see the benefit of it.

afalhambra (approved):
Sorry @MarianMacik, I didn't notice that even new input task vars were not being taken into account at first place, that's why I was asking. 
Looks very great to me! thanks @MarianMacik!

MarianMacik:
> Sorry @MarianMacik, I didn't notice that even new input task vars were not being taken into account at first place, that's why I was asking.

New are taken into account, so if you add a new input task A with value "a" it will be shown. When you then update the task with the same input A with value "b", it will be added (and not overwritten) in the task variable impl table. So you will be able to query the task based on the new and old value ("a" and "b"). To fix that, we would have to run delete query first, which was there, but was removed because of perf issues - as that query in the listener would have to be run also when the task is created for example. That's why I chose to make a compromise, so you are now able to add new inputs and query tasks based on them. If you, however, overwrite it, the old value is still available, because delete wasn't called. I may try to improve it next sprint if I have time.

afalhambra:
Yes, it's pretty cleared now, thanks! - it might have some impact in the perf (specially when querying), but we can deal with it later on. I will be running some perf test to see.

elguardian:
This makes sense to me as allow the system to be consistent in database. otherwise you would see "magic changes" in the  database without any audit.
We cannot sacrifice correctness over speed. And in this case the command is not that widely use.

elguardian (approved):

PR #1806, opened 2020-11-18
[JBPM-9470] User Task output variable mapping to an object attribute

fjtirado:
Reading expressionLanguage from process
Implementing assigment for mvel
Introduced system property (org.kie.jbpm.bpmn2.useDefinitionLanguage) to specify if expressionLanguage should be read from definitions if language is not specified in Assigment (default value false, since there are working process which have mvel as definition and are using Xpath)

**JIRA**: 

[JBPM-9470](https://issues.redhat.com/browse/JBPM-9470)

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/elements/DataAssociationFactory.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/elements/DataAssociationFactory.java
@@ -17,9 +17,15 @@
 package org.jbpm.bpmn2.xml.elements;
 
 import java.util.ArrayList;
+import java.util.Collection;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
+import org.jbpm.process.builder.dialect.ProcessDialectRegistry;
+import org.jbpm.util.PatternConstants;
 import org.jbpm.workflow.core.node.Assignment;
 import org.jbpm.workflow.core.node.DataAssociation;
 import org.jbpm.workflow.core.node.Transformation;
@@ -30,8 +36,13 @@ import org.w3c.dom.NodeList;
 
 public final class DataAssociationFactory {
     
+    protected static final String LANG_EXPRESSION_ATTR = "language";
+    protected static final String DEFAULT_DIALECT = "XPath";
+    protected static final String USE_DEFINITION_LANGUAGE_PROPERTY = "org.kie.jbpm.bpmn2.useDefinitionLanguage";
     private static final Logger logger = LoggerFactory.getLogger(DataAssociationFactory.class);
     
+    private static Map<String, Pattern> dialectPatterns = buildDialectPatterns(ProcessDialectRegistry.getDialects());
+
     private DataAssociationFactory() {
         // do nothing
     }
@@ -63,7 +74,7 @@ public final class DataAssociationFactory {
                 }
                 case "transformation":
                 {
-                    String lang = subNode.getAttributes().getNamedItem("language").getNodeValue();
+                    String lang = subNode.getAttributes().getNamedItem(LANG_EXPRESSION_ATTR).getNodeValue();
                     String expression = subNode.getTextContent();
                     transformation = new Transformation(lang, expression, source);
                     break;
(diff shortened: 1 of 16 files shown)


mareknovotny:
@MarianMacik but this PR looks like turn out the deadlock for GH actions on jdk 8 vs jdk11 re: https://issues.redhat.com/browse/JBPM-9456

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo-Complete.bpmn2
```diff
@@ -0,0 +1,50 @@
...
+             xmlns="http://www.omg.org/spec/BPMN/20100524/MODEL"
+             xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+             xsi:schemaLocation="http://www.omg.org/spec/BPMN/20100524/MODEL BPMN20.xsd"
+             xmlns:g="http://www.jboss.org/drools/flow/gpd"
+             xmlns:bpmndi="http://www.omg.org/spec/BPMN/20100524/DI"
+             xmlns:dc="http://www.omg.org/spec/DD/20100524/DC"
+             xmlns:di="http://www.omg.org/spec/DD/20100524/DI"
+             xmlns:tns="http://www.jboss.org/drools">
+
+  <itemDefinition id="_instanceMetadataItem" structureRef="org.jbpm.bpmn2.DataTest$Person" />
```
gmunozfe, line 15: This value including "$" is not valid as QName:
_The 'structureRef' attribute is invalid - The value 'org.jbpm.bpmn2.DataTest$Person' is invalid according to its datatype 'http://www.w3.org/2001/XMLSchema:QName' - The '$' character, hexadecimal value 0x24, cannot be included in a name._

which it's shown in the parser as an error:
```
ERROR Node Info: id:_instanceMetadataItem name: 
Parser message: (null: 15, 94): cvc-datatype-valid.1.2.1: 'org.jbpm.bpmn2.DataTest$Person' is not a valid value for 'QName'.
```

though dollar-sign ($) is the way that mvel 2.0 uses to access nested classes

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo-Complete.bpmn2
```diff
@@ -0,0 +1,50 @@
...
+        <inputSet>
+          <dataInputRefs>coId</dataInputRefs>
+        </inputSet>
+        <outputSet>
+        </outputSet>
+      </ioSpecification>
+      <dataInputAssociation>
+        <sourceRef>instanceMetadata</sourceRef>
+        <targetRef>coId</targetRef>
+        <assignment language="http://www.mvel.org/2.0">
```
gmunozfe, line 36: _language_ is not a valid attribute inside _**tAssignment**_ type:
_The 'language' attribute is not allowed._

gmunozfe (commented):
Some issues with the xml validation (dollar sign used in mvel nested classes is not valid as a QName, and language is not an attribute of tAssignment type).
Some mvel expressions may include also question mark to treat nulls (if add tests for them)

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo-Complete.bpmn2
fjtirado: This $ is the name of the inner class, since this is test, I think we can live with it (I will prefer to not make this class available in a different file since it only applies to this test)

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo-Complete.bpmn2
fjtirado: yes, thats surprising, since bpmn specification https://www.omg.org/spec/BPMN/2.0/PDF, in page 223, states:
The Assignment class is used to specify a simple mapping of data elements using a specified Expression language.The default Expression language for all Expressions is specified in the Definitions element, using the expressionLanguage attribute. It can also be overridden on each individual Assignment using the same attribute
Anyway, I remove language attribute from XMLs and, in order to achieve independence from JIRA https://issues.redhat.com/browse/JBPM-9471, try to deduce if the language used in the assigment in the expression is mvel based on the context of from and to, but I left the code that search for language attribute in Assignment (which will always be null right now) so it gets executed when both XSD is fixed and Designed sets this attribute.

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo-Complete.bpmn2
```diff
@@ -0,0 +1,50 @@
...
+          <from>#{instanceMetadata.address.city}</from>
```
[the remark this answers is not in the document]
fjtirado: Good idea, done

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo.bpmn2
```diff
@@ -0,0 +1,50 @@
...
+          <dataInputRefs>coId</dataInputRefs>
+        </inputSet>
+        <outputSet>
+        </outputSet>
+      </ioSpecification>
+      <dataInputAssociation>
+        <sourceRef>instanceMetadata</sourceRef>
+        <targetRef>coId</targetRef>
+        <assignment>
+          <from language="http://www.mvel.org/2.0">#{address.city}</from>
```
gmunozfe, line 37: Could you add a comment indicating that language attribute validation against xsd will fail here (though it shouldn't, according to the standard, as you mention before)?
It's just to keep in mind in the future that this is a known issue to us in the xsd definition.

gmunozfe (approved):
Just a minor comment to indicate in one of the bpmn files that the attribute language is not currently supported by the standard xsd.

jbpm-bpmn2/src/test/resources/BPMN2-DataInputAssociations-Pojo.bpmn2
fjtirado: This is a typo and should not be there.

gmunozfe (approved):

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/dialect/mvel/MVELAssignmentAction.java
```diff
@@ -0,0 +1,102 @@
...
+        if (targetObject == null || notEvalTarget()) {
+            producer.accept(context, nodeInstance, notEvalSrc() ? srcObject : MVELSafeHelper.getEvaluator().eval(
+                    from, srcObject, resolver));
+        } else {
+            resolver.addExtraParameter(srcExpr, srcObject);
+            resolver.addExtraParameter(targetExpr, targetObject);
+            MVELSafeHelper.getEvaluator().eval(ensureLocated(targetExpr, to).concat("=").concat(ensureLocated(srcExpr,
+                    from)), resolver);
+        }
+    }
```
elguardian, line 73: The concept of "." as "this" is does not feel very standard.
The concept of expression [] for arrays is as well very strange without setting the name of the array.

To make things a bit more standard we should remove the logic of ensureLocated as you are trying to infer things. 
We can can set the extra parameters as source/target for instance as a pointer to the object. or even something like 
@source[3].myprop = @target.myprop. This will avoid the problem of having node output express with the same name as target. (Edge case but could happen)

This should leave trail in the audit table so I would suggest to add a test about that to evaluate the source expression
target expression
the call producer

in three different steps.

Please also remove the need of #{} if the UI is no setting the language you have already more than enough check to work that properly.

fjtirado: this keyword introduced to cope with the egde case, if expression has same name than target, this usage is mandatory, otherwise is optional

fjtirado: examples edited to not use #{} if language is specified in expression, otherwise it needs to be specified to identify the expression as mvel

fjtirado: this prefix is now used in list assigment example

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/DataTest.java
```diff
@@ -495,7 +615,109 @@ public void executeWorkItem(WorkItem workItem,
...
+                        assertEquals("Spain", person.getAddress().getCountry());
+                    }
+
+                });
+        ksession.startProcess("process", Collections.singletonMap("instanceMetadata", new Person("Napoleon",
+                new Address("Paris", "France"))));
+    }
+
+    @Test
+    public void testDataOutputAssociationsWithPojoList() throws Exception {
```
afalhambra, line 688: I think it would be good to add a test to prove it works fine when using Maps as Anton's comment in the jira ticket. So we are covering all possible scenarios, wdyt?

afalhambra (changes requested):
Very nice implementation. I just left some comments for your review.
Thanks!

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/elements/DataAssociationFactory.java
```diff
@@ -79,22 +90,75 @@ public static DataAssociation readDataOutputAssociation(Node xmlNode, Map<String
...
+            }
+            while (parentNode != null && !parentNode.getLocalName().equals("Definitions"));
+            if (parentNode != null) {
+                dialect = findDialect(parentNode.getAttributes().getNamedItem("expressionLanguage"));
             }
         }
-        return new Assignment("XPath", from, to);
+        // finally, if still not able to determine language, check if from or to contains a mvel expression 
+        if (dialect == null && PatternConstants.PARAMETER_MATCHER.matcher(from.getTextContent()).matches() ||
+            PatternConstants.PARAMETER_MATCHER.matcher(to.getTextContent()).matches()) {
```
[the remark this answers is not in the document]
fjtirado: Good catch

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/elements/DataAssociationFactory.java
```diff
@@ -79,22 +90,75 @@ public static DataAssociation readDataOutputAssociation(Node xmlNode, Map<String
...
+
+        // there are some working process which declares MVEL in definition but use XPATH, in order
+        // to prevent these files to fail, we check a flag (disable by default) before reading expression 
+        // language from definition
+        if (dialect == null && Boolean.getBoolean(USE_DEFINITION_LANGUAGE_PROPERTY)) {
+            Node parentNode;
+            do {
+                parentNode = node.getParentNode();
+            }
+            while (parentNode != null && !parentNode.getLocalName().equals("Definitions"));
```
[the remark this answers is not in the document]
fjtirado: I do not think there is an infinite loop, parentNode is always evaluted before the termination condition

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/dialect/mvel/MVELAssignmentAction.java
```diff
@@ -0,0 +1,113 @@
...
+        this.to = toMatcher.find() ? toMatcher.group(1) : assignment.getTo();
+        this.src = source;
+        this.target = target;
+        this.srcExpr = sourceExpr;
+        this.targetExpr = targetExpr;
+        this.producer = producer;
+    }
+
+    @Override
+    public void execute(NodeInstance nodeInstance, ProcessContext context) throws Exception {
```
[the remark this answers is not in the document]
fjtirado: good catch

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/DataTest.java
fjtirado: Done

afalhambra (approved):
Awesome work @fjtirado! thanks

elguardian (approved):
Good to go

PR #1807, opened 2020-11-19
[DROOLS-5780] descope ecj

mariofusco:


mareknovotny:
@MarianMacik just fyi the deadlock we were trying to discuss on https://issues.redhat.com/browse/JBPM-9456 is here for jdk11 build not present, :o which looks like if that is related only to jdk8?

PR #1810, opened 2020-11-24
[JBPM-9485] Missing dependency for ibm jdk8 after descope ecj

gmunozfe:
**JIRA**: [JBPM-9485](https://issues.redhat.com/browse/JBPM-9485)


gmunozfe:
@mariofusco if you agree with the fix, I'll also apply it in droolsjbpm-integration tests failing (hacep, camel, taskassigning, etc)

mariofusco (commented):
This should be no longer necessary. We realized that excluding ecj by default from drools-compiler could be the cause of many problem similar to this and decided to add it back (making possible to explicitly exclude it if required). I just implemented this with this commit https://github.com/kiegroup/drools/pull/3262

gmunozfe:
thanks @mariofusco, I will close it then

PR #1815, opened 2020-12-09
[JBPM-9501] New Queries process tasks by variables fail if variable n…

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9501


afalhambra (approved):
Nice job!

Maybe `assertThat` (deprecated) in the test can be replaced by something like `assertEquals` but I'll leave to your opinion.

PR #1824, opened 2020-12-14
DISPOSE_CLEANUP

elguardian:
This certifies that runtime engines are clean up during signal sent


gmunozfe (changes requested):
Missing PerCaseRuntimeManager

gmunozfe (approved):
Looks good to me, nailed down @elguardian !

PR #1826, opened 2020-12-16
[LOCKING_STRATEGY] implementing different strategies for diagnose and…

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9521

strategies and lock diagnose.

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
@@ -15,12 +15,10 @@
  */
 package org.jbpm.runtime.manager.impl;
 
+
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.drools.core.time.TimerService;
 import org.drools.persistence.api.OrderedTransactionSynchronization;
@@ -34,7 +32,9 @@ import org.jbpm.process.core.timer.impl.GlobalTimerService;
 import org.jbpm.runtime.manager.api.SchedulerProvider;
 import org.jbpm.runtime.manager.impl.error.DefaultExecutionErrorStorage;
 import org.jbpm.runtime.manager.impl.error.ExecutionErrorManagerImpl;
+import org.jbpm.runtime.manager.impl.lock.RuntimeManagerLockStrategyFactory;
 import org.jbpm.runtime.manager.impl.tx.NoOpTransactionManager;
+import org.jbpm.runtime.manager.spi.RuntimeManagerLockStrategy;
 import org.jbpm.services.task.impl.TaskContentRegistry;
 import org.jbpm.services.task.impl.command.CommandBasedTaskService;
 import org.jbpm.services.task.wih.ExternalTaskEventListener;
@@ -90,6 +90,8 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
     private static final Logger logger = LoggerFactory.getLogger(AbstractRuntimeManager.class);
 
     protected RuntimeManagerRegistry registry = RuntimeManagerRegistry.get();
+    protected RuntimeManagerLockStrategyFactory lockStrategyFactory = new RuntimeManagerLockStrategyFactory();
+
     protected RuntimeEnvironment environment;
     protected DeploymentDescriptor deploymentDescriptor;
     protected KieContainer kieContainer;
@@ -104,8 +106,8 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
     
     protected SecurityManager securityManager = null;
     protected ExecutionErrorManager executionErrorManager;
-        
-    protected ConcurrentMap<Long, ReentrantLock> engineLocks = new ConcurrentHashMap<Long, ReentrantLock>(); 
+    protected RuntimeManagerLockStrategy runtimeManagerLockStrategy;
+
     
     public AbstractRuntimeManager(RuntimeEnvironment environment, String identifier) {
         this.environment = environment;
@@ -113,6 +115,7 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
         if (registry.isRegistered(identifier)) {
             throw new IllegalStateException("RuntimeManager with id " + identifier + " is already active");
         }
+
         ((SimpleRuntimeEnvironment)environment).getEnvironmentTemplate().set(EnvironmentName.DEPLOYMENT_ID, this.getIdentifier());
         internalSetDeploymentDescriptor();
         internalSetKieContainer();
@@ -129,7 +132,7 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
         ((SimpleRuntimeEnvironment)environment).getEnvironmentTemplate().set(EnvironmentName.EXEC_ERROR_MANAGER, executionErrorManager);
         logger.info("{} is created for {}", this.getClass().getSimpleName(), identifier);
     }
-    
+
     private void internalSetDeploymentDescriptor() {
     	this.deploymentDescriptor = (DeploymentDescriptor) ((SimpleRuntimeEnvironment)environment).getEnvironmentTemplate().get("KieDeploymentDescriptor");
     	if (this.deploymentDescriptor == null) {
@@ -141,7 +144,13 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
     	this.kieContainer = (KieContainer) ((SimpleRuntimeEnvironment)environment).getEnvironmentTemplate().get("KieContainer");
 	}
 
-	public abstract void init();
+	public void init() {
+        if (!isUseLocking()) {
+            runtimeManagerLockStrategy = lockStrategyFactory.createFreeLockStrategy();
+        } else {
+            runtimeManagerLockStrategy = lockStrategyFactory.createLockStrategy(identifier);
+        }
+	}
     
 	protected void registerItems(RuntimeEngine runtime) {
         RegisterableItemsFactory factory = environment.getRegisterableItemsFactory();
@@ -431,49 +440,28 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
     }
     
     protected void createLockOnGetEngine(Long id, RuntimeEngine runtime) {
-        if (!isUseLocking()) {
-            logger.debug("Locking on runtime manager disabled");
+        if (id == null) {
             return;
         }
-        
-        if (id != null) {
-            ReentrantLock newLock = new ReentrantLock();
-            ReentrantLock lock = engineLocks.putIfAbsent(id, newLock);
-            if (lock == null) {
-                lock = newLock;
-                logger.debug("New lock created as it did not exist before");
-            } else {
-                logger.debug("Lock exists with {} waiting threads", lock.getQueueLength());
-            }
-            logger.debug("Trying to get a lock {} for {} by {}", lock, id, runtime);
-            lock.lock();
-            logger.debug("Lock {} taken for {} by {} for waiting threads by {}", lock, id, runtime, lock.hasQueuedThreads());
-            
+        try {
+            runtimeManagerLockStrategy.lock(id, runtime);
+        } catch(InterruptedException e) {
+            throw new RuntimeException("Runtime manager was not able to lock for {" + id +"}", e);
         }
-        
+    }
+
+    protected void releaseAndCleanLock(Long id, RuntimeEngine runtime) {
+        if (id == null) {
+            return;
+        }
+        runtimeManagerLockStrategy.unlock(id, runtime);
     }
     
     protected void createLockOnNewProcessInstance(Long id, RuntimeEngine runtime) {
-        if (!isUseLocking()) {
-            logger.debug("Locking on runtime manager disabled");
-            return;
-        }
-        ReentrantLock newLock = new ReentrantLock();
-        ReentrantLock lock = engineLocks.putIfAbsent(id, newLock);
-        if (lock == null) {
-            lock = newLock;
-        }
-        lock.lock();
-        logger.debug("[on new process instance] Lock {} created and stored in list by {}", lock, runtime);
+        createLockOnGetEngine(id, runtime);
     }
     
-    
     protected void releaseAndCleanLock(RuntimeEngine runtime) {
-        if (!isUseLocking()) {
-            logger.debug("Locking on runtime manager disabled");
-            return;
-        }
-        
         if (((RuntimeEngineImpl)runtime).getContext() instanceof ProcessInstanceIdContext) {
             Long piId = ((ProcessInstanceIdContext) ((RuntimeEngineImpl)runtime).getContext()).getContextId();
             if (piId != null) {
@@ -482,23 +470,7 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
         }
     }
     
-    protected void releaseAndCleanLock(Long id, RuntimeEngine runtime) {
 
-        if (id != null) {
-            ReentrantLock lock = engineLocks.get(id);
-            if (lock != null) {
-                if (!lock.hasQueuedThreads()) {
-                    logger.debug("Removing lock {} from list as non is waiting for it by {}", lock, runtime);
-                    engineLocks.remove(id);
-                }
-                if (lock.isHeldByCurrentThread()) {
-                    lock.unlock();
-                    logger.debug("{} unlocked by {}", lock, runtime);
-                }
-            }
-        }
-        
-    }
     
     protected boolean isActive() {
         if (hasEnvironmentEntry("Active", false)) {
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
@@ -1,0 +1,133 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.runtime.manager.impl.lock;
+
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import org.jbpm.runtime.manager.spi.RuntimeManagerLock;
+import org.jbpm.runtime.manager.spi.RuntimeManagerLockFactory;
+import org.jbpm.runtime.manager.spi.RuntimeManagerLockStrategy;
+import org.kie.api.runtime.manager.RuntimeEngine;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+
+abstract class AbstractRuntimeManagerLockStrategy implements RuntimeManagerLockStrategy {
+
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractRuntimeManagerLockStrategy.class);
+
+    protected Map<Long, RuntimeManagerLockThreadsInfo> engineLocks; 
+
+    private RuntimeManagerLockFactory runtimeManagerLockFactory;
+
+    
+    protected AbstractRuntimeManagerLockStrategy() {
+        this.runtimeManagerLockFactory = new DefaultRuntimeManagerLockFactory();
+        this.engineLocks = new HashMap<>();
+    }
+
+    @Override
+    public void init(RuntimeManagerLockFactory factory) {
+        this.runtimeManagerLockFactory = factory;
+    }
+
+    @Override
+    public void lock(Long id, RuntimeEngine runtime) throws InterruptedException {
+        RuntimeManagerLockThreadsInfo lockThreadsInfo = null;
+        synchronized (engineLocks) {
+            lockThreadsInfo = engineLocks.computeIfAbsent(id, (Long key) -> new RuntimeManagerLockThreadsInfo(runtimeManagerLockFactory.newRuntimeManagerLock()));
+            lockThreadsInfo.set();
+        }
+        logger.debug("Trying to get a lock {} for {} by {}", lockThreadsInfo, id, runtime);
+        try {
+            lock(lockThreadsInfo.getRuntimeManagerLock());
+        } catch(InterruptedException e) {
+            logger.warn("Interrupted lock {}", lockThreadsInfo.getRuntimeManagerLock());
+            throw e;
+        }
+        logger.debug("Lock {} taken for {} by {} for waiting threads by {}", lockThreadsInfo, id, runtime, lockThreadsInfo.count());
+    }
+
+    protected abstract void lock(RuntimeManagerLock lock) throws InterruptedException;
+    
+    @Override
+    public void unlock(Long id, RuntimeEngine runtime) {
+        RuntimeManagerLockThreadsInfo lockThreadsInfo = null;
+        synchronized (engineLocks) {
+            lockThreadsInfo = engineLocks.get(id);
+            if (lockThreadsInfo == null) {
+                logger.warn("[LOCK] lock {} is already removed for {} unlocked by {}", id, lockThreadsInfo, runtime);
+                return;
+            }
+            lockThreadsInfo.unset();
+            if (lockThreadsInfo.count() == 0) {
+                logger.debug("[LOCK] Removing lock for {} for lock  {} from list as non is waiting for it by {}", id, lockThreadsInfo, runtime);
+                engineLocks.remove(id);
+            }
+        }
+
+        if (lockThreadsInfo.isHeldByCurrentThread()) {
+            unlock(lockThreadsInfo.getRuntimeManagerLock());
+            logger.debug("[LOCK] process instance id {} with thread info {} unlocked by {}", id, lockThreadsInfo, runtime);
+        } else {
+            logger.warn("[LOCK] trying to unlock for {} for lock {} no lock held by {}", id, lockThreadsInfo, runtime);
+        }
+    }
+
+    protected abstract void unlock(RuntimeManagerLock lock);
+}
+
+class RuntimeManagerLockThreadsInfo {
+
+    private Set<Long> threadIds;
+    private RuntimeManagerLock runtimeManagerLock;
+
+    public RuntimeManagerLockThreadsInfo(RuntimeManagerLock runtimeManagerLock) {
+        this.threadIds = new HashSet<>();
+        this.runtimeManagerLock = runtimeManagerLock;
+    }
+    public void set() {
+        this.threadIds.add(Thread.currentThread().getId());
+    }
+    public void unset() {
+        this.threadIds.remove(Thread.currentThread().getId());
+    }
+    public int count() {
+        return this.threadIds.size();
+    }
+
+    public void lock() {
+        runtimeManagerLock.lock();
+    }
+    public void unlock() {
+        runtimeManagerLock.unlock();
+    }
+    public boolean hasQueuedThreads() {
+        return runtimeManagerLock.internalLock().hasQueuedThreads();
+    }
+    public boolean isHeldByCurrentThread() {
+        return runtimeManagerLock.internalLock().isHeldByCurrentThread();
+    }
+    
+    public RuntimeManagerLock getRuntimeManagerLock() {
+        return runtimeManagerLock;
+    }
+
+}
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2020 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.runtime.manager.impl.lock;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantLock;
+
+import org.jbpm.runtime.manager.spi.RuntimeManagerLock;
+
+
+public class DebugRuntimeManagerLock implements RuntimeManagerLock {
+
+    private ReentrantLock lock = new ReentrantLock(true);
+    private List<StackTraceElement[]> traces;
+    private long currentThreadId;
+
+    public DebugRuntimeManagerLock() {
+        traces = new ArrayList<>();
+        currentThreadId = -1;
+    }
+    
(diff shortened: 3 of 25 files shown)


elguardian:
Jenkins run fdb

elguardian:
guys I need some eyes in here.

gmunozfe (changes requested):
Check-style plugin failure due to missing header for RuntimeManagerLockStrategyTest.java

elguardian:
Jenkins retest this

elguardian:
Jenkins retest this

elguardian:
Jenkins retest this

elguardian:
Jenkins retest this

elguardian:
Jenkins retest this

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -15,13 +15,15 @@
  */
 package org.jbpm.runtime.manager.impl;
 
+import java.util.Iterator;
```
gmunozfe, line 18: this is not used

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -15,13 +15,15 @@
...
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.locks.ReentrantLock;
```
gmunozfe, line 24: these 3 imports are not used

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -15,13 +15,15 @@
...
 
+import com.sun.tools.javac.util.ServiceLoader;
```
gmunozfe, line 26: Use java util instead:
```suggestion
import java.util.ServiceLoader;
```

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private void addTraces() {
+        long newThreadId = Thread.currentThread().getId();
+        if(currentThreadId < 0 || currentThreadId != newThreadId) {
+            currentThreadId = newThreadId;
+            traces.clear();
+        }
+        // we add all stack traces during reentrant lock
+        traces.add(Thread.currentThread().getStackTrace());
```
MarianMacik, line 66: Can it even happen in the engine that we call lock() more than once in the same thread? By looking at the code when we first get the engine, it is locked and saved to the thread local (saveLocalRuntime method) and then it is reused within the same thread when calling getRuntimeEngine(), so we never reach the line with createLockOnGetEngine(). Just my observation.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -34,7 +36,13 @@
 import org.jbpm.runtime.manager.api.SchedulerProvider;
 import org.jbpm.runtime.manager.impl.error.DefaultExecutionErrorStorage;
 import org.jbpm.runtime.manager.impl.error.ExecutionErrorManagerImpl;
+import org.jbpm.runtime.manager.impl.lock.DefaultRuntimeManagerLockFactory;
+import org.jbpm.runtime.manager.impl.lock.FreeRuntimeManagerLockStrategy;
+import org.jbpm.runtime.manager.impl.lock.LegacyRuntimeManagerLockStrategy;
```
gmunozfe, line 41: These 3 imports are not used

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -34,7 +36,13 @@
...
+import org.jbpm.runtime.manager.impl.lock.RuntimeManagerLockStrategyFactory;
 import org.jbpm.runtime.manager.impl.tx.NoOpTransactionManager;
+import org.jbpm.runtime.manager.spi.RuntimeManagerLockFactory;
```
gmunozfe, line 44: This is not used

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -431,43 +448,31 @@ protected void createLockOnGetEngine(Context<?> context, RuntimeEngine runtime)
...
+
+    protected void releaseAndCleanLock(Long id, RuntimeEngine runtime) {
+        if (id == null) {
+            return;
+        }
+        runtimeManagerLockStrategy.unlock(id, runtime);
     }
     
     protected void createLockOnNewProcessInstance(Long id, RuntimeEngine runtime) {
         if (!isUseLocking()) {
```
gmunozfe, line 469: Check first if id is null (same as rest of methods)

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+public abstract class AbstractRuntimeManagerLockStrategy implements RuntimeManagerLockStrategy {
+
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractRuntimeManagerLockStrategy.class);
+
+    protected Map<Long, RuntimeManagerLockThreadsInfo> engineLocks; 
+
+    private RuntimeManagerLockFactory runtimeManagerLockFactory;
+
+    
+    public AbstractRuntimeManagerLockStrategy() {
```
gmunozfe, line 41: As it is abstract, it shouldn't be public
```suggestion
    protected AbstractRuntimeManagerLockStrategy() {
```

MarianMacik (commented):
Few remarks, but otherwise looks good and understandable.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+    public void init(RuntimeManagerLockFactory factory) {
+        this.runtimeManagerLockFactory = factory;
+    }
+
+    @Override
+    public void lock(Long id, RuntimeEngine runtime) throws InterruptedException {
+        RuntimeManagerLockThreadsInfo lockThreadsInfo = null;
+        synchronized (engineLocks) {
+            if(!engineLocks.containsKey(id)) {
+                engineLocks.put(id, new RuntimeManagerLockThreadsInfo(runtimeManagerLockFactory.newRuntimeManagerLock()));
```
gmunozfe, line 56: consider use "computeIfAbsent" instead

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+        this.engineLocks = new HashMap<>();
+    }
+
+    @Override
+    public void init(RuntimeManagerLockFactory factory) {
+        this.runtimeManagerLockFactory = factory;
+    }
+
+    @Override
+    public void lock(Long id, RuntimeEngine runtime) throws InterruptedException {
```
gmunozfe, line 52: Check first if id is null to return

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+            logger.warn("Interrupted lock {}", lockThreadsInfo.getRuntimeManagerLock());
+            throw e;
+        }
+        logger.debug("Lock {} taken for {} by {} for waiting threads by {}", lockThreadsInfo, id, runtime, lockThreadsInfo.hasQueuedThreads());
+    }
+
+    protected abstract void lock(RuntimeManagerLock lock) throws InterruptedException;
+    
+    @Override
+    public void unlock(Long id, RuntimeEngine runtime) {
```
gmunozfe, line 74: Check first if id is null to return

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+        this.engineLocks = new ConcurrentHashMap<>();
+        this.runtimeManagerLockFactory = new DefaultRuntimeManagerLockFactory();
+    }
+    @Override
+    public void init(RuntimeManagerLockFactory factory) {
+        this.runtimeManagerLockFactory = factory;
+    }
+
+    @Override
+    public void lock(Long id, RuntimeEngine runtime) {
```
gmunozfe, line 50: Check first if id is null to return

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+            logger.debug("Lock exists with {} waiting threads", lock.internalLock().getQueueLength());
+        }
+        logger.debug("Trying to get a lock {} for {} by {}", lock, id, runtime);
+        lock.lock();
+        logger.debug("Lock {} taken for {} by {} for waiting threads by {}", lock, id, runtime, lock.internalLock().hasQueuedThreads());
+
+    }
+
+    @Override
+    public void unlock(Long id, RuntimeEngine runtime) {
```
gmunozfe, line 66: Check first if id is null to return

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/SerializableRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.runtime.manager.impl.lock;
+
+import org.jbpm.runtime.manager.spi.RuntimeManagerLock;
+
+
+public class SerializableRuntimeManagerLockStrategy extends AbstractRuntimeManagerLockStrategy {
```
gmunozfe, line 22: Shouldn't it implement "Serializable" or is it not necessary?

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/lock/CustomRuntimeManagerLockStrategyTest.java
```diff
@@ -0,0 +1,92 @@
...
+        executorService = Executors.newFixedThreadPool(NUMBER_OF_WORKING_THREADS);
+    }
+
+    @After
+    public void destroy() {
+        executorService.shutdown();
+        executorService = null;
+    }
+
+    @Test
```
gmunozfe, line 52: timeout could be also added to the test

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/lock/CustomRuntimeManagerLockStrategyTest.java
```diff
@@ -0,0 +1,92 @@
...
+        executorService.shutdown();
+        executorService.awaitTermination(20, TimeUnit.SECONDS);
+
+        // assert hangs
+        Assert.assertTrue(executorService.isTerminated());
+        logger.info("exiting test TimeoutRuntimeManagerLockStrategy");
+    }
+
+
+    @Test
```
gmunozfe, line 72: timeout could be also added to the test

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/lock/RuntimeManagerLockStrategyTest.java
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    @After
+    public void destroy() {
+        executorService.shutdown();
+        executorService = null;
+    }
+
+    @Test
+    public void test() throws InterruptedException {
```
gmunozfe, line 77: more meaningful method name is appreciated, something like _testLockStrategy_

gmunozfe (commented):
Looks good to me, just minor comments to protect methods receiving null id.
Fantastic work @elguardian !

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -104,15 +114,22 @@
...
+
     
     public AbstractRuntimeManager(RuntimeEnvironment environment, String identifier) {
         this.environment = environment;
         this.identifier = identifier;
         if (registry.isRegistered(identifier)) {
             throw new IllegalStateException("RuntimeManager with id " + identifier + " is already active");
         }
+
+        if (!isUseLocking()) {
```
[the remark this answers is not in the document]
elguardian: done

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
elguardian: that is correct. Just trying to add the case of missuse of the runtime manager.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: removed

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: removed

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: removed

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: removed

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: removed

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
elguardian: done

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
elguardian: no need. it is guaranteed at abstract runtime manager level.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
elguardian: id won't be null ... it is guaranteed at runtime manager level

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/SerializableRuntimeManagerLockStrategy.java
elguardian: serializable is in the sense of requests not really about passsivation.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
elguardian: id cannot be null it is guaranteed at runtime manager level

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/lock/RuntimeManagerLockStrategyTest.java
elguardian: done

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/lock/CustomRuntimeManagerLockStrategyTest.java
elguardian: added

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/lock/CustomRuntimeManagerLockStrategyTest.java
elguardian: added

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
elguardian: id cannot me null. it is guaranteed by runtime manager

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
elguardian: good catch

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: in this case the call is delegated to createLockOnGetEngine... so it is the same thing (it is checked in there.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+        RuntimeManagerLock lock = engineLocks.get(id);
+        if (lock != null) {
+            if (!lock.internalLock().hasQueuedThreads()) {
+                logger.debug("Removing lock {} from list as non is waiting for it by {}", lock, runtime);
+                engineLocks.remove(id);
```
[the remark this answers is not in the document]
elguardian: Legacy implementation is just taking the old code and putting in here. 
You are right about the atomicity of those queue threads.

I had some issues with the thread reaper regarding queue threads. not always the owner of the thread tries to unlock. If that happens i t should not be dequeued. That is the reason you need to avoid the situation holding information about which threads are going to remove it from the set. 

1 thread is operation
2 thread reaper kicks in and executes a dispose... because detects has queued threads it removes it from the list. (it is not the owner actually). This allows coherence with the next call isHeldByCurrentThread and is consistent with the idea that only threads that lock can unlock and operate the engine locks map.

The idea is like only those threads operation locks can unlock.... thread reaper is not one of them.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+        this.runtimeManagerLockFactory = factory;
+    }
+
+    @Override
+    public void lock(Long id, RuntimeEngine runtime) throws InterruptedException {
+        RuntimeManagerLockThreadsInfo lockThreadsInfo = null;
+        synchronized (engineLocks) {
+            if(!engineLocks.containsKey(id)) {
+                engineLocks.put(id, new RuntimeManagerLockThreadsInfo(runtimeManagerLockFactory.newRuntimeManagerLock()));
+            }
```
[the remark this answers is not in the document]
elguardian: computeIfAbsent idea from gonzalo.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+                return;
+            }
+            lockThreadsInfo.unset();
+            if (lockThreadsInfo.count() == 0) {
+                logger.debug("[LOCK] Removing lock for {} for lock  {} from list as non is waiting for it by {}", id, lockThreadsInfo, runtime);
+                engineLocks.remove(id);
+            }
+        }
+
+        if (lockThreadsInfo.isHeldByCurrentThread()) {
```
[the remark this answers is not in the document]
elguardian: Actually there are more threads involved... the thread reaper can try to call the disposable when calls the tx synchronzation after completion. it something that surprised me after testing. I did not understand why that was there.

If you throw another exception during after completion the behavior of the tx manager is unknown. at least for me. that is the reason I set a warn message.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,136 @@
...
+            if (lockThreadsInfo == null) {
+                logger.warn("[LOCK] lock {} is already removed for {} unlocked by {}", id, lockThreadsInfo, runtime);
+                return;
+            }
+            lockThreadsInfo.unset();
+            if (lockThreadsInfo.count() == 0) {
+                logger.debug("[LOCK] Removing lock for {} for lock  {} from list as non is waiting for it by {}", id, lockThreadsInfo, runtime);
+                engineLocks.remove(id);
+            }
+        }
```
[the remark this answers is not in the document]
elguardian: see my comment above.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+    @Override
+    public void unlock() {
+        if(currentThreadId == Thread.currentThread().getId()) {
+            traces.clear(); // clean up
+        }
+        lock.unlock();
```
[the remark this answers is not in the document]
elguardian: if there is a place where lock for the same process instance id in the same thread is calling more than once. the logic of that part is broken. A thread (request execution) should only invoke once the lock. That is also the reason a added more that one stack traces... to add diaganose that possibility (more than one call to lock in one request)

That is pretty good idea actually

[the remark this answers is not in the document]
elguardian: done

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/RuntimeManagerLockStrategyFactory.java
```diff
@@ -0,0 +1,95 @@
...
+            if (currentRuntimeManagerLockFactory.getClass().getName().equals(lockFactoryClassName)) {
+                runtimeManagerLockFactory = currentRuntimeManagerLockFactory;
+                break;
+            }
+        }
+
+        // fallback to default
+        if (runtimeManagerLockFactory == null) {
+            runtimeManagerLockFactory = new DefaultRuntimeManagerLockFactory();
+        }
```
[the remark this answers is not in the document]
elguardian: ok

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/RuntimeManagerLockStrategyFactory.java
```diff
@@ -0,0 +1,95 @@
...
+            if (lockStrategy.getClass().getName().equals(lockStrategyClassName)) {
+                runtimeManagerLockStrategy = lockStrategy;
+                break;
+            }
+        }
+
+
+        if (runtimeManagerLockStrategy == null) {
+            runtimeManagerLockStrategy = new LegacyRuntimeManagerLockStrategy();
+        }
```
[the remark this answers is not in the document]
elguardian: ok makes a lot of sense

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,133 @@
...
+
+    @Override
+    public void init(RuntimeManagerLockFactory factory) {
+        this.runtimeManagerLockFactory = factory;
+    }
+
+    @Override
+    public void lock(Long id, RuntimeEngine runtime) throws InterruptedException {
+        RuntimeManagerLockThreadsInfo lockThreadsInfo = null;
+        synchronized (engineLocks) {
```
fjtirado, line 54: If ConcurrentHashMap is used for engineLocks I think we do not need that synchronize

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/AbstractRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,133 @@
...
+    protected static final Logger logger = LoggerFactory.getLogger(AbstractRuntimeManagerLockStrategy.class);
+
+    protected Map<Long, RuntimeManagerLockThreadsInfo> engineLocks; 
+
+    private RuntimeManagerLockFactory runtimeManagerLockFactory;
+
+    
+    protected AbstractRuntimeManagerLockStrategy() {
+        this.runtimeManagerLockFactory = new DefaultRuntimeManagerLockFactory();
+        this.engineLocks = new HashMap<>();
```
fjtirado, line 43: ConcurrentHashMap?

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/spi/RuntimeManagerLock.java
```diff
@@ -0,0 +1,35 @@
+/*
```
fjtirado, line 1: Just a quesiton, why are not direclty using an custom implementation of java.util.concurrent.Lock that delegates into the real lock

fjtirado: Or, just extends Lock and add methods of ReentrantLock that we are using

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+        RuntimeManagerLock lock = engineLocks.putIfAbsent(id, newLock);
+        if (lock == null) {
+            lock = newLock;
+            logger.debug("New lock created as it did not exist before");
+        } else {
+            logger.debug("Lock exists with {} waiting threads", lock.internalLock().getQueueLength());
+        }
+        logger.debug("Trying to get a lock {} for {} by {}", lock, id, runtime);
+        lock.lock();
+        logger.debug("Lock {} taken for {} by {} for waiting threads by {}", lock, id, runtime, lock.internalLock().hasQueuedThreads());
```
fjtirado, line 61: Is that trace right, last argument is boolean

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/spi/RuntimeManagerLock.java
elguardian: because we need a way to ensure that other implementations can do other operations within....example of DebugRuntimeManagerLock which is the one taking data of the owner stacktrace that got the lock

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -431,49 +440,28 @@ protected void createLockOnGetEngine(Context<?> context, RuntimeEngine runtime)
...
-                logger.debug("Lock exists with {} waiting threads", lock.getQueueLength());
-            }
-            logger.debug("Trying to get a lock {} for {} by {}", lock, id, runtime);
-            lock.lock();
-            logger.debug("Lock {} taken for {} by {} for waiting threads by {}", lock, id, runtime, lock.hasQueuedThreads());
-            
+        try {
+            runtimeManagerLockStrategy.lock(id, runtime);
+        } catch(InterruptedException e) {
+            throw new RuntimeException("Runtime manager was not able to lock for {" + id +"}", e);
```
fjtirado, line 449: Style, show we just throw a more specifc runtime here? For example, IllegalStateException

fjtirado (commented):
LGTM, but I have some style comments that might worth a look

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/LegacyRuntimeManagerLockStrategy.java
elguardian: right. needs to be count

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
elguardian: I didn't create any specific actually and it is not really a problem with the state.. should be something like RuntimeManagerLockException but it is quite cumbersome...

fjtirado (approved):

elguardian:
Jenkins retest this

elguardian:
Jenkins retest this

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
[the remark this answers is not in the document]
MarianMacik: Could we use here just `getHoldCount()` and clear traces only if it is equals to 1? Maybe we wouldn't need an additional counter for that. As right now when we have 2 locks called from the same thread (I know it shouldn't happen), the counter is decremented in a non-thread-safe way I think. Same goes for increments in case there are 2 lock operations called by the same thread. Also the counter should be probably reset in case a new thread has acquired the lock. I think getHoldCount() would be OK for this as it uses the internal counter of the lock.

MarianMacik (approved):

elguardian:
Jenkins retest this

elguardian:
Jenkins restest this

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/lock/DebugRuntimeManagerLock.java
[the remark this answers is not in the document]
elguardian: will change

PR #1849, opened 2021-01-28
[RHPAM-3425] PerCaseRuntimeManager getting stuck because bad cleanup (error 504)

elguardian:
Jira https://issues.redhat.com/browse/RHPAM-3425

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
@@ -185,6 +185,7 @@ public class PerCaseRuntimeManager extends AbstractRuntimeManager {
             processInstanceId = (Long) contextId;
         }
         Long ksessionId = mapper.findMapping(context, this.identifier);
+        ((RuntimeEngineImpl) runtime).setLockedKieSessionId(ksessionId);
         createLockOnGetEngine(ksessionId, runtime);
         saveLocalRuntime(caseId, processInstanceId, runtime);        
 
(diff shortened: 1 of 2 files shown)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -185,6 +185,7 @@ public RuntimeEngine getRuntimeEngine(Context<?> context) {
             processInstanceId = (Long) contextId;
         }
         Long ksessionId = mapper.findMapping(context, this.identifier);
+        ((RuntimeEngineImpl) runtime).setLockedKieSessionId(ksessionId);
```
krisv, line 188: @elguardian you're setting locked kieSessionId just before calling createLockOnGetEngine(..), and then using this lockedSessionId to release the lock, are we sure this is the only way to create a lock? Because if we miss one then trying to dispose the engine could break later if lockedKieSessionId is now null?  I assumed the MaintainMappingListener was also creating a lock on beforeProcessStarted, but it doesn't seem to be doing that for cases, only for per process.  Were you explicitly trying to limit the changes only for PerCaseRuntimeStrategy? Or would trying to move it to AbstractRuntimeManager make sense, where we can do it just before locking the runtime so it will work like this across the board?  I guess both have somewhat different risks so probably either approach could work, just trying to make sure we doublecheck to avoid issues later.

elguardian: it is not a good fix.

elguardian:
This prooved not to solve all the issues. give another round.

elguardian:
supperseeded by https://github.com/kiegroup/jbpm/pull/1850

PR #1850, opened 2021-01-29
[JBPM-9576] PerCaseRuntimeManager getting stuck because bad cleanup

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9576

This fixes
1) get lock / release lock when a process instance is being close during operation
2) new kie sessions created because removal mapping info in the case of case kie session scope


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
```diff
@@ -81,23 +87,53 @@ public Long findMapping(Context context, String ownerId) {
...
+                return null;
+            }
+
+            Query findQuery = em.createNamedQuery("FindContextMapingByAuditContextId", ContextMappingInfo.class)
+                                .setParameter("contextId", context.getContextId().toString())
+                                .setParameter("ownerId", ownerId);
+
+            ContextMappingInfo contextMappingInfo = (ContextMappingInfo) findQuery.getSingleResult();
+            if (contextMappingInfo != null) {
+                return contextMappingInfo.getKsessionId();
```
MarianMacik, line 106: Not sure if for the future we shouldn't directly return just Long from the DB. This could probably improve performance.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -116,66 +115,31 @@ public RuntimeEngine getRuntimeEngine(Context<?> context) {
...
+            logger.debug("Runtime engine found in local {} with context {}", localRuntime, contextId);
+            return localRuntime;
+        }
+
+        Long ksessionId = mapper.findMapping(context, this.getIdentifier());
+        if(ksessionId == null && mapper instanceof InternalMapper) {
+            // we fall back as process instance must be closed but not really case and kie session must be still there.
+            logger.debug("PerCaseRuntimeManager KieSessionId found fallingback for context {} and owner {}", context.getContextId(), getIdentifier());
+            ksessionId = ((InternalMapper) mapper).findLogMapping(context, identifier);
+        }
```
MarianMacik, line 130: Maybe we could put this into a separate method and defer this further in case we don't want to eagerly initialize the ksession? Before it was done in initializer.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -324,12 +367,14 @@ public boolean validate(Long ksessionId, Long processInstanceId) {
 
         @Override
         public void afterProcessCompleted(ProcessCompletedEvent event) {
+            logger.debug("Removing persistence mapping for kieSessionId {} runtime engine {} manager {} and case id",ksessionId, runtime, managerId, caseId);
```
gmunozfe, line 370: Missing placeholder for the last param

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -324,12 +367,14 @@ public boolean validate(Long ksessionId, Long processInstanceId) {
...
             mapper.removeMapping(new EnvironmentAwareProcessInstanceContext(event.getKieRuntime().getEnvironment(), event.getProcessInstance().getId()), managerId);
-            removeLocalRuntime(runtime, event.getProcessInstance().getId());                       
+            removeLocalRuntime(runtime, event.getProcessInstance().getId());
         }
 
         @Override
         public void beforeProcessStarted(ProcessStartedEvent event) {
+            logger.debug("Saving persistence mapping for kieSessionId {} runtime engine {} manager {} and case id",ksessionId, runtime, managerId, caseId);
```
gmunozfe, line 377: Same here, missing placeholder for the last param

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/jpa/ContextMappingInfo.java
```diff
@@ -44,19 +44,15 @@
...
-    {@NamedQuery(name="FindContextMapingByContextId", 
-                query="from ContextMappingInfo where contextId = :contextId"
-                		+ " and ownerId = :ownerId"),
-                @NamedQuery(name="FindContextMapingByKSessionId", 
-                query="from ContextMappingInfo where ksessionId = :ksessionId"
-                		+ " and ownerId = :ownerId"),
-                @NamedQuery(name="FindKSessionToInit", 
-                query="select cmInfo.ksessionId from ContextMappingInfo cmInfo, "
+    {@NamedQuery(name="FindContextMapingByContextId", query="from ContextMappingInfo where contextId = :contextId and ownerId = :ownerId"),
+     @NamedQuery(name="FindContextMapingByAuditContextId", query="select o from ContextMappingInfo o join ProcessInstanceLog log on log.correlationKey = o.contextId where CONCAT(log.processInstanceId, '') = :contextId and o.ownerId = :ownerId"),
```
gmunozfe, line 48: Is there another way to convert _log.processInstanceId_ to String not using CONCAT function?

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -116,66 +115,31 @@ public RuntimeEngine getRuntimeEngine(Context<?> context) {
...
+        RuntimeEngine localRuntime = findLocalRuntime(contextId);
+        if (localRuntime != null) {
+            logger.debug("Runtime engine found in local {} with context {}", localRuntime, contextId);
+            return localRuntime;
+        }
+
+        Long ksessionId = mapper.findMapping(context, this.getIdentifier());
+        if(ksessionId == null && mapper instanceof InternalMapper) {
+            // we fall back as process instance must be closed but not really case and kie session must be still there.
+            logger.debug("PerCaseRuntimeManager KieSessionId found fallingback for context {} and owner {}", context.getContextId(), getIdentifier());
```
gmunozfe, line 128: Trace message is not clear, it should state that kieSessionId is null (or not found)

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/InternalMapper.java
```diff
@@ -18,10 +18,20 @@
...
 import org.kie.internal.runtime.manager.Mapper;
 
 
 public abstract class InternalMapper implements Mapper {
 
     public abstract List<String> findContextIdForEvent(String eventType, String ownerId);
+
+    /**
+     * Finds ksession for given context
+     * @param context instance of the context
```
gmunozfe, line 31: Missing ownerId param in the javadoc

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
```diff
@@ -81,23 +87,53 @@ public Long findMapping(Context context, String ownerId) {
...
+            }
+            return null;
+        } catch (NoResultException e) {
+            return null;
+        } catch (NonUniqueResultException e) {
+            return null;
```
gmunozfe, line 112: May use catch multiple exceptions

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -116,66 +115,31 @@ public RuntimeEngine getRuntimeEngine(Context<?> context) {
...
-                    ksession = factory.findKieSessionById(ksessionId);
-                }
-            }
+        RuntimeEngine localRuntime = findLocalRuntime(contextId);
+        if (localRuntime != null) {
+            logger.debug("Runtime engine found in local {} with context {}", localRuntime, contextId);
+            return localRuntime;
+        }
+
+        Long ksessionId = mapper.findMapping(context, this.getIdentifier());
```
gmunozfe, line 125: This is now out of the synchronized block, wouldn't it be an issue in concurrent scenarios?

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -116,66 +115,31 @@ public RuntimeEngine getRuntimeEngine(Context<?> context) {
...
+        if(ksessionId == null && mapper instanceof InternalMapper) {
+            // we fall back as process instance must be closed but not really case and kie session must be still there.
+            logger.debug("PerCaseRuntimeManager KieSessionId found fallingback for context {} and owner {}", context.getContextId(), getIdentifier());
+            ksessionId = ((InternalMapper) mapper).findLogMapping(context, identifier);
```
gmunozfe, line 129: Same here, out of the synchronized block.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
```diff
@@ -25,6 +25,7 @@
 import javax.persistence.NonUniqueResultException;
 import javax.persistence.Query;
 
+import org.jbpm.runtime.manager.impl.PerCaseRuntimeManager;
```
gmunozfe, line 28: Remove it, not used

gmunozfe (commented):
Looks good, just minor things related to trace accuracy.
Also a question about findMapping, findLogMapping outside the synchronized block

jbpm-case-mgmt/jbpm-case-mgmt-impl/src/test/resources/logback-test.xml
```diff
@@ -9,7 +9,7 @@
     </encoder>
   </appender>
 
-  <logger name="org.jbpm" level="info"/>
+  <logger name="org.jbpm" level="debug"/>
```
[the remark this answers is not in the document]
elguardian: done

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -426,52 +426,60 @@ public void setKieContainer(KieContainer kieContainer) {
...
-        if (!isUseLocking()) {
-            logger.debug("Locking on runtime manager disabled");
-            return;
-        }
-        
         if (context instanceof ProcessInstanceIdContext) {
             Long piId = ((ProcessInstanceIdContext) context).getContextId();
             createLockOnGetEngine(piId, runtime);
+        } else {
+            logger.debug("Trying to release lock with a non proper context onn runtime manager engine {}", runtime);
```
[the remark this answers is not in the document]
elguardian: typo fixed

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/TimerInitFailureRuntimeManagerTest.java
```diff
@@ -184,7 +184,6 @@ private RuntimeEnvironment createEnvironment(NodeLeftCountDownProcessEventListen
                 .entityManagerFactory(emf)
                 .userGroupCallback(userGroupCallback)
                 .registerableItemsFactory(new DefaultRegisterableItemsFactory(){
-
```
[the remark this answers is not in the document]
elguardian: removed from index.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
```diff
@@ -81,23 +87,53 @@ public Long findMapping(Context context, String ownerId) {
...
+            em.remove(contextMappingInfo);
         }
         if (!info.isShared()) {
     		em.close();
     	}
     }
     
-    protected Context resolveContext(Context orig, EntityManager em) {
+    @SuppressWarnings("unchecked")
+    protected <T> Context<T> resolveContext(Context orig, EntityManager em) {
```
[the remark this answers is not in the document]
elguardian: avoids cast in caller.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
elguardian: context mapping info is necessary stuff for computing the kie session, either you get 3 fields or one, I don't think we will get much different. In any case without numbers is difficult to say

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
elguardian: It has always been done during building of the runtime engine (not init) before getting the lock
https://github.com/kiegroup/jbpm/blob/7.44.0.Final/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java#L187
So it is mandatory to compute the kie session id. Here we are not initializing anything. 
Actually lazy init performed this query twice:
https://github.com/kiegroup/jbpm/blob/7.44.0.Final/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java#L669 (here two)
so we removed one query actually

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
elguardian: added

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
elguardian: added

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
elguardian: changed to not found.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
elguardian: done

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
elguardian: removed

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
elguardian: The integrity of the mapping existence is controlled by DB so there is no need for this to be in a sync block. Any operation adding / removing the mapping is atomic to the transaction so that is not going to change during this tx no matter how many calls you do.
The sync block in this case does not make much sense.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
elguardian: The integrity of the mapping existence is controlled by DB so there is no need for this to be in a sync block. Any operation adding / removing the mapping is atomic to the transaction so that is not going to change during this tx no matter how many calls you do.
The sync block in this case does not make much sense.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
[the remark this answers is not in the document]
MarianMacik: I tried this locally. If you leave this returning just `Context` and then in the `saveMapping` method just write:

```
Context resolvedContext = resolveContext(context, em);
em.persist(new ContextMappingInfo(resolvedContext.getContextId().toString(),ksessionId, ownerId));
```
It will work without any type parameters.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/mapper/JPAMapper.java
MarianMacik: Might be.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
MarianMacik: Actually same holds for eager initialization. Once it was called in init block, then just before getting the lock. So makes sense.

MarianMacik (approved):
Looks good to me. We can just get rid of the `Context<Object>`, so I would do that as it is easier to read.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/jpa/ContextMappingInfo.java
elguardian: that could work in every db, no.

PR #1859, opened 2021-02-05
JBPM-9594 - Improve synchronization of testPerCaseCleanup

MarianMacik:
**JIRA**: [JBPM-9594](https://issues.redhat.com/browse/JBPM-9594)

* Improved synchronization so now it is guaranteed that at least 2 out of 6 iterations are run after the process instance is aborted. Before it was possible that the spawned thread finished before the abort happened, thus not testing the desired scenario. I also considered to interleave the threads deterministically with 2 latches, but the code was less clear and not the worth it.
* Swapped the order of asserts and logs to first print log message and only then potentially fail on assert.


gmunozfe:
@MarianMacik could you backport it to 7.48.x as well?

PR #1869, opened 2021-02-17
[JBPM-9614] Problem invoking WebServiceWorkItemHandler on Java 11

fjtirado:
Previous solution was not valid because URLS generated by
VFSResouceLoader has vfs protocol, which is not understood by cxf.

Hence I sligthly modified original approach. First list of
ResourceLoaders is obtained (package method). Then, depending on ResourceLoader type (we
are really only interested on Jar or VFS) we access different internal
private fields (since ResourceLoader unfortunately does not have getRootURL method).

In the case of VFS, we need extra logic to generate the right path to
jar file (went to parent and the append the jar name, as in original
fix)

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9614)

### DIFF
--- a/jbpm-workitems/jbpm-workitems-webservice/src/main/java/org/jbpm/process/workitem/webservice/WebServiceWorkItemHandler.java
+++ b/jbpm-workitems/jbpm-workitems-webservice/src/main/java/org/jbpm/process/workitem/webservice/WebServiceWorkItemHandler.java
@@ -16,14 +16,16 @@
 
 package org.jbpm.process.workitem.webservice;
 
+import java.io.File;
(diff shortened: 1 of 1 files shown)


fjtirado:
run fdb

fjtirado:
jenkins run fdb

afalhambra (approved):
LGTM

afalhambra:
failure test not related

fjtirado:
Unrelated failure on fdb
@MarianMacik @elguardian I think we are ready to go with this one

MarianMacik (approved):

PR #1873, opened 2021-02-22
[JBPM-9625] Allow to bypass the authenticated user in the KIE REST API that is adding users or groups as potential owners

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9625

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/AddPeopleAssignmentsCommand.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/AddPeopleAssignmentsCommand.java
@@ -40,6 +40,7 @@ public class AddPeopleAssignmentsCommand extends UserGroupCallbackTaskCommand<Vo
     private OrganizationalEntity[] entities;
     private boolean removeExisting;
 
+
     public AddPeopleAssignmentsCommand(String userId, long taskId, int type, OrganizationalEntity[] entities, boolean removeExisting) {
         super();
         setUserId(userId);
--- a/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/admin/UserTaskAdminService.java
+++ b/jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/admin/UserTaskAdminService.java
@@ -102,6 +102,56 @@ public interface UserTaskAdminService {
     void addBusinessAdmins(String deploymentId, long taskId,  boolean removeExisting, OrganizationalEntity ...orgEntities) throws TaskNotFoundException;
     
     /**
+     * Adds new potential owners for a given task. Can be instructed to remove existing potential owners
+     * which means that the task will be released in case it was claimed already. 
+     * Task must be in active state.
+     * 
+     * @param userId user doing the update
+     * @param deploymentId
+     * @param taskId unique task id
+     * @param removeExisting specifies if existing potential owners should be removed 
+     * @param orgEntities one or more potential owner to be added to a task
+     * @throws TaskNotFoundException thrown when there is no task with given id or is not associated with given deployment id
+     */
+    void addPotentialOwners(String userId,
+                            String deploymentId,
+                            long taskId,
+                            boolean removeExisting,
+                            OrganizationalEntity... orgEntities) throws TaskNotFoundException;
+    /**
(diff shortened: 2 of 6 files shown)


jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/admin/UserTaskAdminService.java
```diff
@@ -95,12 +95,64 @@
      * 
      * @param deploymentId
      * @param taskId unique task id
-     * @param removeExisting specifies if existing business admins should be removed 
+     * @param removeExisting specifies if existing business a@Override
```
sutaakar, line 98: Seems to be unintentional change.

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/admin/UserTaskAdminService.java
```diff
@@ -95,12 +95,64 @@
...
+
+    /**
+     * Adds new business admin for a given task. Can be instructed to remove existing business admins. 
+     * Task must be in active state.
+     * 
+     * @param userId user doing the update
+     * @param deploymentId
+     * @param taskId unique task id
+     * @param removeExisting specifies if existing business a@Override
+    dmins should be removed
```
sutaakar, line 147:
```suggestion
     * @param removeExisting specifies if existing business admins should be removed 
```

elguardian: done

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/admin/UserTaskAdminService.java
elguardian: done

elguardian:
Jenkins retest this

elguardian:
Jenkins retest this please

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/AddPeopleAssignmentsCommand.java
```diff
@@ -56,7 +62,7 @@ public Void execute(Context cntxt) {
         
         Task task = context.getTaskQueryService().getTaskInstanceById(taskId);       
         // security check
-        if (!isBusinessAdmin(userId, task.getPeopleAssignments().getBusinessAdministrators(), context)) {
+        if (!skipOwnerCheck && !isBusinessAdmin(userId, task.getPeopleAssignments().getBusinessAdministrators(), context)) {
```
MarianMacik, line 65: We were discussing this with @sutaakar and realized that the same property is now used for 2 slightly different things. It can be set for example for getting tasks assigned as a potential user, in which case a userId has to be provided and this will be used for the query. Or the same can be used for example for comments, when even when using the bypass we have to provide the correct user for that particular case, otherwise the security check for the case will fail. So it can be said that in both these examples we can bypass the authentication by providing a different, but always valid user with regards to the context of the operation. So bypass in these cases just allows users to behave as another user, but still a valid one.

However, here it seems it doesn't really matter which user we provide, the only important thing is if the `skipOwnerCheck` is true, which means if bypass is activated. In that case we can completely bypass the authentication without knowing any context of the operation. So we can even provide "non-existing-user" and the operation would succeed. Wouldn't it be better to use the same approach as we have used until now? That means that even with bypass activated a user would have to provide a valid user for this context, so they would have to provide the correct business administrator user. So bypass actually allows to "impersonate" another, still valid user, in that case a business administrator. What do you think?

Also I noticed that actually the passed `userId` doesn't seem to be used at all when a bypass is turned on in the current implementation.

elguardian:
Jenkins retest this

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/AddPeopleAssignmentsCommand.java
elguardian: the current user ends up in the TaskEventImpl. See test case.
Removed the skipping flag.

sutaakar (approved):

PR #1880, opened 2021-03-01
[JBPM-9632] Intermediate Throw event signal stops processing workflow before signal is processed.

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9632

To avoid problems with the action nodes (intermediate throw) I added a new metadata for the node
"executeActionAfterComplete" when is set to true the signal is process after finishing processing the current flow.(like a passthrough)

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
@@ -187,6 +187,9 @@ public class IntermediateThrowEventHandler extends AbstractNodeHandler {
 			final String uri, final String localName,
 			final ExtensibleXmlParser parser) throws SAXException {
 		ActionNode actionNode = (ActionNode) node;
+		String executeActionAfterCompleteMetadata = (String) node.getMetaData().get("executeActionAfterComplete");
+		Boolean executeActionAfterComplete = executeActionAfterCompleteMetadata != null && Boolean.parseBoolean(executeActionAfterCompleteMetadata);
+		actionNode.setExecuteActionAfterComplete(executeActionAfterComplete);
 		org.w3c.dom.Node xmlNode = element.getFirstChild();
 		while (xmlNode != null) {
 			String nodeName = xmlNode.getNodeName();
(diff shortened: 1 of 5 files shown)


jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
```diff
@@ -187,6 +187,9 @@ public void handleSignalNode(final Node node, final Element element,
 			final String uri, final String localName,
 			final ExtensibleXmlParser parser) throws SAXException {
 		ActionNode actionNode = (ActionNode) node;
+		String executeActionAfterCompleteMetadata = (String) node.getMetaData().get("executeActionAfterComplete");
+		Boolean executeActionAfterComplete = executeActionAfterCompleteMetadata != null ? Boolean.parseBoolean(executeActionAfterCompleteMetadata) : false;
```
gmunozfe, line 191: It could be simplified without a ternary operator:
```suggestion
		Boolean executeActionAfterComplete = executeActionAfterCompleteMetadata != null && Boolean.parseBoolean(executeActionAfterCompleteMetadata);
```

jbpm-test-coverage/src/test/java/org/jbpm/test/regression/subprocess/IntermediateEventSubprocessCatchTest.java
```diff
@@ -0,0 +1,50 @@
...
+    public void init() throws Exception {
+        ksession = createKSession(INTERMEDIATE_CATCH_EVENT_SUBPROCESS_PROCESS);
+    }
+
+    @Test(timeout = 30000)
+    public void testIntermediateCatchEventSubprocess() throws Exception {
+
+
+        ProcessInstance pi = ksession.startProcess(INTERMEDIATE_CATCH_EVENT_SUBPROCESS_PROCESS_ID, Collections.singletonMap("id", "1"));
+        Thread.sleep(5000L);
```
gmunozfe, line 46: Could you use a listener instead of sleeping to check that the process has finished?

gmunozfe (commented):
Looks good to me, a minor comment and also an improvement suggestion about the test: another test for the opposite scenario -_executeActionAfterComplete_ metadata equals false or no metadata (default)- could be added as well.

jbpm-test-coverage/src/test/java/org/jbpm/test/regression/subprocess/IntermediateEventSubprocessCatchTest.java
elguardian: sure

PR #1887, opened 2021-03-03
(GSS)[7.48.x][JBPM-9638] - Test case for JBPM-9623 (Dynamic Sub process is getting invoked twice instead of once when implemented in loop back)

afalhambra:
**[JBPM-9638](https://issues.redhat.com/browse/JBPM-9638)**: Test case for JBPM-9623 (Dynamic Sub process is getting invoked twice instead of once when implemented in loop back) - backport.


gmunozfe (approved):
Looks good to me, thanks for adding the test, awesome job @afalhambra !

afalhambra:
@mareknovotny or @elguardian would you please merge? Thanks!

PR #1890, opened 2021-03-04
[JBPM-9641] Convert process variable to right data type

fjtirado:
**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9641)

**referenced Pull Requests**: 
https://github.com/kiegroup/droolsjbpm-integration/pull/2419

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
@@ -22,6 +22,7 @@ import java.util.List;
 
 import org.jbpm.process.core.Context;
 import org.jbpm.process.core.context.AbstractContext;
+import org.jbpm.process.core.datatype.DataType;
 import org.jbpm.process.core.datatype.impl.type.ObjectDataType;
 
 /**
@@ -42,6 +43,7 @@ public class VariableScope extends AbstractContext {
         this.variables = new ArrayList<Variable>();
     }
     
+    @Override
     public String getType() {
         return VariableScope.VARIABLE_SCOPE;
     }
@@ -61,7 +63,7 @@ public class VariableScope extends AbstractContext {
         final String[] result = new String[this.variables.size()];
         if (this.variables != null) {
             for ( int i = 0; i < this.variables.size(); i++ ) {
-                result[i] = ((Variable) this.variables.get( i )).getName();
+                result[i] = this.variables.get( i ).getName();
             }
         }
         return result;
@@ -83,6 +85,7 @@ public class VariableScope extends AbstractContext {
         return null;
     }
 
+    @Override
     public Context resolveContext(Object param) {
         if (param instanceof String) {
             return findVariable((String) param) == null ? null : this;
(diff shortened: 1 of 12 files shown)


gmunozfe:
jenkins retest this

afalhambra (approved):
Awesome work @fjtirado!

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
```diff
@@ -91,22 +92,26 @@ public Context resolveContext(Object param) {
...
-	    	if (!isValidType) {
-	    		throw new IllegalArgumentException("Variable '" + name +"' has incorrect data type expected:" 
-	    						+ var.getType().getStringType() + " actual:" + value.getClass().getName());
-	    	}
-    	}
-	}
+    public Object validateVariable(String processName, String name, Object value) {
+        Variable var = findVariable(name);
+        if (var == null) {
+            if (variableStrictEnabled) {
```
elguardian, line 98: you did change the semantics of the flag variableStrictEnabled

fjtirado: no, if variableStrictEnabled is true, code behaves the same. I change behaviour when semantic is false, to try to convert values to the right data type and if that attemps fails, use the same object that was provided, so the only change is that if user pass a string that can be converted to int for a int data type, int will be used (as requested by customer)

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
@elguardian finally green, please take another look

jbpm-flow/src/main/java/org/jbpm/process/core/datatype/DataType.java
```diff
@@ -37,4 +37,14 @@
...
     String getStringType();
 
+    /**
+     * Tries to convert parameter to an object of this type
+     * @param value to be converted. 
+     * For the sake of performance, it is assumed to not be null and not of <code>this</code> data type. 
+     * @return object converted to this type, if not possible to convert, the same object passed as parameter
+     */
+    default Object convertTo(Object value) {
+        return value;
```
elguardian, line 47: if we are going to do this, let's try to refine it a bit. convertTo is not really a good name. Would be better naming something like convertFrom  -> value is the source not the target.

All object primitive types has the valueOf method, so it would not be that bad to

Class -> get the class from String type
method = get the method valueOf  with argument value.getClass()
if the method is not found you can throw an error
method.invoke(class, valueOF, argument)

this is more generic and will allow more fine grain control even for user types.
Also this is generic so you can remove the impl from the sub types.

elguardian (changes requested):
change the impl plz, this one could cause some problems.

jbpm-flow/src/main/java/org/jbpm/process/core/datatype/DataType.java
fjtirado: Lets reduce the scope to be less ambition, I will change the method signature to valueOf(String value) and override it only for those types which implementation is trivial

PR #1892, opened 2021-03-05
[JBPM-6943] Renamed deprecated method inside AbstractProcessAssembler

gitgabrio:
@danielezonca @afalhambra  @jiripetrlik

See https://issues.redhat.com/browse/JBPM-9643

Scope of this PR is to allow is to remove deprecated method.
A fix has already been made with https://github.com/kiegroup/droolsjbpm-knowledge/pull/508 but FDB take too much time.
This is an attempt to speed up resolution.
In my local machine I started full clean install and I've successfully  gone past the module that was failing, but the full build takes hours.

@jiripetrlik I'm not sure if you are the QE responsible for this project :smile:

### DIFF
--- a/jbpm-flow-builder/src/main/java/org/jbpm/assembler/AbstractProcessAssembler.java
+++ b/jbpm-flow-builder/src/main/java/org/jbpm/assembler/AbstractProcessAssembler.java
@@ -33,7 +33,7 @@ import org.kie.internal.builder.KnowledgeBuilderResult;
 
 public abstract class AbstractProcessAssembler implements KieAssemblerService {
 
-    public void addResource(
+    public void addResourceAfterRules(
             Object kbuilder,
             Resource resource,
             ResourceType type,


afalhambra:
Jenkins run fdb

MarianMacik:
@gitgabrio @danielezonca @jiripetrlik Next time, please, if you don't want to run FDB, just name your branches in all repositories with the same name and PRs will connect together. So if you didn't already merge the drools PR and instead opened this one with the same branch from your fork and ran PR check here, it would test it together.

But because you didn't to it, our nightly test pipelines failed and we have a large portion of tests providing false negative results.
I agree that FDB may take some time, but if you already know you need to change something in other repositories, do it cross-repo with the same branch and merge all at once, please. Thank you!

gitgabrio:
@MarianMacik 
Sorry, but I named all my branch the same way (DROOLS-6048), as I always do.
The point here are different, namely
1) I did not think (my fault) I had to change JBPM
2) I wrongly thought cdb was enough to detect all cases

MarianMacik:
@gitgabrio OK I thought that you knew the jbpm needed a change as well. Then the only way how to find it out would be FDB, that's for sure.

MarianMacik (approved):

gitgabrio:
@kiegroup/gatekeepers

Could you please merge this, squashing commits? Thanks!

PR #1896, opened 2021-03-16
[JBPM-9650] Potential owners of tasks retrieved by query don't match expected owners

fjtirado:
Comparison to remove the ID column should be case sensitive. Constant is
upper case and column names in pgsql/mysql are lower case. This causes
the ID column to be grouped, which leads to redundant rows that makes
the test fails

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9650)


afalhambra (approved):
Looks good to me.

It works fine now.

PR #1904, opened 2021-03-26
rename sync process api

elguardian:


gmunozfe (approved):
Looks good to me, nice renaming @elguardian

PR #1916, opened 2021-04-15
[JBPM-9700] Signal name not resolved when adding listener

fjtirado:
**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9700)

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
@@ -433,7 +433,7 @@ public abstract class WorkflowProcessInstanceImpl extends ProcessInstanceImpl
                         }
 
     	                RuntimeEngine runtime = manager.getRuntimeEngine(context);
-						KieRuntime managedkruntime = (KieRuntime) runtime.getKieSession();
+						KieRuntime managedkruntime = runtime.getKieSession();
     	                managedkruntime.signalEvent("processInstanceCompleted:" + getId(), this);
                 	} catch (SessionNotFoundException e) {
                 		// in case no session is found for parent process let's skip signal for process instance completion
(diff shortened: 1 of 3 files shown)


gmunozfe (commented):

I also have a comment for the code below (not modified by this PR) that it's adding the _eventListener_ twice in case of variable expression. It doesn't seem correct at all (I think it's more clear as it's done in this PR with an "else clause):
https://github.com/kiegroup/jbpm/blob/62f92c98f5d3f68f08fd519638259be97aa2d574/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java#L573-L576

jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -566,7 +567,12 @@ public TimerInstance configureSLATimer(String slaDueDateExpression, String timer
     protected void registerExternalEventNodeListeners() {
         for (Node node : getWorkflowProcess().getNodes()) {
             if (node instanceof EventNode && "external".equals(((EventNode) node).getScope())) {
-                addEventListener(((EventNode) node).getType(), EMPTY_EVENT_LISTENER, true);
+                String eventType = ((EventNode) node).getType();
+                if (isVariableExpression(eventType)) {
+                    addEventListener(resolveVariable(eventType), EMPTY_EVENT_LISTENER, true);
```
elguardian, line 572: if you modify the add... modify the remove unregisterExternalEventNodeListeners

elguardian:
add test case (you have the bpmn2)
tbh honest I don't consider this a valid use case as we are tying this to a process variable which may vary dynamically to a external source.

gmunozfe (approved):

elguardian:
probably will superseeded by
https://github.com/kiegroup/jbpm/pull/1909

PR #1926, opened 2021-05-04
[JBPM-9717] Timer instance incorrectly removed upon process execution error

elguardian:
during rollback transaction the timer were incorrecly removed

Jira https://issues.redhat.com/browse/JBPM-9717

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
@@ -20,7 +20,6 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.ConcurrentModificationException;
-import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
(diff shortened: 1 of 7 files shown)


jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -215,13 +218,27 @@ public void clearProcessInstances() {
...
             TimerManager timerManager = ((InternalProcessRuntime)kruntime.getProcessRuntime()).getTimerManager();
-            
+            TimerService timerService = timerManager.getTimerService();
+            if(timerService instanceof RegisteredTimerServiceDelegate) {
+                timerService = ((RegisteredTimerServiceDelegate) timerService).getTimerService();
+            }
+            if(timerService instanceof GlobalTimerService) {
+                // if the timer instance is transactional we don't need to clean up the timers as the rollback part of the transaction awareness
+                // is happening
+                if(!((GlobalTimerService) timerService).isTransactional()) {
```
martinweiler, line 235: @elguardian shouldn't this method return if the timerService *is* actually transactional? If yes, the ! negation should be removed here imo.

elguardian: well... in spite it seems. the answer is no. When is not transactinal the clean up is done in another place (interceptor).
This is a bit strange. This clean up was done 9 years ago while more modern parts (regarding interceptors were done few years later. Also ejb has a flag that is totally out of place (true/false if they can be executed in tx when ejb are always executed within tx. 
Tbh honest this clean up should be removed but at this point it is not really safe.
so this is the best I can do. move the clean up when it is not transactional to the interceptor and remove that situation from here.
This saves the rest of the cases like quartz in spring boot and jdk thread pools

martinweiler: I tested this again and it does not work for me. Running in kie-server, isTransactional is true, so the cleanup code is still called.

gmunozfe:
jenkins retest this

elguardian:
Jenkins retest this please

elguardian:
Jenkins retest this please

gmunozfe (approved):
Looks good to me, fabulous work @elguardian !
Just a couple of very minor format fixes, in case you want to update and run CI again...

Tested with different properties (`org.jbpm.ejb.timer.local.cache=false/true` `org.jbpm.ejb.timer.tx=false/true`) and bpmn and works like a charm.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/tx/NoTransactionalTimerResourcesCleanupAwareSchedulerServiceInterceptor.java
```diff
@@ -0,0 +1,174 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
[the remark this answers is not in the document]
elguardian: done

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/tx/TransactionAwareSchedulerServiceInterceptor.java
```diff
@@ -94,8 +98,21 @@ public void beforeCompletion() {
...
+            } else {
+                if (status == TransactionManager.STATUS_ROLLEDBACK) {
+                    // we remove from the timer from the map in timer manager
+                    JobContext ctxorig = timerJobInstance.getJobContext();
+                    ctxorig = (ctxorig instanceof SelfRemovalJobContext) ? ((SelfRemovalJobContext) ctxorig).getJobContext() : ctxorig; // unwrap
+                    if(!(ctxorig instanceof ProcessJobContext)) {
+                        return;
+                    }
+                    Optional<InternalKnowledgeRuntime> runtime = ctxorig.getInternalKnowledgeRuntime();
+                    if(runtime.isPresent()) {
```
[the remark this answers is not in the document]
elguardian: done

PR #1932, opened 2021-05-18
[JBPM-9743] Event Emitters - Adding event context information

fjtirado:
Adding operation and initiator to TaskInstanceView

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9743)


fjtirado:
jenkins retest this

fjtirado:
apparently unrelated (cannot reproduce on local env), jenkins retest this

fjtirado:
Failed test seems unrelated (not able to reproduce locally), jenkins retest this

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/TaskInstanceView.java
```diff
@@ -70,8 +74,19 @@
     public TaskInstanceView() {        
     }
     
-    public TaskInstanceView(Task source) {
```
gmunozfe, line 73: Doesn't it break backward compatibility?

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/test/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitterTest.java
```diff
@@ -308,6 +308,90 @@ protected ExecutorService buildExecutorService() {
...
+                protected ExecutorService buildExecutorService() {
+                    return createExecutor(latch);
+                }
+
+            };
+            emitter.apply(views);
+            latch.await(5, TimeUnit.SECONDS);
+
+            // always close emitter to clean resources
+            emitter.close();
```
gmunozfe, line 386: Could it be moved to the finally clause?

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/test/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitterTest.java
```diff
@@ -308,6 +308,90 @@ protected ExecutorService buildExecutorService() {
...
+
+            assertThat(responseCollector).hasSize(1);
+            assertThat(responseCollector.get(0)).isEqualToNormalizingNewlines(expectedResult);
+        } finally {
+            System.clearProperty("org.jbpm.event.emitters.elasticsearch.ignoreNull");
+        }
+    }
+
```
gmunozfe, line 394: Could you include a test with initiator different than "unknown" and targetEntity not null?

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/TaskInstanceView.java
```diff
@@ -54,6 +55,9 @@
     private Long parentId;
     private String processId;
     private String containerId;
+    private Operation operation;
+    private String initiator;
+     private String targetEntity;
```
gmunozfe, line 60: check indentation :eyes:

gmunozfe (commented):
Great job @fjtirado !
Just a question about backward compatibility with TaskView constructor, and request for adding non-default value test.

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/TaskInstanceView.java
fjtirado: not really, TaskInstanceView is internal and this constructor was only used where it was modified

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/TaskInstanceView.java
fjtirado: thanks

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/test/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitterTest.java
fjtirado: sure

fjtirado:
unrelated, jenkins retest this

gmunozfe (approved):
Looks good to me, great job @fjtirado !
Let's wait jenkins to have it green

gmunozfe:
jenkins retest this

fjtirado:
jenkins retest this

afalhambra (approved):
Just a minor comment, otherwise looks awesome

fjtirado:
@gmunozfe @afalhambra Approach changed as per discussion with customer, please re review

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
unrelated jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
unrelated failed test will be fixed with https://github.com/kiegroup/jbpm/pull/1976
@afalhambra @krisv 
please review

fjtirado:
> ). Would it maybe make sense to have a public isActive() operation that would prevent creating the View objects in case the emitter is not active?
@krisv great point, code was modified to call isActive before actually building the object.

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/TaskOperationView.java
```diff
@@ -0,0 +1,72 @@
...
+
+import org.jbpm.persistence.api.integration.InstanceView;
+import org.kie.api.task.TaskEvent;
+import org.kie.api.task.TaskLifeCycleEventListener.AssignmentType;
+import org.kie.internal.task.api.model.TaskEvent.TaskEventType;
+
+
+public class TaskOperationView implements InstanceView<TaskEvent> {
+
+    private static final long serialVersionUID = 1L;
```
rbarriuso, line 26: Minor comment: Is this serial OK?

fjtirado: Since the class is new addition, using default one (1) is fine. In fact, being new, whatever number we use will be fine.

fjtirado:
jenkins retest this

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
```diff
@@ -70,64 +64,69 @@
...
                     }
-
-                    content.append("{ \"index\" : { \"_index\" : \"" + index + "\", \"_type\" : \"" + type + "\", \"_id\" : \"" + id + "\" } }\n");
-                    content.append(json);
-                    content.append("\n");
-
+                    ESRequest request = visitedIds.add(view.getCompositeId()) ? instanceView.index(view) : instanceView
+                            .update(view);
+                    content.append(String.format(
+                            "{ \"%s\" : { \"_index\" : \"%s\", \"_type\" : \"%s\", \"_id\" : \"%s\" } }%n%s%n", request
```
rbarriuso, line 127: As far as I know the `_type` field was [deprecated on Elasticsearch 6](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-type-field.html#mapping-type-field). Actually it's not used by the `_bulk` API (see [docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-request-body)).

Is it kept for backwards compatibility?

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
```diff
@@ -70,64 +64,69 @@
...
+                        logger.warn("Unsupported view {} ", view.getClass());
+                        continue;
                     }
-
-                    content.append("{ \"index\" : { \"_index\" : \"" + index + "\", \"_type\" : \"" + type + "\", \"_id\" : \"" + id + "\" } }\n");
-                    content.append(json);
-                    content.append("\n");
-
+                    ESRequest request = visitedIds.add(view.getCompositeId()) ? instanceView.index(view) : instanceView
+                            .update(view);
```
rbarriuso, line 125: Since `view` already contains all the event data, perhaps it's not necessary to differentiate between `index` and `update` operations. According to [the docs](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html#bulk-api-request-body), `index` updates the document if it already exists.

Wdyt?

fjtirado: The idea here is to merge the two views (task and operation) into the same index. 
The operation view does not really contain the same data than the task view, since eventTask source field is in fact transient. OperationView just contain type, userId and assingType (which will be null for all operations different that UPDATE list of users)

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
fjtirado: TBH I did not revisit that part of the emitter, but yes, it is better to keep it for backward compatibility

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
rbarriuso: OK I understand, thanks. Then I think the best approach would be sending `TaskInstanceView`s with the `index` operation and `TaskOperationView`s with `update`, right? Otherwise I can imagine that if a `TaskInstanceView` is sent with the `update` operation, it can leave fields out of sync (e.g. if some fields are removed)

fjtirado:
yes, thats what should be happening, index being used for task view and
update being used for tak operation view

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
fjtirado: yes, thats what should be happening, index being used for task view and
update being used for tak operation view

fjtirado:
unrelated test that will be fixed https://github.com/kiegroup/jbpm/pull/1976

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/test/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitterTest.java
```diff
@@ -308,6 +325,91 @@ protected ExecutorService buildExecutorService() {
...
+            when(peopleAssignments.getBusinessAdministrators()).thenReturn(Collections.emptyList());
+            when(peopleAssignments.getExcludedOwners()).thenReturn(Collections.emptyList());
+            when(peopleAssignments.getPotentialOwners()).thenReturn(Collections.emptyList());
+
+            when(user.getId()).thenReturn("john");
+
+            Map<String, Object> variables = new HashMap<>();
+            variables.put("initiator", "john");
+            variables.put("variable", 123);
```
afalhambra, line 353: There is no `null` value set in this test case to make sure the null values are ignored.

I would set for instance, this to `null` and remove this field from the expected output in the json file `testTaskInstanceThroughEmitterIgnoreNull.json`

afalhambra (changes requested):
Amazin job @fjtirado - just some minor comments, I think we should modify the ignoreNull test a little, but other than that it's an excellent work!

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
rbarriuso: Is it safe to assume that the task view will always come first and then the task operation view? Even if it's 100% sure, perhaps we could make it more explicit (e.g. `if(view instance of TaskOperationView) {...`). This way the code would be easier to understand.

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
```diff
@@ -920,5 +945,16 @@ protected String getActualOwner(Task ti) {
         
         return userId;
     }
+    
+    private void createView (TaskEvent event, TaskEventType type) {
+        createView (event, type, null);
+    }
+    
+    private void createView (TaskEvent event, TaskEventType type, AssignmentType assignType) {
```
rbarriuso, line 953: Minor comment: perhaps a name like `emitTaskOperationEvent()` or `generateTaskOperationEvent()`  or `sendTaskOperationView()` or similar would clarify the purpose of this method

rbarriuso: Actually since this method is always called after `persist()`, perhaps `persist()` can be included as part of the method and extend it's responsibility (with a new method name too)

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
```diff
@@ -71,9 +75,10 @@ public void afterTaskStartedEvent(TaskEvent event) {
         Task ti = event.getTask();
         TaskPersistenceContext persistenceContext = getPersistenceContext(((TaskContext)event.getTaskContext()).getPersistenceContext());
         try {
-            TaskEventImpl taskEventImpl = new TaskEventImpl(ti.getId(), org.kie.internal.task.api.model.TaskEvent.TaskEventType.STARTED, ti.getTaskData().getProcessInstanceId(), ti.getTaskData().getWorkItemId(), userId);
+            TaskEventImpl taskEventImpl = new TaskEventImpl(ti.getId(), TaskEventType.STARTED, ti.getTaskData().getProcessInstanceId(), ti.getTaskData().getWorkItemId(), userId);
             event.getMetadata().put(METADATA_TASK_EVENT, taskEventImpl);
             persistenceContext.persist(taskEventImpl);
+            createView(event, TaskEventType.STARTED);
```
rbarriuso, line 81: I've got some doubts about how this 2-step approach impacts `EventEmitter` implementations which aim to be transactional and synchronous (e.g. sending the data to Elastic in `deliver()` instead of in `apply()`).

I understand that if `persist()` fails (e.g. connectivity error throwing a `RuntimeException` when sending the `TaskInstanceView` to Elastic), the operation is rolled-back, the process instance doesn't change state and `createView()` isn't called. That's the expected behavior.

Nevertheless, if the error happens in `createView()`, the first `TaskInstanceView` would have already been sent to Elastic, and the process instance would have already moved on. Nevertheless the `TaskOperationView` will never be stored in Elastic, hence the event information may be lost.

Is this correct?

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
fjtirado: Yes it is safe to assume that. 
This code should not perform instanceof, since it will stop being generic

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
fjtirado: I did not want to change persist on purpose, the creation of the view is responsabiilty of the event listener, not the task context.
Regarding the name, I will change to createTaskOperationView (Im not fan of long names for methods, but it is ok)

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
fjtirado: your understanding is correct, but thats the way the emitter works. It is executed in a trasnsactional context, but the emission itself is decoupled from that context

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/InstanceView.java
```diff
@@ -39,4 +39,6 @@
      * PersistentEventManager will invoke it before delivering if it was not done before.
      */
     void copyFromSource();
+    
+    String getCompositeId();
```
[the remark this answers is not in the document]
fjtirado: Ok to the javadoc
Since @override is not the purpose of that JIRA, Im only adding to the classes that needs to be changed anyway, but not adding a new class to the modified set just because adding override.

jbpm-persistence/jbpm-persistence-api/src/main/java/org/jbpm/persistence/api/integration/model/TaskOperationView.java
```diff
@@ -0,0 +1,72 @@
...
+    private transient TaskEvent source;
+
+    private String compositeId;
+    private TaskEventType type;
+    private AssignmentType assignType;
+    private String userId;
+    
+    public TaskOperationView(TaskEvent taskEvent, TaskEventType type) {
```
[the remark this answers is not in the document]
fjtirado: Note that source is never null, which is null is assingType, but yes, that can be removed

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
```diff
@@ -902,13 +927,13 @@ protected void assignmentsUpadted(TaskEvent event, AssignmentType type, List<Org
...
         TaskEventImpl taskEventImpl = new TaskEventImpl(task.getId(),
-                                                        org.kie.internal.task.api.model.TaskEvent.TaskEventType.UPDATED,
+                                                        TaskEventType.UPDATED,
                                                         task.getTaskData().getProcessInstanceId(),
                                                         task.getTaskData().getWorkItemId(),
                                                         userId, message.toString());
         event.getMetadata().put(METADATA_TASK_EVENT, taskEventImpl);
         persistenceContext.persist(taskEventImpl);
-
+        createView (event, TaskEventType.UPDATED, type);
```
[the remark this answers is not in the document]
fjtirado: yes, good catch

afalhambra (approved):
awesome work! nice one

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins run fdb

fjtirado:
jenkins retest this

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
rbarriuso: Thanks @fjtirado . Indeed as you mentioned, I've tested that even if the 2 separate views are created, both are passed to the emitter's `deliver()` and `apply()` methods at the same time. Hence the emission is kept transactional (when using `deliver()` to send the data to Elasticsearch).

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/ElasticSearchEventEmitter.java
rbarriuso: I understand. Nevertheless during local testing I've found some unexpected behavior. For example, I implemented a test emitter with this method:

```java
@Override
public void deliver(Collection<InstanceView<?>> data) {
    logger.debug("Preparing bulk delivery of {} instance views: {}", data.size(),
            data.stream().map(v -> v.getClass().getSimpleName()).collect(Collectors.joining(", ")));
    for (InstanceView<?> view : data) {
        if (!(view instanceof TaskInstanceView) || !(view instanceof ProcessInstanceView)) {
            logger.warn("Unsupported view class {}: {}", view.getClass().getSimpleName(), mapper.writeValueAsString(view));
        }
    }
}
```

I deployed the "evaluation" process that comes with the examples and started a process instance. Surprisingly this is the log the emitter produced:

```
Preparing bulk delivery of 4 instance views: TaskOperationView, ProcessInstanceView, TaskInstanceView, TaskOperationView
Unsupported view class TaskOperationView: {"compositeId":"jbpm-development-server_15","type":"ACTIVATED","assignType":null,"date":"2021-07-28T18:36:07.516+0000","userId":"jbpmAdmin"}
Unsupported view class TaskOperationView: {"compositeId":"jbpm-development-server_15","type":"ADDED","assignType":null,"date":"2021-07-28T18:36:07.510+0000","userId":"jbpmAdmin"}
```

In this case the first view received in the collection is a `TaskOperationView` instead of a `TaskInstanceView` (in case order mattered) and both `TaskOperationView` instances generated have a `compositeId`. Is this expected?

fjtirado:
jenkins retest this
@elguardian Rafael has validated the patch, can you please review?

fjtirado:
jenkins run fdb

fjtirado:
jenkins run fdb

fjtirado:
jenkins retest this

fjtirado:
jenkins full fdb

fjtirado:
jenkins retest this

fjtirado:
jenkins run fdb

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/DefaultESInstanceViewTransformerFactory.java
```diff
@@ -0,0 +1,64 @@
...
+    
+    protected Map<Class<?>, ESInstanceViewTransformer> viewMapping = new ConcurrentHashMap<>();
+     
+    public DefaultESInstanceViewTransformerFactory() {
+        DefaultESInstanceViewTransformer taskView = new DefaultESInstanceViewTransformer("tasks", "task");
+        viewMapping.put(TaskOperationView.class, taskView);
+        viewMapping.put(TaskInstanceView.class, taskView);
+        viewMapping.put(ProcessInstanceView.class, new DefaultESInstanceViewTransformer("processes", "process"));
+        viewMapping.put(CaseInstanceView.class, new DefaultESInstanceViewTransformer("cases", "case"));
+    }
```
elguardian, line 47: I might missing the logic somewhere but just in case to double check:
The logic I cannot find is to load any other views the user might create for himself.
lookup for ESInstanceViewTransformer and set it class to itself. that way we allow users to do their own. Now that we created a proper framework for this we can allo the user to do it by himself

elguardian:
@fjtirado  Looks good. Just please review my comment and do if it is really missed out.

jbpm-event-emitters/jbpm-event-emitters-elasticsearch/src/main/java/org/jbpm/event/emitters/elasticsearch/DefaultESInstanceViewTransformerFactory.java
fjtirado: the idea I have in mind is that users should provide (or extend this default implementation) and specify this system property `org.jbpm.event.emitters.elasticsearch.factory` to replace default
https://github.com/kiegroup/jbpm/pull/1932/files#diff-204ae3aefeb9d9ef9abd24afbfa15e08b53048301b8b220646008791630432d7R73-R83

elguardian (approved):
LGTM. basically what it was discussed in june
https://issues.redhat.com/browse/BAPL-1932?focusedCommentId=16319237&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-16319237

elguardian:
failure not related.

PR #1933, opened 2021-05-21
[JBPM-9748] Avoid count check in hibernate for audit table.

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9748


afalhambra (approved):
Great job!

PR #1945, opened 2021-06-02
[JBPM-9767] Process instance does not become Completed state even though all nodes in the instance are completed.

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9767

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/async/AsyncOnErrorSignalCommandCallback.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/async/AsyncOnErrorSignalCommandCallback.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.jbpm.process.core.async;
+
+import java.io.Serializable;
+
+import org.jbpm.workflow.instance.impl.WorkflowProcessInstanceImpl;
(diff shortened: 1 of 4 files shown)


jbpm-test-coverage/src/test/java/org/jbpm/test/functional/async/TriggerAsyncNodeTest.java
```diff
@@ -0,0 +1,79 @@
...
+    private ExecutorService executorService;
+
+    @Before
+    @Override
+    public void setUp() throws Exception {
+        super.setUp();
+
+        executorService = ExecutorServiceFactory.newExecutorService(getEmf());
+        executorService.setInterval(1);
+        executorService.setRetries(0);
```
gmunozfe, line 49: One question here: if we set up the retries to more than one, `handler.onCommandError` is not executed yet (though there are more retries to try) but there's no retryDelay in the context, so probably won't be fired as the scheduled time is the current time, is it fine? Then, process is completed and no execution of the new command is invoked. Perhaps an additional test with these retries scenarios would be valuable, wdyt?

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/async/TriggerAsyncNodeTest.java
```diff
@@ -0,0 +1,79 @@
...
+        executorService.init();
+        addEnvironmentEntry("ExecutorService", executorService);
+        addWorkItemHandler("Rest", new FirstErrorWorkItemHandler());
+        addProcessEventListener(new DefaultProcessEventListener() {
+            @Override
+            public void afterProcessCompleted(ProcessCompletedEvent event) {
+                System.out.println("PROCESS COMPLETED");
```
gmunozfe, line 56: You could use a logger instead, but up to you

jbpm-flow/src/main/java/org/jbpm/process/core/async/AsyncOnErrorSignalCommandCallback.java
```diff
@@ -0,0 +1,89 @@
...
+    private static final long serialVersionUID = 7491132555750784258L;
+
+    @Override
+    public void onCommandDone(CommandContext ctx, ExecutionResults results) {
+        // do nothing
+    }
+
+    @Override
+    public void onCommandError(CommandContext ctx, Throwable exception) {
+        if(ctx.getData("NodeInstanceId") == null) {
```
gmunozfe, line 45: add blank before parenthesys

gmunozfe (commented):
Just minor comments and a question related to retries scenarios, good job @elguardian !

PR #1955, opened 2021-06-18
[JBPM-9786] Add REST Endpoint for claiming tasks in bulk

fjtirado:
Bulk method implementation.
I followed the same approach that was used for other bulk methods,
meaning that any not id found exception is ignored.

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9786)

**referenced Pull Requests**: 
https://github.com/kiegroup/droolsjbpm-integration/pull/2518

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
@@ -17,6 +17,7 @@
 package org.jbpm.kie.services.impl;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
@@ -192,6 +193,17 @@ public class UserTaskServiceImpl implements UserTaskService, VariablesAware {
 		}
 
 	}
+
+    @Override
+    public void claim(String deploymentId, Collection<Long> taskIds, String userId) {
+        taskIds.forEach(taskId -> {
+            try {
+                claim(deploymentId, taskId, userId);
+            } catch (TaskNotFoundException | PermissionDeniedException e) {
+                logger.warn("Problem claiming task id {}", taskId, e);
+            }
+        });
+    }
 
 	@Override
     public void complete(Long taskId, String userId, Map<String, Object> params) {
@@ -719,7 +731,7 @@ public class UserTaskServiceImpl implements UserTaskService, VariablesAware {
 		try {
 			TaskService taskService = engine.getTaskService();
 			// perform actual operation
-            return ((InternalTaskService) taskService).addContentFromUser(taskId, userId, (Map<String, Object>) values);
+            return ((InternalTaskService) taskService).addContentFromUser(taskId, userId, values);
 		} finally {
 			disposeRuntimeEngine(manager, engine);
 		}
--- a/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskServiceImplTest.java
+++ b/jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskServiceImplTest.java
@@ -20,6 +20,7 @@ import java.io.File;
 import java.io.FileOutputStream;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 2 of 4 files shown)


fjtirado:
jenkins retest this

Ginxo:
jenkins retest this

elguardian (changes requested):
create a ClaimBulkCommand based on that ClaimTaskCommand

fjtirado:
> create a ClaimBulkCommand based on that ClaimTaskCommand

I have not problem on doing that, but it will mean we are taking a different desing than other bulk operations, which just call the individual operation several times

elguardian (approved):
LGTM

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
```diff
@@ -192,6 +193,17 @@ public void claim(String deploymentId, Long taskId, String userId) {
 		}
 
 	}
+	
+	@Override
+    public void claim(String deploymentId, Collection<Long> taskIds, String userId) {
+        taskIds.forEach(taskId -> {
+            try {
+                claim(deploymentId, taskId, userId);
+            } catch (TaskNotFoundException e) {
```
gmunozfe, line 202: if the task is not in the expected status to be claimed, it could also throw other exceptions like `PermissionDeniedException`. In that case, it wouldn't continue with the loop.
You could also add a test case for this scenario, by claiming a "Reserved" task.

gmunozfe (commented):
In the loop, only TaskNotFoundException is caught. In some scenarios -like tasks already in "Reserved" status or others that are not allowed to be claimed- should fail with another exception breaking the loop.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
fjtirado: But I think thats right, in that case we can cancel the the operation. @elguardian wdyt?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
```diff
@@ -192,6 +193,17 @@ public void claim(String deploymentId, Long taskId, String userId) {
...
+                logger.warn("Cannot find task id {}", taskId, e);
+            }
+        });
+    }
```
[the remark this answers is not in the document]
fjtirado: I fix the indented comment, but there is a mix of formatting in this file that cannot be fixed wihtout polluting the diff

jbpm-services/jbpm-services-api/src/main/java/org/jbpm/services/api/UserTaskService.java
```diff
@@ -60,6 +61,15 @@
...
 	void claim(Long taskId, String userId);
 	
+	/**
+     * Claim responsibility for a list of tasks, i.e. set the tasks to status Reserved
+     * 
+     * @param deploymentId
+     * @param taskIds list of task to be claimed 
+     * @param userId
+     */
+    void claim(String deploymentId, Collection<Long> taskIds, String userId);
```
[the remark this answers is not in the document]
fjtirado: I fix the indented comment, but there is a mix of formatting in this file that cannot be fixed wihtout polluting the diff

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
fjtirado: hmmm, I think that since this is launching individual command, we can catch `PermissionDeniedException` and log it. Any other exception you can think of @gmunozfe ?

gmunozfe: my understanding was that bulk operations are best-effort, trying to complete as many individual operations as possible

gmunozfe: @fjtirado The other exception I can see is `IllegalStateException`: "Task instance ... is owned by another deployment expected "

fjtirado: hmmmm, problem is that illegalstate can hide a different issue, and in any case it will a gross failure and I think we should report

gmunozfe (approved):
Looks good to me, astounding work @fjtirado !

afalhambra (approved):
1 minor comment on the tests, but just optional.
nice job!

jbpm-services/jbpm-kie-services/src/test/java/org/jbpm/kie/services/test/UserTaskServiceImplTest.java
```diff
@@ -186,6 +187,28 @@ public void testReleaseAndClaim() {
     	assertEquals(Status.Reserved.toString(), task.getStatus());
     }
     
+    
+    @Test
+    public void testReleaseAndBulkClaim() {
```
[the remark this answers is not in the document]
fjtirado: I prefer to keep the test simple, the code is trivial so I do not see need for that

PR #1963, opened 2021-06-26
JBPM-9794: Locking issue with aborted parent process instance

krisv:
https://issues.redhat.com/browse/JBPM-9794


elguardian:
no way around it. Very old bug.
https://github.com/kiegroup/jbpm/commit/59e55a6ee629ef8c01be43d7b2565d2ca7a2412e

Right approach IMO.

PR #1964, opened 2021-06-28
[ENG] [7.52.x] [JBPM-9794] Locking issue with aborted parent process instance

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9794
backport

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
@@ -424,20 +424,24 @@ public abstract class WorkflowProcessInstanceImpl extends ProcessInstanceImpl
             if (isSignalCompletion()) {
                 RuntimeManager manager = (RuntimeManager) kruntime.getEnvironment().get(EnvironmentName.RUNTIME_MANAGER);
                 if (getParentProcessInstanceId() > 0 && manager != null) {
-                	try {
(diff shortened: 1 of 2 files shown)


jbpm-flow/src/main/java/org/jbpm/workflow/instance/impl/WorkflowProcessInstanceImpl.java
```diff
@@ -424,20 +424,22 @@ public void setState(final int state, String outcome) {
...
+                        context = CaseContext.get(caseId);
+                    }
 
-    	                RuntimeEngine runtime = manager.getRuntimeEngine(context);
-						KieRuntime managedkruntime = runtime.getKieSession();
-    	                managedkruntime.signalEvent("processInstanceCompleted:" + getId(), this);
-                	} catch (SessionNotFoundException e) {
-                		// in case no session is found for parent process let's skip signal for process instance completion
-                	}
+	                RuntimeEngine runtime = manager.getRuntimeEngine(context);
```
krisv, line 434: @elguardian you've removed the try (line 427), but I think we might still need it in case the engine is executed with eager initialization?  In that case the exception will be thrown during manager.getRuntimeEngine(..) I believe, but that would still need to be caught and ignored (although calling dispose will not be necessary in that case as there won't be any locks acquired yet)?

elguardian: no need as is eager it won't get the lock in the first place. so that situation won't happen

elguardian: ok you are right

PR #1969, opened 2021-07-01
[JBPM-9804] Protect AbstractDeploymentService against unexpected exception in listeners

fjtirado:
**JIRA**:

[link](https://issues.redhat.com/browse/JBPM-9804)


elguardian:
wrong direction. listener cannot contribute to deployments so expecting failures is not a good idea. Clients of the listeners should no throw exceptions.

fjtirado:
> wrong direction. listener cannot contribute to deployments so expecting failures is not a good idea. Clients of the listeners should no throw exceptions.

I disagree, any unexpected exception in the listener will hide the original excepiton, making nearly imposible for the customer implementing the listener to diagnose the issue, as we are suffering with the kafka listener.

fjtirado:
please jenkins test this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

fjtirado:
jenkins retest this

krisv:
Exceptions in the catch section that could hide the original exception could be problematic, in all honesty I don't really care how it is handled, if we don't prefer using another try-catch, maybe just logging the error immediately after the catch (possibly even at debug level if we don't want to show it by default) would help in this case?

PR #1970, opened 2021-07-01
[JBPM-9803] Remove audit log date archived from queries automatically

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9803


elguardian:
@gmunozfe added the scripts for removing the column.
Regarding your question about removing the spi, that is not possible. This is based on the assumption that after version is released we cannot deliver a funcionality but we can let the customer implement the spi to activate the functionality.

PR #1995, opened 2021-08-10
[JBPM-9849] Remove default behavior for end_date

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9849

### DIFF
--- a/jbpm-audit/src/main/java/org/jbpm/process/audit/AuditLoggerArchiveTreat.java
+++ b/jbpm-audit/src/main/java/org/jbpm/process/audit/AuditLoggerArchiveTreat.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2021 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


jbpm-audit/src/main/java/org/jbpm/process/audit/AuditLoggerArchiveTreat.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+import java.util.ServiceLoader;
+
+public interface AuditLoggerArchiveTreat {
+
+    default List<ArchiveLoggerProvider> initArchiveLoggerProvider() {
+        List<String> props = Arrays.asList(System.getProperty("org.kie.jbpm.persistence.archive-provider", "").split(","));
+        List<ArchiveLoggerProvider> archiveLoggerProvider = new ArrayList<>();
+        Iterator<ArchiveLoggerProvider> iterator = ServiceLoader.load(ArchiveLoggerProvider.class).iterator();
+        while(iterator.hasNext()) {
```
gmunozfe, line 31: missing a blank
```suggestion
        while (iterator.hasNext()) {
```

jbpm-audit/src/main/java/org/jbpm/process/audit/AuditLoggerArchiveTreat.java
```diff
@@ -0,0 +1,40 @@
...
+            ArchiveLoggerProvider provider = iterator.next();
+            if(props.contains(provider.getClass().getCanonicalName())) {
```
gmunozfe, line 33: Same
```suggestion
            if (props.contains(provider.getClass().getCanonicalName())) {
```

gmunozfe (commented):
Looks good to me, good work @elguardian !
My only concern is that we are changing the current default behavior, and those expecting the same result from previous releases have to add the property. Could we add an additional property and keep previous behavior out of the box, and make this new one configurable? wdyt?

PR #1997, opened 2021-08-12
[JBPM-9858] sql server scripts use identities instead of sequences for spring boot

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9858


gmunozfe (approved):
Looks good to me, awesome work @elguardian !
It works like a charm

PR #1998, opened 2021-08-13
fixing OptimisticLockException in write-mode after readonly access

fomich-artem:
When sequentially accessing ProcessInstance within one transaction, first in readonly access mode, then in write access mode, the cached ProcessInstanceInfo with the old value of the version attribute remains in the EntityManager. This results in an OptimisticLockException "Row was updated or deleted by another transaction" when concurrently accessing ProcessInstance, as well as some other problems.

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
@@ -23,9 +23,12 @@ import java.util.ConcurrentModificationException;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 
+import javax.persistence.EntityManager;
+
 import org.drools.core.common.InternalKnowledgeRuntime;
 import org.drools.persistence.api.TransactionManager;
 import org.drools.persistence.api.TransactionManagerHelper;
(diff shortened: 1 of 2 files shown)


fomich-artem:
Greetings!
Please, somebody review my pull-request :)

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPAProcessInstanceManager.java
```diff
@@ -172,6 +175,11 @@ public ProcessInstance getProcessInstance(long id, boolean readOnly) {
 
             if (readOnly) {
                 internalRemoveProcessInstance(processInstance);
+                if (ppcm instanceof JpaProcessPersistenceContextManager) { // for MapBasedPersistenceTest
```
fjtirado, line 178: In order to avoid the instanceof here, I would add a new default method to ProcessPersistenceContextManager with signature checkProcessInstanceInfo (ProcessInstanceInfo processInstanceInfo, boolean readOnly). 
The other alternative is to modify getProcessPersistenceContext signature adding boolean readOnly parameter and do the detach there (but this will imply that both this methods and getPRocessPersistenceContext has to be invoke findProcessInstanteInfo in the same thread)

elguardian:
thx @fomich-artem . this is a great catch and very good work. I superseeded this in order to accomodate to the current set of interfaces. 
https://github.com/kiegroup/jbpm/pull/2130

PR #2012, opened 2021-09-06
[JBPM-9861] Mvel expressions should fire variable change event

fjtirado:
Variable change event was not triggered when evaluating mvel expression

**JIRA**:

[link](https://issues.redhat.com/browse/JBPM-9861)

### DIFF
--- a/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/DataTest.java
+++ b/jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/DataTest.java
@@ -16,6 +16,12 @@
 
 package org.jbpm.bpmn2;
 
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+
 import java.io.ByteArrayInputStream;
 import java.io.Serializable;
 import java.util.Arrays;
@@ -24,11 +30,13 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 
 import javax.xml.parsers.DocumentBuilder;
 import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
 
+import org.drools.core.event.DefaultProcessEventListener;
 import org.jbpm.bpmn2.core.Association;
 import org.jbpm.bpmn2.core.DataStore;
 import org.jbpm.bpmn2.core.Definitions;
@@ -43,6 +51,7 @@ import org.junit.runner.RunWith;
 import org.junit.runners.Parameterized;
 import org.junit.runners.Parameterized.Parameters;
 import org.kie.api.KieBase;
+import org.kie.api.event.process.ProcessVariableChangedEvent;
 import org.kie.api.runtime.process.ProcessInstance;
 import org.kie.api.runtime.process.WorkItem;
 import org.kie.api.runtime.process.WorkItemHandler;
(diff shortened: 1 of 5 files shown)


jbpm-flow-builder/src/main/java/org/jbpm/process/builder/dialect/mvel/MVELResolverFactory.java
```diff
@@ -0,0 +1,93 @@
...
+
+        @Override
+        public int getFlags() {
+            return resolver.getFlags();
+        }
+
+        @Override
+        public Object getValue() {
+            if (clonedValue == null) {
+                clonedValue = ObjectCloner.clone(resolver.getValue(), new Config().deepCloneCollections(!Boolean.getBoolean("org.kie.jbpm.mvel.notCloneCollections")));
```
gmunozfe, line 83: You could extract the property in a constant for better readibility

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ObjectCloner.java
```diff
@@ -0,0 +1,479 @@
...
+
+    private static Object cloneWithConstructor(Object object, Class<?> objectClass) {
+        return constructorExecutorMap.computeIfAbsent(objectClass, ObjectCloner::getConstructorExecutor).apply(object);
+    }
+
+    private static UnaryOperator<Object> getConstructorExecutor(Class<?> objectClass) {
+        ConstructorInfo constructorInfo = new ConstructorInfo(objectClass);
+        if (constructorInfo.getCopyConstructor() != null) {
+            return new CopyConstructorExecutor(constructorInfo.getCopyConstructor());
+        } else {
```
gmunozfe, line 195: this "else" is not strictly necessary if returns in the "if" (one indentation level can be removed)

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ObjectCloner.java
```diff
@@ -0,0 +1,479 @@
...
+            Collection<ArgResolver> args = new ArrayList<>();
+            boolean found = true;
+            for (int i = 0; found && i < parameters.length; i++) {
+                ArgResolver arg = getArg(parameters[i], methodInfo);
+                found = arg != null;
+                if (found) {
+                    args.add(arg);
+                }
+            }
+            if (found) {
```
gmunozfe, line 225: if the for-loop above is partially executed until one null argument, then found will be "false" but args will contain the previous not null argument, but will return default constructor, is it right?

jbpm-flow/src/test/java/org/jbpm/process/core/impl/ObjectClonerTest.java
```diff
@@ -0,0 +1,414 @@
...
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+
+import org.jbpm.process.core.impl.ObjectCloner.Config;
+import org.junit.Test;
+
+public class ObjectClonerTest {
```
gmunozfe, line 35: you could add also some tests for rainy scenarios and exceptions like ReflectiveOperationException, wdyt?

gmunozfe (approved):
Looks good to me, such an incredible job @fjtirado!

Some minor comments to discuss.

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ObjectCloner.java
fjtirado: those previusly calculated args are ignored since the next enclosing loop iteration (search for another suitable constructor) is performed.

jbpm-flow/src/test/java/org/jbpm/process/core/impl/ObjectClonerTest.java
fjtirado: hmmm, I do no think is a case where this exception is going to be throwb, since this is reflection inside the same object and Im using public methods (Im catching because the compiles enforces me to do so)

jbpm-flow/src/main/java/org/jbpm/process/core/impl/ObjectCloner.java
fjtirado: ok, code rearranged to avoid too many returns in same method

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/dialect/mvel/MVELResolverFactory.java
fjtirado: I want the property to be read every time (so you do not need to restart if the property is changed) 
I put the propery name as constant

jbpm-flow-builder/src/main/java/org/jbpm/process/builder/dialect/mvel/MVELAssignmentAction.java
```diff
@@ -57,24 +62,45 @@ public MVELAssignmentAction(Assignment assignment, String sourceExpr, String tar
...
+            if (targetExpr != null) {
+                producer.accept(context, nodeInstance, targetObject);
+            }
         } else {
+            String lValue = ensureLocated(targetExpr, to);
+            String rootName; 
+            if (targetExpr != null) {
+                rootName = targetExpr;
+            }
+            else {
```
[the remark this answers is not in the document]
fjtirado: changed to tertiary expression

jbpm-bpmn2/src/test/java/org/jbpm/bpmn2/DataTest.java
```diff
@@ -815,6 +909,8 @@ public void executeWorkItem(WorkItem workItem,
                         mgr.completeWorkItem(workItem.getId(), Collections.singletonMap("output", Arrays.asList(
                                 new Person("Javierito", new Address("Paris", "France")), new Person("Napoleon",
                                         new Address("Sevilla", "Spain")))));
+                        person = (Map<String, Object>) processInstance.getVariable(
+                                "instanceMetadata");
```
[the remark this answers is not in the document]
fjtirado: I thiink we are abusing assertNotNull in the test. Im not removing them where they are already present, but I do not want to add more

gmunozfe (approved):

fjtirado:
jenkins retest this

PR #2020, opened 2021-09-14
[7.59.x][JBPM-9895] Fix flaky tests at LocalTaskAuditWithDeadlineTest

gmunozfe:
**JIRA**: [JBPM-9895](https://issues.redhat.com/browse/JBPM-9895)


elguardian (approved):
This is just bypassing the race condition IMO, but it is ok for now.

gmunozfe:
> This is just bypassing the race condition IMO, but it is ok for now.

Yes, the point is if reassign time has expired at that point (because of slow execution), these assertions will fail

gmunozfe:
@mareknovotny ready to merge

PR #2039, opened 2021-10-04
[JBPM-9914] Eagerly initialize emitters

fjtirado:

[link](https://issues.redhat.com/browse/JBPM-9914)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -129,6 +131,9 @@ public AbstractRuntimeManager(RuntimeEnvironment environment, String identifier)
         if (eagerInit != null) {
         	engineInitEager = Boolean.parseBoolean(eagerInit);
         }
+        if (engineInitEager) {
```
fjtirado, line 134: Im reusing existing variable, but maybe is better to use a new one, wdyt?

elguardian: ok. There is a problem. Nobody uses that flag and that one is a bit dangerous. Better a new one.

fjtirado: ok changed

fjtirado:
@gmunozfe I need your help to see how to check this in integration

PR #2047, opened 2021-10-14
[JBPM-9926] Auto ack of errors in process doesn't work on Sybase

gmunozfe:
**JIRA**: [JBPM-9926](https://issues.redhat.com/browse/JBPM-9926)


mareknovotny:
don't we need to run tests in droolsjbpm-integration repository?

PR #2056, opened 2021-11-08
JBPM-9924- Define a default value of the process variable at design time

VaniHaripriya:
JBPM-9924- https://issues.redhat.com/browse/JBPM-9924


jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -173,6 +173,7 @@ public void enforceRequiredVariables() {
         VariableScope variableScope = getVariableScope();
         for (Variable variable : variableScope.getVariables()) {
             String name = variable.getName();
+			Object defaultValue = variable.getMetaData("defaultValue");
```
elguardian, line 176: this will cause the check to be triggered before tha default value is set, therefore causing an error when default value is provided (which is not coherent)

elguardian:
please keep in mind that the proper place to do this is not here but the process instance factory where parameters are set and not here.

elguardian (changes requested):
this requires some rework

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
elguardian, line 176: removed

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -186,6 +187,9 @@ public void enforceRequiredVariables() {
                 }
                 
             }
+			if (defaultValue != null) {
```
elguardian, line 190: this needs to be removed.

jbpm-flow/src/main/java/org/jbpm/process/instance/AbstractProcessInstanceFactory.java
```diff
@@ -52,6 +54,16 @@ public ProcessInstance createProcessInstance(Process process, CorrelationKey cor
         }
         // TODO: should be part of processInstanceImpl?
         VariableScope variableScope = (VariableScope) ((ContextContainer) process).getDefaultContext( VariableScope.VARIABLE_SCOPE );
+		if (variableScope != null) {
+			for (Variable variable : variableScope.getVariables()) {
+				DataType dataType = variable.getType();
+				Object defaultValue = variable.getMetaData("defaultValue");
+				if (defaultValue != null && defaultValue instanceof String && variable.getValue() == null) {
+					defaultValue = dataType.readValue((String) defaultValue);
+					variable.setValue(defaultValue);
```
elguardian, line 63: the variable scope is the "model" ... yo need to set the "instance"

if(varaibleScopeInstance.getVariable() == null) {
   varaibleScopeInstance.setVariable(name, defaultValue)
}

elguardian (changes requested):

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
```diff
@@ -171,22 +171,25 @@ public void setContextInstanceContainer(ContextInstanceContainer contextInstance
...
-                        
-                    }
-                    // otherwise check variables                    
-                } else if (!hasData(variables.get(name))) {
-                    throw new VariableViolationException(getProcessInstance().getId(), name, "Variable '" + name + "' is required but not set");
-                }
-                
-            }
-        }
+		for (Variable variable : variableScope.getVariables()) {
```
elguardian, line 174: try to avoid unnecessary changes like this one. otherwise PR gets difficult to track.

elguardian (approved):
LGTM waiting QE to check on this.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -162,4 +166,37 @@ public void testMultiValuedKeyUniqueButInclusive() {
         ProcessInstance processInstance2 = ksession.startProcess(PROCESS, key2, null);
         assertProcessInstanceActive(processInstance2.getId());
     }
+    
+    @Test
+    public void testCreateProcessInstanceDefaultValue() {
+        CorrelationKey key = keyFactory.newCorrelationKey(SIMPLE_KEY);
+
+        Map<String, Object> parameters = new HashMap<String, Object>();
```
fjtirado, line 174: Collections.emptyMap()

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -162,4 +166,37 @@ public void testMultiValuedKeyUniqueButInclusive() {
...
+    @Test
+    public void testCreateProcessInstanceComplexTypeDefaultValue() {
+       	String address = "<org.jbpm.test.domain.Address>\n"
+    			+ "          	   <street>abc</street>\n"
+    			+ "						   <number>29</number>\n"
+    			+ "						   <city>def</city>\n"
+    			+ "						</org.jbpm.test.domain.Address>";
+        CorrelationKey key = keyFactory.newCorrelationKey(SIMPLE_KEY);
+
+        Map<String, Object> parameters = new HashMap<String, Object>();
```
fjtirado, line 192: Collections.emptyMap

fjtirado (commented):
How about a test checking that if variable value is provided as parameter and default value is provided in bpmn file, value provided as parameter is taken. 
Also, this is for @elguardian, should default vale trigger a variable value changed event? My guess is yes, but Im unsure

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -162,4 +166,37 @@ public void testMultiValuedKeyUniqueButInclusive() {
...
+        assertEquals(Integer.valueOf(1), processInstance.getVariable("secondVar"));
+        
+    }    
+    
+    @Test
+    public void testCreateProcessInstanceComplexTypeDefaultValue() {
+       	String address = "<org.jbpm.test.domain.Address>\n"
+    			+ "          	   <street>abc</street>\n"
+    			+ "						   <number>29</number>\n"
+    			+ "						   <city>def</city>\n"
```
gmunozfe, line 188: Test failing because blanks shouldn't be present after tabs.
Anyway, I think this will also fail in windows due to a different separator. Better to use "xmlunit" library for this kind of comparison.

gmunozfe (changes requested):

jbpm-flow/src/main/java/org/jbpm/process/instance/AbstractProcessInstanceFactory.java
```diff
@@ -20,7 +20,9 @@
 
 import org.drools.core.common.InternalKnowledgeRuntime;
 import org.jbpm.process.core.ContextContainer;
+import org.jbpm.process.core.context.variable.Variable;
 import org.jbpm.process.core.context.variable.VariableScope;
+import org.jbpm.process.core.datatype.DataType;
```
gmunozfe, line 25: This is not used, please remove

jbpm-flow/src/main/java/org/jbpm/process/instance/AbstractProcessInstanceFactory.java
```diff
@@ -64,6 +66,17 @@ public ProcessInstance createProcessInstance(Process process, CorrelationKey cor
...
+			for (Variable variable : variableScope.getVariables()) {
+				String name = variable.getName();
+				Object defaultValue = variable.getMetaData("defaultValue");
+				if (variableScopeInstance.getVariable(name) == null && defaultValue != null) {
+					variableScopeInstance.setVariable(name,
+							variableScope.validateVariable(process.getName(), name, defaultValue));
+
+				}
+			}
+		}
```
gmunozfe, line 79: Check indentation, avoid tabs in favour of spaces

elguardian:
@fjtirado setting the variables at the beginning will set the same triggers as provided by the end user. so that should be ok.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
VaniHaripriya: Sure, updated the test cases accordingly.

jbpm-flow/src/main/java/org/jbpm/process/instance/context/variable/VariableScopeInstance.java
gmunozfe: @VaniHaripriya could you undo these indentation changes as Enrique suggested?

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -162,4 +168,55 @@ public void testMultiValuedKeyUniqueButInclusive() {
...
+        assertEquals(Integer.valueOf(1), processInstance.getVariable("secondVar"));
+
+    }
+
+    @Test
+    public void testCreateProcessInstanceComplexTypeDefaultValue() throws SAXException, IOException {
+
+        String address = "<org.jbpm.test.domain.Address>\n" + "          	 <street>abc</street>\n"
+                + "						   <number>29</number>\n" + "						   <city>def</city>\n"
+                + "						</org.jbpm.test.domain.Address>";
```
gmunozfe, line 206: I think you can write this without spaces and carriage returns (that's why xmlunit is useful)

gmunozfe (changes requested):

gmunozfe:
jenkins retest this

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -162,4 +168,53 @@ public void testMultiValuedKeyUniqueButInclusive() {
...
+
+        CorrelationKey key = keyFactory.newCorrelationKey(SIMPLE_KEY);
+
+        Map<String, Object> parameters = new HashMap<String, Object>();
+
+        WorkflowProcessInstance processInstance = (WorkflowProcessInstance) ksession.startProcess(PROCESS, key,
+                parameters);
+
+        DetailedDiff diff = new DetailedDiff(
+                XMLUnit.compareXML(address, (String) processInstance.getVariable("complexVar")));
```
gmunozfe, line 214: It's not clear to me that this testing is working properly in the sense of returned variable is an instance of String. I was expecting to get a custom org.jbpm.test.domain.Address. For example, if the parameters map sets up a "complexVar" with an Address value, it's obtained again an object from this type, wdyt?

gmunozfe (changes requested):
There are test failures due to readValue invocation for some types:
```
at org.jbpm.process.core.datatype.impl.type.IntegerDataType.readValue(IntegerDataType.java:55)
at org.jbpm.process.core.context.variable.VariableScope.validateVariable(VariableScope.java:111)
```

gmunozfe:
Failures are unrelated, let's try again.
Jenkins retest this

jbpm-test-coverage/pom.xml
```diff
@@ -125,6 +125,11 @@
         </exclusion>
       </exclusions>
     </dependency>
+    <dependency>
+      <groupId>xmlunit</groupId>
+      <artifactId>xmlunit</artifactId>      
+      <scope>test</scope>
+    </dependency>
```
gmunozfe, line 132: this is not longer used, you can remove it

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -16,26 +16,35 @@
...
 package org.jbpm.test.functional;
 
+import static org.junit.Assert.assertEquals;
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
+import org.custommonkey.xmlunit.XMLUnit;
```
gmunozfe, line 26: Same

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -16,26 +16,35 @@
...
 import org.assertj.core.api.Assertions;
+import org.custommonkey.xmlunit.DetailedDiff;
```
gmunozfe, line 28: Same

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -16,26 +16,35 @@
...
+import org.jbpm.test.domain.Address;
 import org.junit.Test;
 import org.kie.api.runtime.manager.audit.VariableInstanceLog;
 import org.kie.api.runtime.process.ProcessInstance;
+import org.kie.api.runtime.process.WorkflowProcessInstance;
 import org.kie.internal.KieInternalServices;
 import org.kie.internal.process.CorrelationAwareProcessRuntime;
 import org.kie.internal.process.CorrelationKey;
 import org.kie.internal.process.CorrelationKeyFactory;
+import org.xml.sax.SAXException;
```
gmunozfe, line 39: Same

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/CorrelationKeyTest.java
```diff
@@ -46,7 +55,7 @@
     @Override
     public void setUp() throws Exception {
         super.setUp();
-
+        XMLUnit.setIgnoreWhitespace(true);
```
gmunozfe, line 58: Same here

gmunozfe (approved):
Looks good to me, great work @VaniHaripriya !
Just a few comments regarding non-used dependencies, if you want to delete them.

jbpm-flow/src/main/java/org/jbpm/process/core/context/variable/VariableScope.java
```diff
@@ -102,14 +102,17 @@ public Object validateVariable(String processName, String name, Object value) {
...
         } else if (value != null) {
             DataType type = var.getType();
+            ObjectDataType obj = new ObjectDataType();
             boolean isRightType = type == null || type.verifyDataType(value);
             if (!isRightType ) {
                 if (variableStrictEnabled) {
                     throw new IllegalArgumentException("Variable '" + name + "' has incorrect data type expected:" + var
                             .getType().getStringType() + " actual:" + value.getClass().getName());
-                } else if (value instanceof String) {
+                } else if (value instanceof String && (type.getClass() != obj.getClass())) {
```
gmunozfe, line 111: Probably you can write it more clear like this
```suggestion
                } else if (value instanceof String) {
                    if (type.getClass() == obj.getClass()) {
                        value = type.readValue(value.toString());
                    else {
                        value = type.valueOf(value.toString());
                    }
                }
```

gmunozfe (commented):
Also a suggestion for better readibility

jbpm-test-coverage/pom.xml
VaniHaripriya: Sure Gonzalo, will make the changes accordingly.. Thank you!

PR #2063, opened 2021-11-22
[JBPM-9952] Include sort  processes by variables and tasks queries

fjtirado:
Removing sort from named query and sorting in memory according to the
sorted list of ids

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9952)

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractAdvanceRuntimeDataServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractAdvanceRuntimeDataServiceImpl.java
@@ -16,7 +16,14 @@
 
 package org.jbpm.kie.services.impl;
 
+import static java.util.Collections.emptyList;
+import static java.util.Collections.singletonMap;
+import static java.util.stream.Collectors.toList;
+import static org.jbpm.services.api.query.model.QueryParam.all;
+
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
@@ -37,11 +44,6 @@ import org.jbpm.shared.services.impl.QueryManager;
 import org.jbpm.shared.services.impl.TransactionalCommandService;
 import org.jbpm.shared.services.impl.commands.QueryNameCommand;
 import org.kie.api.runtime.query.QueryContext;
-
-import static java.util.Collections.emptyList;
-import static java.util.Collections.singletonMap;
-import static java.util.stream.Collectors.toList;
-import static org.jbpm.services.api.query.model.QueryParam.all;
 
 public abstract class AbstractAdvanceRuntimeDataServiceImpl {
 
@@ -65,11 +67,12 @@ public abstract class AbstractAdvanceRuntimeDataServiceImpl {
                                                                                                     int processType,
                                                                                                     String varPrefix,
                                                                                                     QueryContext queryContext) {
-        BiFunction<StringBuilder, StringBuilder, String> mainSQLProducer = (derivedTables, globalWhere) -> "SELECT DISTINCT pil.processInstanceId " +
+        setDefaultSorting (queryContext, "pil.processInstanceId");
+        BiFunction<StringBuilder, StringBuilder, String> mainSQLProducer = (derivedTables, globalWhere) -> getSelectFields("pil.processInstanceId", queryContext) +
                                                                                                            " FROM ProcessInstanceLog pil \n " +
                                                                                                            derivedTables +
                                                                                                            " WHERE  pil.processType = :processType " + globalWhere;
-        setDefaultSorting (queryContext, "pil.processInstanceId");
+        
         return queryProcessUserTasksByVariables(attributes, processVariables, emptyList(), null, processType, varPrefix, queryContext, mainSQLProducer, this::collectProcessData);
     }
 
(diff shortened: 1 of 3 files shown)


elguardian:
Not very clear to me this is going to work. 
https://github.com/kiegroup/jbpm/blob/8e3ea62e6a98e765c7bd58b9d88b591b167d3c46/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractAdvanceRuntimeDataServiceImpl.java#L332

this cut is done keeping in mind the order if you remove it you don't know what kind of order the query is going to be returned if you are using pagination. meaning if you have a page size of 2 and you have 5 elements  you don't know if the first two elements will be the ones belonging to it or not as databases does not guaranteed the order not even among the query in the same transaction.

@afalhambra keep in mind what I just said and be wary of this being properly tested.

afalhambra:
> Not very clear to me this is going to work.
> 
> https://github.com/kiegroup/jbpm/blob/8e3ea62e6a98e765c7bd58b9d88b591b167d3c46/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/AbstractAdvanceRuntimeDataServiceImpl.java#L332
> 
> this cut is done keeping in mind the order if you remove it you don't know what kind of order the query is going to be returned if you are using pagination. meaning if you have a page size of 2 and you have 5 elements you don't know if the first two elements will be the ones belonging to it or not as databases does not guaranteed the order not even among the query in the same transaction.
> 
> @afalhambra keep in mind what I just said and be wary of this being properly tested.

@elguardian @fjtirado - Some results of the tests I ran:

**1. queryProcessByVariables**

Tested and works fine when using `processInstanceId` as the column to sort by, even when we have pagination.

Now, the problem I see is that it only works with that particular column, due to the fact that the first SQL fetches only the processInstanceIds in resultset, meaning that if we try to order by some other column, that query is going to fail. 

For instance, below is thrown when sorting by `processId` (which is a column of the table `ProcessInstanceLog`)

```
Caused by: org.h2.jdbc.JdbcSQLException: Order by expression "PROCESSID" must be in the result list in this case; SQL statement:
SELECT DISTINCT pil.processInstanceId  FROM ProcessInstanceLog pil 
 INNER JOIN (SELECT A1.processInstanceId 
FROM VariableInstanceLog A1 
LEFT JOIN VariableInstanceLog A2 ON A1.processId = A2.processId AND A1.processInstanceId = A2.processInstanceId AND A1.variableInstanceId = A2.variableInstanceId AND A2.id > A1.id  
WHERE A2.id IS NULL AND ((A1.variableId = ? AND A1.value = ? ) OR (A1.variableId = ? AND A1.value = ? )) GROUP BY A1.processInstanceId HAVING COUNT(*) = ? ) TABLE_PROC_VAR ON TABLE_PROC_VAR.processInstanceId = pil.processInstanceId 
 WHERE  pil.processType = ?  AND pil.processId = ?  ORDER BY processId ASC [90068-197]
 ```

**2. queryUserTasksByVariables** 

Same as queryProcessByVariables but in this case it only works with `task.id` (pagination also works fine) since it's the only resultset available in the SQL statement. 
Below result when trying to order by `processId` column which is part of the `Task` table:

```
Caused by: org.h2.jdbc.JdbcSQLException: Order by expression "PROCESSID" must be in the result list in this case; SQL statement:
SELECT DISTINCT task.id  FROM Task task  INNER JOIN ProcessInstanceLog pil ON pil.processInstanceId = task.processInstanceId 
 INNER JOIN (
    SELECT taskId 
    FROM TaskVariableImpl 
    WHERE (name = ? AND value = ? ) 
    GROUP BY taskId 
    HAVING COUNT(*) >= ? 
) TABLE_TASK_VAR ON TABLE_TASK_VAR.taskId = task.id  
 WHERE  pil.processType = ?  AND pil.externalId = ?  ORDER BY processId ASC [90068-197]
 ```

fjtirado:
@afalhambra It is cleary NOT working
@elguardian Maybe I should remove the code you indicate, because the order is now taken care at the end (the point is the the list of ids passsed to that method is now already sorted)

elguardian:
> @afalhambra It is cleary NOT working @elguardian I think the issue is a different one, working with Antonio to find out. Take into accound that the idea is that the list of ids is already in order and the comparator is sorting according to it. Since the query retrieves all the records in the list of ids, that should work (I just remove the order by because it should not be needed any longer)

The problem you need to take into account is this.
If you have in the database 1,5,2,4,5,1,2

and you need to do pagination size 2 even if you sort things out you can get wrong outcome.

SELECT * FROM DATABASE OFFSET 0 LIMIT 2
you might get 1,5  or in another run 5,1 or any other combination like 2,4 or 5,5

even if you sort this out for instance 1,5 you might leave left behind expected data. mysql for instance usually gives the same outcome for the SQL but not every database is like that.

fjtirado:
> 
> The problem you need to take into account is this. If you have in the database 1,5,2,4,5,1,2
> 
> and you need to do pagination size 2 even if you sort things out you can get wrong outcome.
> 
> SELECT * FROM DATABASE OFFSET 0 LIMIT 2 you might get 1,5 or in another run 5,1 or any other combination like 2,4 or 5,5
> 
> even if you sort this out for instance 1,5 you might leave left behind expected data. mysql for instance usually gives the same outcome for the SQL but not every database is like that.

The idea is that the first query will be SELECT * FROM DATABASE OFFSET 0 LIMIT 2 ORDER BY x so it will return 1,1. 
The the second query will retrieve 1,1 and sort them in memory (which will be quick enough) before returning them

elguardian (approved):
The idea looks fine I leave the verification to QE.

fjtirado:
jenkins retest this

afalhambra (approved):
Tested and works like a charm.
Awesome work @fjtirado!
Thanks!

fjtirado:
unrelated, jenkins retest this

fjtirado:
@elguardian ready to merge

elguardian:
good work @fjtirado

PR #2069, opened 2021-12-03
(GSS) [7.59.x] [JBPM-9915] EJBTimerScheduler.localCache contains entries from already fired timers

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9915

cherry-pick: https://github.com/kiegroup/jbpm/commit/0f229c0097ba4c06cfc7b0667fefa48593b011c9

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -62,11 +62,6 @@ public class EJBTimerScheduler {
 
 	private static final Logger logger = LoggerFactory.getLogger(EJBTimerScheduler.class);
 
-    private enum TimerExceptionPolicy {
-        RETRY,
-        PLATFORM
-    };
-
     private static final Long TIMER_RETRY_INTERVAL = Long.parseLong(System.getProperty("org.kie.jbpm.timer.retry.interval", "5000"));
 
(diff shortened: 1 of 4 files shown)


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -340,4 +338,9 @@ public TimerJobInstance getTimerByName(String jobName) {
 		return found;
 	}
 
+    public void evictCache(JobHandle jobHandle) {
+        String jobName =  ((EjbGlobalJobHandle) jobHandle).getUuid();
+        logger.debug("Invalidate job {} with job name {} in cache", jobName, localCache.remove(jobName));
```
gmunozfe, line 343: Shouldn't remove call be out of the logger invocation? If not in debug level, it won't be executed

gmunozfe (changes requested):

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
elguardian: that will be always executed (it is not a function) as you need to pass it as parameter to the debug message. A whole different thing if it is logged or not.

gmunozfe (approved):
Looks good to me, awesome work @elguardian !

afalhambra (approved):
LGTM - Same as cherry-pick

PR #2077, opened 2021-12-16
[JBPM-9961] Process with multiple timer startNodes only triggers one

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-9961


gmunozfe (changes requested):
As the name is not mandatory for nodes in the bpmn (also can be repeated), it's better to use id.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/event/StartEventTest.java
```diff
@@ -192,4 +195,31 @@ public void testDelayingTimerStartEvent() throws Exception {
         process.clear();
     }
 
+    @Test(timeout=100000000)
```
gmunozfe, line 198: Why so long? probably 10-30 seconds is enough, wdyt?

gmunozfe (approved):
Looks good to me, just a minor comment in the test about timeout, perfectly done @elguardian !

PR #2084, opened 2022-01-13
[JBPM-9964] Avoid cast exception when priority is integer

fjtirado:
**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-9964)

### DIFF
--- a/jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/AbstractHTWorkItemHandler.java
+++ b/jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/AbstractHTWorkItemHandler.java
@@ -128,11 +128,14 @@ public abstract class AbstractHTWorkItemHandler implements WorkItemHandler {
         
         task.setSubject(comment);
         
-        String priorityString = (String) workItem.getParameter("Priority");
+        
         int priority = 0;
-        if (priorityString != null) {
+        Object priorityObject = workItem.getParameter("Priority");
+        if (priorityObject instanceof Number) {
+            priority = ((Number) priorityObject).intValue();
+        } else if (priorityObject != null) {
             try {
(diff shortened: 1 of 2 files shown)


fjtirado:
jenkins retest this

jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/AbstractHTWorkItemHandler.java
```diff
@@ -128,16 +128,17 @@ protected Task createTaskBasedOnWorkItemParams(KieSession session, WorkItem work
...
+            task.setPriority(((Number) priorityObject).intValue());
+        } else if (priorityObject != null) {
             try {
-                priority = new Integer(priorityString);
+                task.setPriority(Integer.parseInt(priorityObject.toString()));
             } catch (NumberFormatException e) {
                 // do nothing
             }
         }
-        task.setPriority(priority);
```
gmunozfe, line 140: Previously, when Priority was not present (null), it was set up as 0, but here, it wouldn't be and it is not initialized with zero value in the implementation classes, so potentially it won't have priority, breaking backward compatibility, wdyt?

gmunozfe (commented):
One comment related to null priority

jbpm-human-task/jbpm-human-task-workitems/src/main/java/org/jbpm/services/task/wih/AbstractHTWorkItemHandler.java
fjtirado: You are right. 
For some reason, when I did the change yesterday, I was assuming that priority was an int, not an Integer, in which case will be initialized to 0 anyway.

elguardian (approved):

gmunozfe (approved):
Looks good to me, nailed down @fjtirado !

PR #2090, opened 2022-01-26
[JBPM-10001] Variable name might be null in IntermediateThrowHandler

fjtirado:
**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-10001)

### DIFF
--- a/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
+++ b/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
@@ -27,7 +27,6 @@ import org.jbpm.bpmn2.core.Message;
 import org.jbpm.bpmn2.handler.SendMessageAction;
 import org.jbpm.bpmn2.handler.SendSignalAction;
 import org.jbpm.compiler.xml.ProcessBuildData;
-import org.jbpm.process.core.context.variable.Variable;
 import org.jbpm.process.core.impl.DataTransformerRegistry;
 import org.jbpm.ruleflow.core.RuleFlowProcess;
 import org.jbpm.workflow.core.Node;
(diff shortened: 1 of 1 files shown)


jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
```diff
@@ -241,9 +241,15 @@ public void handleMessageNode(final Node node, final Element element,
 							"Could not find message " + messageRef);
 				}
                 message.addOutgoingNode(node);
+                String varName = null;
                 String mappingVariable = (String) actionNode.getMetaData("MappingVariable");
                 Variable v = (Variable) ((ProcessBuildData) parser.getData()).getMetaData("Variable");
-                String varName = v != null ? (String) v.getMetaData(mappingVariable) : mappingVariable;
+                if (v != null) {
```
fjtirado, line 247: In previous implementation if Variable was not null but Variable.getMedata(mappingVariable) was, then rather that use mappingVariable (as I think should be the case) we will be setting varName to null. 
This makes the message object used by the action (which is also used for Kafka  extension) null.

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
```diff
@@ -325,6 +331,7 @@ protected void readDataInputAssociation(org.w3c.dom.Node xmlNode,
             subNode = subNode.getNextSibling();
             if (subNode != null && "transformation".equals(subNode.getNodeName())) {
                 String lang = subNode.getAttributes().getNamedItem("language").getNodeValue();
+                if (text.startsWith("\"") && text.endsWith("\"")) {
```
gmunozfe, line 334: Compilation error, missing closing bracket:
[ERROR] /home/jenkins/workspace/KIE/main/pullrequest/jbpm-main.pr/bc/kiegroup_jbpm/jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java:[382,9] illegal start of expression

gmunozfe (changes requested):

jbpm-bpmn2/src/main/java/org/jbpm/bpmn2/xml/IntermediateThrowEventHandler.java
```diff
@@ -241,9 +240,7 @@ public void handleMessageNode(final Node node, final Element element,
 							"Could not find message " + messageRef);
 				}
                 message.addOutgoingNode(node);
-                String mappingVariable = (String) actionNode.getMetaData("MappingVariable");
-                Variable v = (Variable) ((ProcessBuildData) parser.getData()).getMetaData("Variable");
-                String varName = v != null ? (String) v.getMetaData(mappingVariable) : mappingVariable;
+                String varName = (String) actionNode.getMetaData("MappingVariable");
```
fjtirado, line 243: The var name should be retireved in the same way than in EndEventHandler

gmunozfe (approved):
Looks good to me, well done @fjtirado !

PR #2092, opened 2022-01-31
[JBPM-10004] GET "instances/{processInstanceId}/timers" endpoint is not returning process SLA timers

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10004

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/ListTimersCommand.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/ListTimersCommand.java
@@ -67,6 +67,14 @@ public class ListTimersCommand implements ExecutableCommand<List<TimerInstance>>
         	throw new ProcessInstanceNotFoundException("No process instance can be found for id " + processInstanceId);
         }
 
+        Long processSlaTimer = wfp.getSlaTimerId();
+        if (processSlaTimer != null && processSlaTimer != -1L) {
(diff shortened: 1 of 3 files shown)


gmunozfe (approved):
Looks good to me, works also with timer expressions.
Nailed down @elguardian !

gmunozfe:
Tested with integration worked fine as well. The only point (enhancement) here is to add a name, same to SLA for tasks, with [SLA]process.id, wdyt @elguardian ?

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/ListTimersCommand.java
```diff
@@ -66,6 +66,12 @@ public ListTimersCommand(long processInstanceId) {
         	throw new ProcessInstanceNotFoundException("No process instance can be found for id " + processInstanceId);
         }
 
+        Long processSlaTimer = wfp.getSlaTimerId();
+        if (processSlaTimer != null && processSlaTimer != -1L) {
+            timers.add(buildTimer(tm.getTimerMap().get(processSlaTimer)));
```
gmunozfe, line 71:
```suggestion
           TimerInstanceImpl details = buildTimer(tm.getTimerMap().get(processSlaTimer));
            details.setTimerName("[SLA-Process]" + wfp.getProcessName());
            timers.add(details);
```

elguardian: done

PR #2093, opened 2022-01-31
[JBPM-10005] Add suspendUntil timer to getTimers call

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10005


elguardian:
test will be included afterwards.... as it will collide with 
https://github.com/kiegroup/jbpm/pull/2092/files

gmunozfe (approved):
Looks good to me, great approach @elguardian !
Just pending to add a test when possible

PR #2105, opened 2022-02-11
[JBPM-10016] Drools/jBPM integration: high number of instances waitin…

mariofusco:
…g for signal adversely impacts execution time

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPASignalManager.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPASignalManager.java
@@ -16,12 +16,14 @@
 
 package org.jbpm.persistence.processinstance;
 
+import java.util.ArrayList;
 import java.util.List;
 
 import org.drools.core.common.InternalKnowledgeRuntime;
 import org.jbpm.persistence.api.ProcessPersistenceContext;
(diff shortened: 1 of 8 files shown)


elguardian (approved):

gmunozfe:
@mariofusco, add the reference to the drools PR in the description: https://github.com/kiegroup/drools/pull/4196

gmunozfe (changes requested):
It's not working as expected: the 3 threads are taking the same batch, instead of distributing the load among them, @elguardian could you take a look?

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/processinstance/JPASignalManager.java
```diff
@@ -16,12 +16,14 @@
 
 package org.jbpm.persistence.processinstance;
```
elguardian, line 17: @mariofusco  do you mind to remove this file change. I will treat the batch system in a different PR.

elguardian: yes. it should

elguardian: @mariofusco

PR #2107, opened 2022-02-16
[JBPM-10020]-Add incrementing and decrementing prometheus metric for …

VaniHaripriya:
JIRA- https://issues.redhat.com/browse/JBPM-10020

Created a new command to throw execution error to use it for testing a new prometheus metric created for kie server execution errors.

Related Pull Request: https://github.com/kiegroup/droolsjbpm-integration/pull/2716

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2017 Red Hat, Inc. and/or its affiliates.
+ * Copyright 2022 Red Hat, Inc. and/or its affiliates.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
@@ -31,10 +31,15 @@ public class JobExecutionErrorCommand implements Command{
     
     private static final Logger logger = LoggerFactory.getLogger(JobExecutionErrorCommand.class);
 
-    public ExecutionResults execute(CommandContext ctx) {    	
-    	ctx = null; //Assigning null value to throw null pointer exception   	 		
-    	logger.info("Command executed on executor with data {}", ctx.getData());    	
-    	        
+    @SuppressWarnings("null")
+	public ExecutionResults execute(CommandContext ctx) {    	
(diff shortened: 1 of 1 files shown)


jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
```diff
@@ -0,0 +1,47 @@
...
+    
+    private static final Logger logger = LoggerFactory.getLogger(JobExecutionErrorCommand.class);
+
+    @SuppressWarnings("null")
+	public ExecutionResults execute(CommandContext ctx) {    	
+    	ctx = null; //Assigning null value to throw null pointer exception
+		try {
+			logger.info("Command executed on executor with data {}", ctx.getData());
+		} catch (NullPointerException e) {
+
```
elguardian, line 40: this won't add up as an error as you are catching the null pointer. to simulate an error you just can
throw new RuntimeException()

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
```diff
@@ -0,0 +1,47 @@
...
+ * Just for testing purpose.
+ * 
+ */
+public class JobExecutionErrorCommand implements Command{
+    
+    private static final Logger logger = LoggerFactory.getLogger(JobExecutionErrorCommand.class);
+
+    @SuppressWarnings("null")
+	public ExecutionResults execute(CommandContext ctx) {    	
+    	ctx = null; //Assigning null value to throw null pointer exception
```
gmunozfe, line 36: this assignment is not correct, better follow Enrique's approach.

gmunozfe (changes requested):

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/JobExecutionErrorCommand.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2017 Red Hat, Inc. and/or its affiliates.
```
gmunozfe, line 2:
```suggestion
 * Copyright 2022 Red Hat, Inc. and/or its affiliates.
```

VaniHaripriya:
> If this is for testing, shouldn't it be at "test" folder instead of "main"?

Agree with you @gmunozfe , will move it to test folder.

elguardian (approved):

gmunozfe:
jenkins retest this

VaniHaripriya:
> > If this is for testing, shouldn't it be at "test" folder instead of "main"?
> 
> Agree with you @gmunozfe , will move it to test folder.

@gmunozfe ,Prometheus Integration test can not access the class in test folder, as we are accessing jbpm-executor from drools-jbpm-integration through the dependency jar. So, leaving JobExecutionErrorCommand.java in main folder.

PR #2115, opened 2022-02-28
[RHPAM-4192] NullPointer related to timers

elguardian:
Jira: https://issues.redhat.com/browse/RHPAM-4192

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/ListTimersCommand.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/ListTimersCommand.java
@@ -21,6 +21,8 @@ import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
+import java.util.function.Supplier;
 import java.util.regex.Matcher;
 
 import org.drools.core.command.SingleSessionCommandService;
@@ -69,8 +71,7 @@ public class ListTimersCommand implements ExecutableCommand<List<TimerInstance>>
 
         Long processSlaTimer = wfp.getSlaTimerId();
         if (processSlaTimer != null && processSlaTimer != -1L) {
-            TimerInstanceImpl details = buildTimer(tm.getTimerMap().get(processSlaTimer));
-            details.setTimerName("[SLA-Process] " + wfp.getProcessName());
+            TimerInstanceImpl details = buildTimer(tm.getTimerMap().get(processSlaTimer), () -> "[SLA-Process] " + wfp.getProcessName());
             timers.add(details);
         }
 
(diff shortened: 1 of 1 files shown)


albfan:
This still doesn't fix the NullPointer:

```
    static class Data {
        private String name;

        public String getName() {
            return name;
        }

        public void setName(String name) {
            this.name = name;
        }
    }
    public static String wrap(String name) {
        return name;
    }

    public static String hello(Supplier<String> resolveName) {
        return resolveName.get();
    }

    public static void main(String[] args) {
        System.out.println(hello(() -> {return wrap(null);}));
        final Data data = new Data();
        data.setName("john");
        System.out.println(hello(() -> {return wrap(data.getName());}));
        System.out.println(hello(() -> wrap(data.getName())));
        final Data dataNull = null;
        System.out.println(hello(() -> wrap(dataNull.getName())));
    }
```

which outputs:

```
null
john
john
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "org.jbpm.kie.services.impl.admin.commands.ListTimersCommand$Data.getName()" because "dataNull" is null
	at org.jbpm.kie.services.impl.admin.commands.ListTimersCommand.lambda$main$3(ListTimersCommand.java:219)
	at org.jbpm.kie.services.impl.admin.commands.ListTimersCommand.hello(ListTimersCommand.java:209)
	at org.jbpm.kie.services.impl.admin.commands.ListTimersCommand.main(ListTimersCommand.java:219)
```

Still a check for null is needed like:

```
    static class Data {
        private String name;

        public String getName() {
            return name;
        }

        public void setName(String name) {
            this.name = name;
        }
    }
    public static String wrap(String name) {
        return name;
    }

    public static String hello(Supplier<String> resolveName) {
        return resolveName.get();
    }

    public static void main(String[] args) {
        System.out.println(hello(() -> {return wrap(null);}));
        final Data data = new Data();
        data.setName("john");
        System.out.println(hello(() -> {return wrap(data.getName());}));
        System.out.println(hello(() -> wrap(data.getName())));
        final Data dataNull = null;
        System.out.println(hello(() -> {
            return dataNull != null ? wrap(dataNull.getName()): null;
        }));
    }```

outputs:

```
null
john
john
null
```

`resolveVariable()` deals correctly with a null name;

**Comment by elguardian** [comment]
Jenkins retest this please

**Comment by elguardian** [comment]
Hi @albfan 

The line raises a null pointer here 
https://github.com/kiegroup/jbpm/blob/7.59.x/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/admin/commands/ListTimersCommand.java#L114

the only possible thing is that timer is null. If you see the PR the function won't be invoked as long as that timer is null (is within the if. in your example you need to add that check.

**Comment by albfan** [comment]
Yes, just notice that, but unprotect lambda because internally buildTimer() checks timer is not null is like a code coupled, ins't it? 

Previous code do not mention that timer cannot be null setting name in other cases like:

```
TimerInstanceImpl details = buildTimer(tm.getTimerMap().get(processSlaTimer));
details.setTimerName("[SLA-Process] " + wfp.getProcessName());
```

That's why I avoid to wrap everything on buildTimer(), but maybe it can work, I'm unsure about those cases.

PR #2121, opened 2022-03-11
[JBPM-10037] Forcing creation of new transaction for partial delete

fjtirado:
Depends on https://github.com/kiegroup/drools/pull/4258


elguardian (changes requested):
same comment as in drools PR

PR #2136, opened 2022-04-04
[JBPM-10037] LogCleanupCommnand and records per transation

fjtirado:
Rescheduling command automatically if suspecting there might be more
records to delete when records per transaction is set
ProcessInstanceLog deletion are saved for the latests transaction, since the subqueries depends on them

**JIRA**:

[link](https://issues.redhat.com/browse/JBPM-10037)

**referenced Pull Requests**:

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
@@ -17,9 +17,13 @@
 package org.jbpm.executor.commands;
 
 import java.text.SimpleDateFormat;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.Map;
+import java.util.concurrent.TimeUnit;
 
 import javax.persistence.EntityManagerFactory;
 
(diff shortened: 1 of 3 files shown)


jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
```diff
@@ -66,19 +70,23 @@
...
+    private boolean mightBeMore;
 
     @Override
-    public Date getScheduleTime() {
-        if (nextScheduleTimeAdd < 0) {
-            return null;
+    public Date getScheduleTime() {        
+        Date nextSchedule;
+        if (mightBeMore) {
+            // if there are pendings records, reexecute inmediately
```
afalhambra, line 80: typo
```suggestion
            // if there are pending records, reexecute immediately
```

afalhambra (commented):
Great work @fjtirado - I don't know how feasible you see to add a test case to cover this, but it would be nice

elguardian:
failure seems related @fjtirado

elguardian:
failure seems related @fjtirado

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/jobexec/LogCleanupCommandTest.java
```diff
@@ -280,26 +280,27 @@ public void deleteAllLogsOlderThanTomorrowDifferentProcess() throws Exception {
         Assertions.assertThat(getNodeInstanceLogSize(HELLO_WORLD_ID)).isPositive();
     }
 
-    @Test(timeout = 10000)
```
afalhambra, line 283: I wouldn't remove the timeout. Since it's using a countDownLatch, this test in case of error might be hanging forever waiting for the latch to complete. I would keep this to interrupt the test in that scenario

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/jobexec/LogCleanupCommandTest.java
```diff
@@ -280,26 +280,27 @@ public void deleteAllLogsOlderThanTomorrowDifferentProcess() throws Exception {
...
+        int nodeSize = getNodeInstanceLogSize(HELLO_WORLD_ID);
+        Assertions.assertThat(processSize).isEqualTo(3);
+        Assertions.assertThat(nodeSize).isPositive();
+        
+        CountDownAsyncJobListener countDownListener = new CountDownAsyncJobListener(processSize + nodeSize);
+        ((ExecutorServiceImpl) getExecutorService()).addAsyncJobListener(countDownListener);
+        
         // Schedule cleanup job
-        scheduleLogCleanup(false, false, false, null, null, HELLO_WORLD_ID, 1);
+        scheduleLogCleanup(false, false, true, null, null, HELLO_WORLD_ID, 1);
```
afalhambra, line 298: not sure about this change? why are we skipping the ExecutorLogs?

afalhambra (approved):
Well done @fjtirado!
LGTM

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/jobexec/LogCleanupCommandTest.java
fjtirado: Thats to make the test predictable, we want to delete the nodes and the process logs, so the number of transactions, if recordsPerTransaction is 1,  is the sum of number of nodes plus the number of process (which has to be deleted in the latest transactions because they are used to resolve the other items to be deleted) 
At the end, the test is pretty complete, we are checking the boolean, the number of transactions and the items are really deleted.

jbpm-test-coverage/src/test/java/org/jbpm/test/functional/jobexec/LogCleanupCommandTest.java
fjtirado: yes, you are right, restored

PR #2142, opened 2022-04-25
[JBPM-10065] Triggering task activated event in a different way

fjtirado:
Moving the activate event triggering code inside addTask. This should generate a valid sequence. 

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-10065)

### DIFF
--- a/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
+++ b/jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
@@ -20,12 +20,16 @@
  */
 package org.jbpm.services.task.impl;
 
+import java.lang.reflect.InvocationHandler;
+import java.lang.reflect.Method;
+import java.lang.reflect.Proxy;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Consumer;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -106,7 +110,7 @@ public class TaskInstanceServiceImpl implements TaskInstanceService {
 
    
     public long addTask(Task task, Map<String, Object> params) {    	
-    	taskEventSupport.fireBeforeTaskAdded(task, context);
+        fireWithStatus (task, null, this::fireBeforeTaskAdded);
     	
     	persistenceContext.persistTask(task);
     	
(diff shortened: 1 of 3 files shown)


jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
```diff
@@ -143,12 +145,66 @@ public long addTask(Task task, ContentData contentData) {
...
 
+    private void fireEvents(Task task) {
+        fireWithStatus(task, Status.Created, this::fireAfterTaskAdded);
+        if (task.getTaskData().getStatus() != Status.Created) {
+            fireWithStatus(task, Status.Created, this::fireBeforeTaskActivated);
+            taskEventSupport.fireAfterTaskActivated(task, context);
+        }
+    }
+
+    private void fireWithStatus(Task task, Status eventStatus, Consumer<Task> consumer) {
```
gmunozfe, line 160: it's only invoked with `Status.Created ` so this param could be taken out from the input params of the method:
```suggestion
    private void fireWithStatus(Task task, Consumer<Task> consumer) {
```

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
```diff
@@ -143,12 +145,66 @@ public long addTask(Task task, ContentData contentData) {
...
+        if (task.getTaskData().getStatus() == eventStatus) {
```
gmunozfe, line 161: eventStatus is always Created?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
```diff
@@ -143,12 +145,66 @@ public long addTask(Task task, ContentData contentData) {
...
+            consumer.accept(task);
+        } else {
+            consumer.accept(createProxy(new InterceptorInvocationHandler<>(task, createProxy(new InterceptorInvocationHandler<>(task.getTaskData(), eventStatus, "getStatus")),
```
gmunozfe, line 164: Same here

gmunozfe (commented):
Some questions related to `fireWithStatus` eventStatus, as it's always Created.

Could you add any tests for completeness?

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
fjtirado: This was on purpose in case in fuure we need to publish an event with different status in this class

jbpm-human-task/jbpm-human-task-core/src/main/java/org/jbpm/services/task/impl/TaskInstanceServiceImpl.java
fjtirado: You see, after last change we are using fireWithStatus with a different value ;)

gmunozfe: I see, good point

gmunozfe (approved):
Looks good to me, thanks for adding the test and your clarification,
nailed down @fjtirado !

PR #2151, opened 2022-05-24
[JBPM-10080] DDLScripts errors after JBPM-10072

gmunozfe:
**JIRA**: [JBPM-10080](https://issues.redhat.com/browse/JBPM-10080)


jbpm-db-scripts/src/main/resources/db/upgrade-scripts/sqlserver/jbpm-7.69-to-7.70-TimerMappingInfo.sql
```diff
@@ -1 +1 @@
-ALTER TABLE TimerMappingInfo ADD COLUMN info varbinary(MAX);
```
afalhambra, line 1: we should also do the same change for the following script:

- https://github.com/kiegroup/jbpm/blob/main/jbpm-db-scripts/src/main/resources/db/upgrade-scripts/sqlserver2008/jbpm-7.69-to-7.70-TimerMappingInfo.sql

Also, should we keep this file named as is, or should we create a new one named jbpm-7.70-to-7.71-TimerMappingInfo.sql? wdyt?

afalhambra (changes requested):
Looks good! only a missing sql script we should modify as well.

jbpm-db-scripts/src/main/resources/db/upgrade-scripts/sqlserver/jbpm-7.69-to-7.70-TimerMappingInfo.sql
gmunozfe: thanks for pointing out @afalhambra, good catch!
I think we can keep it as it is (even if 7.70 is released) because the feature belongs to that version, and these are static scripts. @elguardian wdyt?

afalhambra (approved):
Great work @gmunozfe!

gmunozfe:
@mareknovotny ready to merge

PR #2153, opened 2022-05-31
[JBPM-10081] Task Description field is set with the same value of Tas…

abhijithumbe:
…k Subject

 

**JIRA**: _(https://issues.redhat.com/browse/JBPM-10081)_


elguardian (changes requested):
This change is not backward compatible.

PR #2201, opened 2022-08-10
[RHPAM-4445] Unable to update the task description with long string(m…

abhijithumbe:
…ore than 255 chars)

**JIRA**: _(https://issues.redhat.com/browse/JBPM-10113)_

### DIFF
--- a/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
+++ b/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
@@ -63,7 +63,7 @@ public class JPATaskLifeCycleEventListener extends PersistableEventListener impl
     private static final Logger logger = LoggerFactory.getLogger(JPATaskLifeCycleEventListener.class);
     private List<ArchiveLoggerProvider> archiveLoggerProviders;
 
-    private  final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
+    private static final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
 
     
     public JPATaskLifeCycleEventListener(boolean flag) {
--- a/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
+++ b/jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
@@ -63,6 +63,8 @@ public class JPATaskLifeCycleEventListener extends PersistableEventListener impl
     private static final Logger logger = LoggerFactory.getLogger(JPATaskLifeCycleEventListener.class);
     private List<ArchiveLoggerProvider> archiveLoggerProviders;
 
+    private  final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
+
     
     public JPATaskLifeCycleEventListener(boolean flag) {
         super(null);
@@ -641,10 +643,10 @@ public class JPATaskLifeCycleEventListener extends PersistableEventListener impl
         
     }
 
-    public String getUpdateFieldLog(String fieldName, String previousValue, String value){
-        return "Updated "+ fieldName
-                + " {From: '"+ (previousValue!=null ? previousValue :  "" )
-                + "' to: '"+ (value!=null ? value :  "" ) + "'}" ;
+    public String getUpdateFieldLog(String fieldName, String previousValue, String value) {
+        return "Updated " + fieldName 
+                + " {From: '"+ (previousValue != null ? previousValue : "") 
+                + "' to: '"+ (value != null ? value : "") + "'}";
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
```diff
@@ -659,10 +661,15 @@ public void afterTaskUpdatedEvent(TaskEvent event) {
                 return;
             }
 
-            if((ti.getDescription() != null && !ti.getDescription().equals(auditTaskImpl.getDescription()))
-                    || (ti.getDescription() == null && auditTaskImpl.getDescription() != null)){
+            if ((ti.getDescription() != null && !ti.getDescription().equals(auditTaskImpl.getDescription()))
+                    || (ti.getDescription() == null && auditTaskImpl.getDescription() != null)) {
                 String message = getUpdateFieldLog("Description", auditTaskImpl.getDescription(), ti.getDescription());
 
+                if (message != null && message.length() > TASK_DESCRIPTION_LENGTH) {
```
elguardian, line 668: this should not be required. it should be trimmed already in the task.

abhijithumbe: Hi @elguardian,  TaskEvent message is getting persisted in the TaskEvent table and here task event message build as  "Updated Description from: 'old value' to:new value' ". If new value of task description is 255 chars, overall task event message length is more than 255 chars  so we have to trim the message as well.

jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/TaskImpl.java
```diff
@@ -65,6 +68,14 @@ public class TaskImpl implements InternalTask {
      * WSHT uses a name for the unique identifier, for now we use a generated ID which is also the key, which can be
      * mapped to the name or a unique name field added later.
      */
+	
+    private static final Logger logger = LoggerFactory.getLogger(TaskImpl.class);
+    
+    @Transient
+    private final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
```
afalhambra, line 75:
```suggestion
    private static final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
```

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
```diff
@@ -63,6 +63,8 @@ public class JPATaskLifeCycleEventListener extends PersistableEventListener impl
     private static final Logger logger = LoggerFactory.getLogger(JPATaskLifeCycleEventListener.class);
     private List<ArchiveLoggerProvider> archiveLoggerProviders;
 
+    private  final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
```
afalhambra, line 66:
```suggestion
    private static final int TASK_DESCRIPTION_LENGTH = Integer.parseInt(System.getProperty("org.jbpm.ht.task.description.length", "255"));
```

afalhambra (commented):
Hi @abhijithumbe, can you please elaborate a little more in the jira ticket. Seeing the PR is not clear to me about the scope, since the DDL scripts are fixed and limiting the length of this field to 255 chars. So even if customer set this new property to a value higher than 255, it will keep on failing if the description is longer than 255 characters.

jbpm-human-task/jbpm-human-task-jpa/src/main/java/org/jbpm/services/task/impl/model/TaskImpl.java
abhijithumbe: Hi @afalhambra Values defined for system-property and DB column size should be in sync. We need to use Alter table query to modify DB column size based on value defined for system-property

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
elguardian: that is the reason it should be done in some other place. you will have a description that won't match database.

abhijithumbe: I agree task description in TaskEvent wont match with task description in other tables, but message field from TaskEvent is not used anywhere else and actual value of TaskDescription is captured and in sync in Task and AuditTaskImpl table. If we dont update the TaskEvent.message then we have to trim the taskDescription more than defined length.  I can modify warn message and capture the task event message before truncating.

afalhambra (approved):

elguardian:
Jenkins retest this please

jbpm-human-task/jbpm-human-task-audit/src/main/java/org/jbpm/services/task/audit/JPATaskLifeCycleEventListener.java
elguardian: @abhijithumbe task event is used in emitters... anycase there is much time for the cut off. I will let this slip

elguardian:
@abhijithumbe plz backport this to7.67 and 7.67-blue

abhijithumbe:
Thanks @elguardian . Opened PR for https://github.com/kiegroup/jbpm/pull/2207 and https://github.com/kiegroup/jbpm/pull/2208 for 7.67 and 7.67-blue branch.

PR #2228, opened 2022-10-20
[JBPM-10129] Signals inside a multinode process are not captured correctly by same process

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10129


gmunozfe (approved):
Looks good to me, outstanding work @elguardian !

PR #2235, opened 2022-11-23
JBPM-10139: Avoid duplicate entries in PerCaseRuntimeManager.local map to avoid multiple signals being sent

martinweiler:
**JIRA**: [JBPM-10139](https://issues.redhat.com/browse/JBPM-10139)

Without the check, the local map contains two entries for the same RuntimeEngine - one with the caseId as key, the other with the processInstanceId. The entries of the local map are then iterated over in the signalEvent method, which is causing duplicate signals to be sent.

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
@@ -20,9 +20,11 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.EventListener;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 import org.drools.core.command.SingleSessionCommandService;
 import org.drools.core.command.impl.CommandBasedStatefulKnowledgeSession;
@@ -274,7 +276,7 @@ public class PerCaseRuntimeManager extends AbstractRuntimeManager {
         // process currently active runtime engines
         Map<Object, RuntimeEngine> currentlyActive = local.get();
         if (currentlyActive != null && !currentlyActive.isEmpty()) {
-            RuntimeEngine[] activeEngines = currentlyActive.values().toArray(new RuntimeEngine[currentlyActive.size()]);
+            Set<RuntimeEngine> activeEngines = new HashSet<>(currentlyActive.values());
             for (RuntimeEngine engine : activeEngines) {
                 Context<?> context = ((RuntimeEngineImpl) engine).getContext();
                 if (context != null && context instanceof ProcessInstanceIdContext && ((ProcessInstanceIdContext) context).getContextId() != null) {
--- a/jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManagerTest.java
+++ b/jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManagerTest.java
@@ -435,6 +435,48 @@ public class PerCaseRuntimeManagerTest extends AbstractBaseTest {
         assertEquals(1, ksessionUsed.size());
         assertEquals(ksession1Id, ksessionUsed.iterator().next().longValue());
     }
+
+    @Test
+    public void testSignalStageWithProcessScope() throws Exception {
+        final List<Long> taskActivations = new ArrayList<Long>();
+
+        RuntimeEnvironment environment = RuntimeEnvironmentBuilder.Factory.get()
+                .newDefaultBuilder()
+                .userGroupCallback(userGroupCallback)
+                .registerableItemsFactory(new DefaultRegisterableItemsFactory(){
+
+                    @Override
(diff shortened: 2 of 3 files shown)


fjtirado:
@martinweiler Im not really sure the two signals are an issue, but without clear requirements I cannot have a strong opinion on that. 
To avoid all these test failures I just added a comment that might or not might work.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
```diff
@@ -451,6 +451,14 @@ protected void saveLocalRuntime(Object caseId, Object processInstanceId, Runtime
...
             map = new HashMap<Object, RuntimeEngine>();
             local.set(map);
+        } else {
+            // check if entry already exists
+            for (Map.Entry<Object, RuntimeEngine> entry : map.entrySet()) {
+                if (runtime.equals(entry.getValue())) {
+                    // runtime already in map
+                    return;
+                }
+            }
```
fjtirado, line 461: Maybe we can delate that and just write 
```
protected void saveLocalRuntime(Object caseId, Object processInstanceId, RuntimeEngine runtime) {

        Map<Object, RuntimeEngine> map = local.get();
        if (map == null) {
            map = new HashMap<Object, RuntimeEngine>();
            local.set(map);
        }
        if (caseId != null) {
            map.put(caseId, runtime);
        }
        else if (processInstanceId != null) {
            map.put(processInstanceId, runtime);
        }

    }
```

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManagerTest.java
```diff
@@ -435,6 +435,48 @@ public void beforeProcessStarted(ProcessStartedEvent event) {
...
+        // since there is no process instance yet we need to get new session
+        RuntimeEngine runtime = manager.getRuntimeEngine(CaseContext.get("Case-1"));
+        KieSession ksession = runtime.getKieSession();
+
+        ProcessInstance pi1 = ksession.startProcess("signal_test.signal_test");
+        manager.disposeRuntimeEngine(runtime);
+        manager.close();
+
+        // there should be only one Task2 instance activation
+        assertEquals(taskActivations.size(), 1);
```
afalhambra, line 477: @martinweiler please check sonar cloud, the order here should be swap:

```suggestion
        assertEquals(1, taskActivations.size());
```

afalhambra (changes requested):

jbpm-runtime-manager/src/test/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManagerTest.java
martinweiler: Thanks for pointing this out, I changed the assert statement.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerCaseRuntimeManager.java
martinweiler: I agree this looks more elegant, but the problem is that this method gets invoked multiple times before the signalEvent method call. On one call, the processInstanceId is null, on the other the caseId is null, so with the else approach, this would still result in duplicate map entries for the same RuntimeEngine:

11:52:45,256 DEBUG [PerCaseRuntimeManager] (default task-8) SaveLocalRuntime called for caseId CASE-0000000036, processInstanceId null, runtimeEngine org.jbpm.runtime.manager.impl.RuntimeEngineImpl@53be1643(KieSessionId=8561)

11:52:45,256 DEBUG [PerCaseRuntimeManager] (default task-8) SaveLocalRuntime called for caseId null, processInstanceId 8339, runtimeEngine org.jbpm.runtime.manager.impl.RuntimeEngineImpl@53be1643(KieSessionId=8561)

fjtirado: @martinweiler Thanks for the explanation. Then, as you currently grasped, my proposal wont work (I was not fully convinced, to be honest ;))
Do you know why the CI is failing?  it seems somehow related

fjtirado: I suggest to try this other approach (maybe with that the CI works)
In line 277
replace ` RuntimeEngine[] activeEngines = currentlyActive.values().toArray(new RuntimeEngine[currentlyActive.size()]);`
by
` Set<RuntimeEngine> activeEngines = new HashSet<>(currentlyActive.values());`

martinweiler: @fjtirado my bad, I thought I had a green build locally, but must have mixed this up. Your approach is more targeted towards the problem at hand, and does not seem to have any side effects. Thanks for the suggestion!

fjtirado (approved):
Thanks Martin!

afalhambra (approved):
LGTM - thanks

PR #2250, opened 2022-12-08
[JBPM-10144] Flush TimerMappingInfo persist operation early to avoid stale timers

martinweiler:
**JIRA**: https://issues.redhat.com/browse/JBPM-10144


fjtirado (approved):
Thanks a lot @martinweiler

PR #2272, opened 2023-02-14
[RHPAM-4648]: Error code: 404 on History button for Process Variable of type: org.jbpm.document.DocumentCollection

danielemams:
[RHPAM-4648](https://issues.redhat.com/browse/RHPAM-4648)


fjtirado (changes requested):
We are using a different approach.
Keeping this open just in case the alternative approach is a dead end (so we roll back to this quick work around)

PR #2282, opened 2023-05-04
[kie-roadmap#52] Improve LogCleanupCommand, do not delete RequestInfo and ErrorInfo records of active process instances

martinweiler:

[kie-roadmap-52](https://github.com/IBM/kie-roadmap/issues/52)

Running LogCleanupCommand with parameter OlderThanPeriod=xd removes inactive records from RequestInfo and ErrorInfo even if the related process instance is still active. These entries should not be deleted as long as the process instance is active, as this breaks the ability to reschedule a failed job. Records for active process instances might also be needed for audit purposes.


elguardian (approved):
LGTM

PR #2285, opened 2023-05-23
[JBPM-10172] Same id used for timers when using per request

fjtirado:
This basically makes timer id to behave as if singleton strategy was used both for per request and per process instance. 
Thats clearly needed on per request. Singleton remains the same and the only concern might be that know the timer id is globally unique and before, when using per process instance, it was unique per process instance. 
But I think that a per process instance id into a globally unique one should not cause any issue (the contrary is not true)

Changed the approach to use the max process instance timer id if the runtime strategy if not per process instance

**JIRA**: 

[link](https://issues.redhat.com/browse/JBPM-10172)


gmunozfe (approved):
Some tests are failing, need to check

gmunozfe (approved):
Looks good to me, it works like a charm, great work @fjtirado !

PR #2292, opened 2023-06-05
[kie-roadmap-119] Scheduled jobs do not keep configured execution time

martinweiler:
When scheduling a LogCleanupCommand with NextRun=1d, the actual execution time for the next run is calculated as (current command execution time + nextRun value). Over time, this can lead to the scheduled execution time to slip, ie. a job scheduled initially at 00:00 gets executed at 00:10, and the execution could be outside of a designated maintenance window.

The expectation is that subsequent executions happen at the defined start time + nextRun value, ie. a job scheduled initially at 00:00 with NextRun=1d should always run at 00:00, regardless of how long it takes to execute it.

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
@@ -56,6 +56,12 @@ import org.slf4j.LoggerFactory;
  * 	<li>DateFormat - date format for further date related params - if not given yyyy-MM-dd is used (pattern of SimpleDateFormat class)</li>
  * 	<li>EmfName - name of entity manager factory to be used for queries (valid persistence unit name)</li>
  * 	<li>SingleRun - indicates if execution should be single run only (true|false)</li>
+ * 	<li>RepeatMode - defines how the next execution time is calculated
+ * 	 <ul>
+ * 	  <li>FIXED - Scheduled time + NextRun parameter value</li>
+ * 	  <li>INTERVAL - End of execution + NextRun parameter value (default)</li>
+ * 	 </ul>
+ * 	</li>
  * 	<li>NextRun - provides next execution time (valid time expression e.g. 1d, 5h, etc)</li>
  * 	<li>OlderThan - indicates what logs should be deleted - older than given date</li>
  * 	<li>OlderThanPeriod - indicated what logs should be deleted older than given time expression (valid time expression e.g. 1d, 5h, etc)</li>
@@ -125,15 +131,6 @@ public class LogCleanupCommand implements Command, Reoccurring {
             emfName = "org.jbpm.domain";
         }
         nextScheduleTimeAdd = TimeUnit.DAYS.toMillis(1);
-        String singleRun = (String) ctx.getData("SingleRun");
-        if ("true".equalsIgnoreCase(singleRun)) {
-            // disable rescheduling
-            nextScheduleTimeAdd = -1;
-        }
-        String nextRun = (String) ctx.getData("NextRun");
-        if (nextRun != null) {
-            nextScheduleTimeAdd = DateTimeUtils.parseDateAsDuration(nextRun);
-        }
 
         // get hold of persistence and create instance of audit service
         EntityManagerFactory emf = EntityManagerFactoryManager.get().getOrCreate(emfName);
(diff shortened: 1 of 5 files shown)


jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/BasicExecutorBaseTest.java
```diff
@@ -308,7 +311,11 @@ public void reoccurringExecutionTest() throws InterruptedException {
         List<RequestInfo> executedRequests = executorService.getCompletedRequests(new QueryContext());
         assertEquals(3, executedRequests.size());
 
-
+        RequestInfo firstRequest = executedRequests.get(0);
+        RequestInfo lastRequest = executedRequests.get(executedRequests.size()-1);
+        long diffInMillies = Math.abs(lastRequest.getTime().getTime() - firstRequest.getTime().getTime());
+        // last request should be 2x NextRun = 4s later than first request, regardless of the sleepTime
+        assertEquals(4, TimeUnit.SECONDS.convert(diffInMillies, TimeUnit.MILLISECONDS));
```
elguardian, line 318: I would not use equals here... as it would lead to probably to flacky  tests

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
```diff
@@ -106,6 +106,7 @@ protected int readInt(Map<String, Object> params, String propName, int defaultVa
 
     @Override
     public ExecutionResults execute(CommandContext ctx) throws Exception {
+        long startTime = Instant.now().toEpochMilli();
```
elguardian, line 109: instead of this I would use the ctx to get the running time. AbstractAvailableJobsExecutor::executeGivenJob during the ctx init.
https://github.com/kiegroup/jbpm/blob/main/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java#L129
There is the proper time in there.

RequestInfo.getTime you have there the time scheduled not something computed, so it would be more accurate strictly speaking.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
```diff
@@ -268,6 +265,12 @@ public ExecutionResults execute(CommandContext ctx) throws Exception {
                 executionResults.setData("ProcessInstanceLogRemoved", piLogsRemoved);
             }
         }
+        // calculate nextRun at the end of the execution
+        String nextRun = (String) ctx.getData("NextRun");
+        if (nextRun != null) {
+            long executionTime = Instant.now().minus(startTime, ChronoUnit.MILLIS).toEpochMilli();
+            nextScheduleTimeAdd = DateTimeUtils.parseDateAsDuration(nextRun) - executionTime;
```
elguardian, line 272: You are modifying the behaviour for every command in here. 
I would add a new flag in here saying Repeated FIXED_RATE, OR FIXED_INTERVAL

FIXED_RATE = scheduled + interval
FIXED_INTERVAL = end of execution + interval (default behaviour)

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/BasicExecutorBaseTest.java
martinweiler: I have changed the test assertion to be in line with others in the same class.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
martinweiler: great idea, thanks

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
martinweiler: Added a parameter to the command, with the default behavior being inline with how it has been calculated so far. Note that I also added a base class for a central place for the calculation.

gmunozfe (commented):
@martinweiler Good work! I think there's an edge scenario that can be handled (nextScheduleTimeAdd is negative after subtraction). Other two minor comments and also you could rebase it as we are in the next snapshot now.

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
```diff
@@ -268,11 +265,40 @@ public ExecutionResults execute(CommandContext ctx) throws Exception {
...
+        long executionTimeInMillis = 0;
+        String repeatMode = (String) ctx.getData("RepeatMode");
+        if( "fixed".equalsIgnoreCase(repeatMode) ) {
+            // return diff between scheduled time and actual time
+            Date scheduledExecutionTime = (Date) ctx.getData("scheduledExecutionTime");
+            executionTimeInMillis = Instant.now().minus(scheduledExecutionTime.toInstant().toEpochMilli(), ChronoUnit.MILLIS).toEpochMilli();
+            logger.debug("Calculated execution time {}ms, based on scheduled execution time {}", executionTimeInMillis, scheduledExecutionTime);
+        } else if( "interval".equalsIgnoreCase(repeatMode) ) {
+            // no calculation required
+        }
```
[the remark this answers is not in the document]
martinweiler: you are right, it is useless. removed

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
```diff
@@ -56,6 +56,12 @@
  * 	<li>DateFormat - date format for further date related params - if not given yyyy-MM-dd is used (pattern of SimpleDateFormat class)</li>
  * 	<li>EmfName - name of entity manager factory to be used for queries (valid persistence unit name)</li>
  * 	<li>SingleRun - indicates if execution should be single run only (true|false)</li>
+ * 	<li>RepeatMode - defines how the next execution time is calculated
+ * 	 <ul>
+ * 	  <li>FIXED - Scheduled time + NextRun parameter value</li>
+ * 	  <li>INTERVAL - End of execution + NextRun parameter value (default)</li>
```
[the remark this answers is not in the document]
martinweiler: done

jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/commands/LogCleanupCommand.java
```diff
@@ -268,11 +265,40 @@ public ExecutionResults execute(CommandContext ctx) throws Exception {
...
+        String singleRun = (String) ctx.getData("SingleRun");
+        if ("true".equalsIgnoreCase(singleRun)) {
+            // disable rescheduling
+            nextScheduleTimeAdd = -1;
+            return;
+        }
+
+        String nextRun = (String) ctx.getData("NextRun");
+        if (nextRun != null) {
+            nextScheduleTimeAdd = DateTimeUtils.parseDateAsDuration(nextRun) - calculateExecutionTimeInMillis(ctx);
```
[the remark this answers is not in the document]
martinweiler: @gmunozfe thanks for the input, that makes sense. I added the check that in case there is an overlap (previous execution took longer than nextRun value), the next execution is scheduled immediately. Also added a new test case for this scenario.

gmunozfe (approved):
Looks good to me, astounding work @martinweiler !
Could you add the labels for automatic backporting to 7.67.x and blue? I think with that it's ready to merge

martinweiler:
Thank you @gmunozfe  ! I have added the labels now.

PR #2318, opened 2023-08-17
[JBPM-10187] Sorting resources to avoid deadlock

fjtirado:
**JIRA**:  

[link](https://issues.redhat.com/browse/JBPM-10187)

**referenced Pull Requests**:
https://github.com/kiegroup/drools/pull/11

### DIFF
--- a/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java
+++ b/jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java
@@ -41,6 +41,7 @@ import org.kie.internal.process.CorrelationKey;
 public class JpaProcessPersistenceContext extends JpaPersistenceContext
     implements
     ProcessPersistenceContext {
+ 
     
     public JpaProcessPersistenceContext(EntityManager em, TransactionManager txm) {
         super( em, txm );
@@ -66,7 +67,10 @@ public class JpaProcessPersistenceContext extends JpaPersistenceContext
     	if( this.pessimisticLocking ) { 
             return em.find( ProcessInstanceInfo.class, processId, lockMode );
         }
-        return em.find( ProcessInstanceInfo.class, processId );
+        logger.trace("Reading process instance info {} with em {}", processId, em);
+        ProcessInstanceInfo pi = em.find(ProcessInstanceInfo.class, processId);
+        logger.trace("Process instance info read {}", pi);
+        return pi;
     }
 
     public void remove(PersistentProcessInstance processInstanceInfo) {
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
@@ -83,6 +83,7 @@ import org.kie.internal.runtime.manager.InternalRuntimeManager;
 import org.kie.internal.runtime.manager.RuntimeManagerRegistry;
 import org.kie.internal.runtime.manager.SecurityManager;
 import org.kie.internal.runtime.manager.SessionFactory;
+import org.kie.internal.runtime.manager.SessionNotFoundException;
 import org.kie.internal.runtime.manager.TaskServiceFactory;
 import org.kie.internal.runtime.manager.context.ProcessInstanceIdContext;
 import org.kie.internal.runtime.manager.deploy.DeploymentDescriptorManager;
@@ -389,8 +390,8 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
                     && tm.getStatus() != TransactionManager.STATUS_COMMITTED) {
                 return false;
             }
-        } catch (Exception e) {
-            logger.warn("Exception dealing with transaction", e);
+        } catch (SessionNotFoundException e) {
+            logger.warn("Session not found exception", e);
         }        
         return true;
     }
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
@@ -212,7 +212,7 @@ public class PerProcessInstanceRuntimeManager extends AbstractRuntimeManager {
                     try {
                         engineImpl.getKieSession().signalEvent(type, event,
                                 ((ProcessInstanceIdContext) context).getContextId());
-                    } catch (SessionNotFoundException ex) {
+                    } catch (org.drools.persistence.api.SessionNotFoundException ex) {
                         logger.warn(
                                 "Signal event cannot proceed because of session not found exception {} for engine {}",
                                 ex.getMessage(), engineImpl.getKieSessionId());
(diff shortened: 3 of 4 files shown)


mareknovotny:
@krisv @albfan is this ok to review as this looks like one fix we would need to review and merge or explain why it is not good change ;)

mareknovotny:
of course we need to remove log trace msgs from code

fjtirado:
> of course we need to remove log trace msgs from code

why? These traces are legit anyway

mareknovotny:
> > of course we need to remove log trace msgs from code
> 
> why? These traces are legit anyway

ok, i thought it is just for debugging purposes. If they are valid even for production, then ok.

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -389,8 +390,8 @@ protected boolean canDispose(RuntimeEngine runtime) {
                     && tm.getStatus() != TransactionManager.STATUS_COMMITTED) {
                 return false;
             }
-        } catch (Exception e) {
-            logger.warn("Exception dealing with transaction", e);
+        } catch (SessionNotFoundException e) {
```
elguardian, line 393: not equivalent. lossing info in other exceptions

elguardian:
TLGTM without going too deep into the details. 
Just to double check:
You are catching a problem in the RuntimeManager where two members in cluster are operating over the same process instance causing a race condition and one tx is deleting the process instance ? @fjtirado

gmunozfe:
jenkins retest this

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java
```diff
@@ -66,7 +67,10 @@ public PersistentProcessInstance findProcessInstanceInfo(Long processId) {
     	if( this.pessimisticLocking ) { 
             return em.find( ProcessInstanceInfo.class, processId, lockMode );
         }
-        return em.find( ProcessInstanceInfo.class, processId );
+        logger.trace("Reading process instance info {} with em {}", processId, em);
+        ProcessInstanceInfo pi = em.find(ProcessInstanceInfo.class, processId);
+        logger.trace("Process instance info read {}", pi);
```
gmunozfe, line 72: To avoid computational time when trace level is not enabled -basically for the find operation-, you could include them inside an if clause, wdyt?:
```suggestion
        if (logger.isTraceEnabled()) {
          logger.trace("Reading process instance info {} with em {}", processId, em);
          ProcessInstanceInfo pi = em.find(ProcessInstanceInfo.class, processId);
          logger.trace("Process instance info read {}", pi);
      }
```

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
```diff
@@ -191,15 +194,35 @@ public void signalEvent(String type, Object event) {
...
-            RuntimeEngine[] activeEngines = currentlyActive.values().toArray(new RuntimeEngine[currentlyActive.size()]);
-            for (RuntimeEngine engine : activeEngines) {
-                Context<?> context = ((RuntimeEngineImpl) engine).getContext();
+            @SuppressWarnings("unchecked")
+            Entry<Object, RuntimeEngine> activeEngines[] = currentlyActive.entrySet()
+                    .toArray(new Entry[currentlyActive.size()]);
+            Set<Object> enginesToDelete = new HashSet<>();
+            for (Entry<Object, RuntimeEngine> engine : activeEngines) {
+                RuntimeEngineImpl engineImpl = (RuntimeEngineImpl) engine.getValue();
+                if (engineImpl.isDisposed() || engineImpl.isInvalid()) {
```
gmunozfe, line 205: add a not-null check to skip possible null pointer
```suggestion
                if (engineImpl==null) {
                     continue;
                }
                if (engineImpl.isDisposed() || engineImpl.isInvalid()) {
```

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
```diff
@@ -191,15 +194,35 @@ public void signalEvent(String type, Object event) {
...
+                    } catch (org.drools.persistence.api.SessionNotFoundException ex) {
+                        logger.warn(
+                                "Signal event cannot proceed because of session not found exception {} for engine {}",
+                                ex.getMessage(), engineImpl.getKieSessionId());
+                        enginesToDelete.add(engine.getKey());
+                    }
                 }
             }
+            if (!enginesToDelete.isEmpty()) {
+                currentlyActive.keySet().removeAll(enginesToDelete);
```
gmunozfe, line 224: you could log the enginesToDelete for tracking (debug level if you don't want to make it verbose)

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
```diff
@@ -191,15 +194,35 @@ public void signalEvent(String type, Object event) {
...
         // process currently active runtime engines
         Map<Object, RuntimeEngine> currentlyActive = local.get();
+
         if (currentlyActive != null && !currentlyActive.isEmpty()) {
-            RuntimeEngine[] activeEngines = currentlyActive.values().toArray(new RuntimeEngine[currentlyActive.size()]);
-            for (RuntimeEngine engine : activeEngines) {
-                Context<?> context = ((RuntimeEngineImpl) engine).getContext();
+            @SuppressWarnings("unchecked")
+            Entry<Object, RuntimeEngine> activeEngines[] = currentlyActive.entrySet()
+                    .toArray(new Entry[currentlyActive.size()]);
```
gmunozfe, line 201: To avoid SuppressWarnings annotation, could it be like this?
```suggestion
            Map.Entry<Object, RuntimeEngine> activeEngines[] = currentlyActive.entrySet()
            .toArray(new Map.Entry[currentlyActive.size()]);
```

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
```diff
@@ -191,15 +194,35 @@ public void signalEvent(String type, Object event) {
...
+            Set<Object> enginesToDelete = new HashSet<>();
+            for (Entry<Object, RuntimeEngine> engine : activeEngines) {
```
gmunozfe, line 203: Same here:
```suggestion
            for (Map.Entry<Object, RuntimeEngine> engine : activeEngines) {
```

gmunozfe (commented):
Some minor comments about tracing, possible null pointers and skipping supresswarnings)

jbpm-persistence/jbpm-persistence-jpa/src/main/java/org/jbpm/persistence/JpaProcessPersistenceContext.java
fjtirado: this will disable the find

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
fjtirado: I do not think it makes anny difference

mareknovotny:
jenkins retest this please

gmunozfe (approved):

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/PerProcessInstanceRuntimeManager.java
baldimir: Hi, sorry for commenting a merged PR, I am not proposing any change. I just want to highlight that Gonzalo's comment actually makes a difference. If we let SuppressWarnings in the code like this, when there is a proposed solution that doesn't need it, the code will get polluted with the annotations. I know it doesn't make difference from functional perspective, however it makes sense from maintenance perspective. For future I would recommend to use the code that doesn't need this annotation.

mareknovotny: @baldimir i guess the suggested change was not having expected result - to not need to suppress the warning. it would be still required or see the compiler warning here.

PR #2319, opened 2023-08-17
[JBPM-10188] Signaling process instances

fjtirado:
Original expression needs to be added as it is to external event listeners. This way it can be converted when persisted.


gmunozfe (approved):
Looks good to me, well done @fjtirado !

PR #2323, opened 2023-08-29
[kie-issues-509] Handle SLA timers during process instance migration

elguardian:
Jira: https://github.com/kiegroup/kie-issues/issues/509


martinweiler (changes requested):
@elguardian great work - thank you! IMO it would be good to also handle process SLAs while we are at this, wdyt? I added two test cases that cover migration of both user task SLA and process SLA processes. Could you please take a look? Thanks!

PR #2335, opened 2023-09-15
[JBPM-10197] Handle TimerMappingInfo in error scenarios

fjtirado:

[link](https://issues.redhat.com/browse/JBPM-10197)

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
@@ -17,12 +17,14 @@
 package org.jbpm.services.ejb.timer;
 
 import java.io.ByteArrayInputStream;
+import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.Serializable;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Function;
 
+import javax.ejb.EJBException;
 import javax.ejb.Timer;
 import javax.ejb.TimerHandle;
 import javax.naming.InitialContext;
@@ -77,13 +79,18 @@ public class EjbSchedulerService implements GlobalSchedulerService {
 		// check if given timer job is marked as new timer meaning it was never scheduled before, 
 		// if so skip the check by timer name as it has no way to exist
 		if (!ctx.isNew()) {
-		    jobInstance = getTimerJobInstance(jobName);
-		    if (jobInstance == null) {
-		        jobInstance = scheduler.getTimerByName(jobName);
+            try {
+                jobInstance = getTimerJobInstance(jobName);
+                if (jobInstance == null) {
+                    jobInstance = scheduler.getTimerByName(jobName);
+                }
+            } catch (InvalidEJBTimer ex) {
+                logger.info(
+                        "Avoiding searching all timers since the timer mapping info pointed to a timer that is not longer valid");
+            }
+            if (jobInstance != null) {
+                return jobInstance.getJobHandle();
 		    }
-    		if (jobInstance != null) {
-    			return jobInstance.getJobHandle();
-    		}
 		}
 		jobInstance = globalTimerService.getTimerJobFactoryManager().createTimerJobInstance(
 														job, 
@@ -99,14 +106,22 @@ public class EjbSchedulerService implements GlobalSchedulerService {
 
     @Override
     public boolean removeJob(JobHandle jobHandle) {
-        String uuid = ((EjbGlobalJobHandle) jobHandle).getUuid();
-        final Timer ejbTimer = getEjbTimer(getTimerMappinInfo(uuid));
-        if (TRANSACTIONAL && ejbTimer == null) {
-            // this situation needs to be avoided as it should not happen
+        if (TRANSACTIONAL) {
             return false;
         }
-        JtaTransactionManager tm = (JtaTransactionManager) TransactionManagerFactory.get().newTransactionManager();
+        final String uuid = ((EjbGlobalJobHandle) jobHandle).getUuid();
+        Timer ejbTimer;
         try {
+            ejbTimer = getEjbTimer(getTimerMappinInfo(uuid));
+            if (ejbTimer == null) {
+                return false;
+            }
+        } catch (InvalidEJBTimer ex) {
+            return false;
+        }
+
+        try {
+            JtaTransactionManager tm = (JtaTransactionManager) TransactionManagerFactory.get().newTransactionManager();
             tm.registerTransactionSynchronization(new TransactionSynchronization() {
                 @Override
                 public void beforeCompletion() {
@@ -119,7 +134,6 @@ public class EjbSchedulerService implements GlobalSchedulerService {
                         scheduler.removeJob(jobHandle, ejbTimer);
                     }
                 }
-                
             });
             logger.debug("register tx to remove job {}", jobHandle);
             return true;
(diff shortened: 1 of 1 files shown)


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
```diff
@@ -99,14 +101,18 @@ public JobHandle scheduleJob(Job job, JobContext ctx, Trigger trigger) {
 
     @Override
     public boolean removeJob(JobHandle jobHandle) {
-        String uuid = ((EjbGlobalJobHandle) jobHandle).getUuid();
-        final Timer ejbTimer = getEjbTimer(getTimerMappinInfo(uuid));
-        if (TRANSACTIONAL && ejbTimer == null) {
-            // this situation needs to be avoided as it should not happen
+        if (TRANSACTIONAL) {
+            return false;
+        }
```
martinweiler, line 106: This can lead to duplicate timers, eg. during process instance migration.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
```diff
@@ -129,43 +134,74 @@ public void afterCompletion(int status) {
...
+        try {
+            return ((TimerHandle) new ObjectInputStream(
+                    new ByteArrayInputStream(timerMappingInfo.getInfo()))
+                    .readObject()).getTimer();
+
+        } catch (ClassNotFoundException | IOException e) {
+            logger.warn("Error {} deserializing the timer handle for timermappinginfo {}", e.getMessage(),
+                    timerMappingInfo);
+            deleteTimerMapping(timerMappingInfo);
+            throw new RuntimeException(e);
```
martinweiler, line 166: I don't think this method should throw an exception if something goes wrong during deserialization.

martinweiler:
@fjtirado I tested this PR with the following use case:
1. Create process with timer, using postgres DB and oid based schema
2. Manually delete the pg_largeobject record linked to TimerMappingInfo.info to simulate an issue with the retrieval of the info blob from the database.
3. Invoke `instances/PROCID/timers` API to force reading of timers

**Findings**: If I remove the throw calls in getEjbTimer method, I can see that the TimerMappingInfo gets recreated, which is what I expected. However, it is missing important information, such as the external id and the info column:

Old TimerMappingInfo entry:
```
 id |           externaltimerid            | kiesessionid | processinstanceid | timerid |  uuid  | info  
----+--------------------------------------+--------------+-------------------+---------+--------+-------
  6 | 35c5299a-c167-4ca8-bef9-0743a2f35d7b |           13 |                 3 |       2 | 13-3-2 | 18130
```

New TimerMappingInfo entry:
```
 id | externaltimerid | kiesessionid | processinstanceid | timerid |  uuid  | info 
----+-----------------+--------------+-------------------+---------+--------+------
  7 |                 |           13 |                 3 |       2 | 13-3-2 |    
```

fjtirado:
@martinweiler I think the behaviour you are experiencing (empty externalid and info) is essentially the same that will occur without the patch. 
I mean this is intended to avoid the performance bottleneck of calling getTimerByName when there is not need to so , not the problem with empty external id and info.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
fjtirado: yes, you are right, I changed it after a self-review already

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
```diff
@@ -77,13 +78,18 @@ public JobHandle scheduleJob(Job job, JobContext ctx, Trigger trigger) {
...
-		    if (jobInstance == null) {
-		        jobInstance = scheduler.getTimerByName(jobName);
+            try {
+                jobInstance = getTimerJobInstance(jobName);
+                if (jobInstance == null) {
+                    jobInstance = scheduler.getTimerByName(jobName);
+                }
+            } catch (InvalidEJBTimer ex) {
+                logger.info(
+                        "Avoiding searching all timers since the timer mapping info pointed to a timer that is not longer valid");
```
martinweiler, line 88: I'd prefer to have this logged at debug level and also have the jobName added so that the message provides more context

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
```diff
@@ -129,43 +139,90 @@ public void afterCompletion(int status) {
...
+                    new ByteArrayInputStream(timerMappingInfo.getInfo()))
+                    .readObject()).getTimer();
+        } catch (EJBException e) {
+            logger.warn("Error {} retrieving the EJB timer from timermappinginfo {}", e.getMessage(), timerMappingInfo);
+            deleteTimerMapping(timerMappingInfo);
+            throw new InvalidEJBTimer(e);
         } catch (Exception e) {
-            logger.warn("wast not able to deserialize info field from timer info for uuid");
+            logger.warn("Error {} objetining the timer handle for timermappinginfo {}", e.getMessage(),
+                    timerMappingInfo);
```
martinweiler, line 189: obtaining (?)

albfan:
I try Martin approach to generate the problem.

Definitely, PUT timers when there're two timers active id 1 and id 2, editing id 2 removes from database timer id 1 and id 2 and id 3 refer to second timer but as this was detected on RHPAM 7.12.1, and customer do not have RHPAM-4192 fix,  I get a null listing timers again.

I cannot find another action that removes the timer from database, while keeping it in process instance memory. JBPM-10198 should fix that so if that's the only source for "corrupt" a timer that will not happen again. Then this code is defensive for any situation where timer is removed from database.

This is: If there's a timermappinginfo, but timer cannot be retrieved, there's an error so do not look for timer in table, just create another one.

In that sense, code to deal with an error should not be merged, so that's why I offer the alternative PR #2351

If search for timer is not a blocker like implementing https://issues.redhat.com/browse/EAPSUP-1336 (add a column with external id from app using EJB timers) we can ignore this too as search will be really quick.

Just to make it easy to understand:

```
select 
  id, 
  initial_date, 
  next_date, 
  timer_state,
  encode(decode(info, 'base64'), 'escape') as info,
  (regexp_match(encode(decode(info, 'base64'), 'escape'), '([0-9]+)-([0-9]+)-(.*-[0-9]+)')) as ids, 
  (regexp_match(encode(decode(info, 'base64'), 'escape'), '[0-9]+-([0-9]+)-.*-[0-9]+'))[1] as pid,
  (regexp_match(encode(decode(info, 'base64'), 'escape'), '[-a-z0-9]+_[0-9]+.[0-9]+.[0-9]+')) as container
from jboss_ejb_timer 
```

This is a query that search in milliseconds what timer is related with what timerID created in RHPAM, but is a hack, we want EJB timers to offer this as a feature

albfan:
@mareknovotny can we agreed to over #2351 as a solution for customer with this problems instead of this defensive code?

mareknovotny:
not mine decision, @krisv @fjtirado @martinweiler @gmunozfe are reviewers

PR #2343, opened 2023-10-02
[kie-issues-597] Test case enhancement to run migration with org.jbpm.timer.disableUnmarshallerRegistration=true|false

martinweiler:
Test case enhancement for https://github.com/apache/incubator-kie-issues/issues/597


martinweiler:
Closing as https://github.com/apache/incubator-kie-issues/issues/597 is closed now.

PR #2346, opened 2023-10-05
[incubator-kie-issues-597] PIM fails to handle timers when org.jbpm.timer.disableUnmarshallerRegistration is set to true

elguardian:
Issue https://github.com/apache/incubator-kie-drools/pull/5547

depends on https://github.com/apache/incubator-kie-drools/pull/5547


elguardian:
Jenkins rerun this please

PR #2347, opened 2023-10-16
RHPAM-4822 find EventNode by unique id

albfan:
**JIRA**: [RHPAM-4822](https://github.com/kiegroup/jbpm/compare/main...albfan:jbpm:RHPAM-4822)

Avoid classcastexception finding event nodes by unique id

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventNodeInstance.java
+++ b/jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventNodeInstance.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.regex.Matcher;
+import java.util.Collection;
 
 import org.jbpm.process.core.context.variable.VariableScope;
 import org.jbpm.process.core.event.EventTransformer;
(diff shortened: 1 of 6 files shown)


mareknovotny:
ok to test

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/EventNodeInstance.java
```diff
@@ -141,7 +151,41 @@ public void removeTimerListeners() {
...
+    /**
+     * Search node by nodeinstanceid and uniqueId, cause node definition id might change
+     * @param id
+     * @param nodeInstanceId
+     * @return
+     */
+    private Node getNodeByNodeInstanceId(long id, long nodeInstanceId) {
+        Node node = null;
+        RuntimeDataService service = (RuntimeDataService) ServiceRegistry.get().service(ServiceRegistry.RUNTIME_DATA_SERVICE);
+        Collection<NodeInstanceDesc> nodes = service.getProcessInstanceNodeInstanceHistory(id, nodeInstanceId, new QueryContext(0, 1));
```
elguardian, line 178: incorrect either. the engine should not access the runtime data service. this goes against the design.

elguardian (changes requested):
Rejected. this entangles the design of the system. The are ways to access the node by uniqueId already

PR #2351, opened 2023-10-26
JBPM-10197 Parameter org.jbpm.ejb.timer.lookup to disable search for timers

albfan:
In databases with lot of timers, search for timer is a bottleneck. Is better to have a failed timer executed than search for it

Alternative to https://github.com/kiegroup/jbpm/pull/2335

This should be resolved with a feature for timers: externalId: https://issues.redhat.com/browse/EAPSUP-1336

ping @martinweiler @fjtirado


albfan:
@mareknovotny this PR do not solve the root cause, just allow users with lot of timers to ensure they will never reach that code.

As change in behaviour depends on parameter, we should be able to merge and document this workaround.

martinweiler:
@albfan I don't understand why you need this flag and also, if it is enabled, that it can lead to duplicate timers getting created. Sure, you'd avoid the costly call to getTimerByName in case the timer retrieval from TimerMappingInfo fails, but it would result in a new timer being scheduled.

albfan:
Yes, that's the idea, if you reach the point to need to look for a timer, a duplicate one is way better than traverse the timers table.

Duplicate timers is problematic, traverse timer table is blocking the app.

We think the origin for this problem could be JBPM-10198

fjtirado:
See alternative here https://github.com/kiegroup/jbpm/pull/2369

PR #2359, opened 2023-11-07
[JBPM-10208] Signalling engine only once

fjtirado:
This keeps track of the engines that has already being signalled to avoid signalling the same engine twice as part of the same signalEvent call.


fjtirado:
jenkins retest this

elguardian:
I need more context about this one. Don't really understand the background or have a test case.

gmunozfe (approved):
Looks good to me, effective solution though I was also wondering why is happening this duplication. A test can be added later as we verified with the reproducer that fixes the issue. Fantastic work @fjtirado !

PR #2143, opened 2022-05-03
[JBPM-10072] EJBTimerScheduler.getTimerByName inefficient with large number of timers

elguardian:
Jira: https://issues.redhat.com/browse/JBPM-10072

PR #2362, opened 2023-11-14
[JBPM-10209] Switching to CMT

fjtirado:
EJBTimerScheduler bean was using BMT. 
A BTM bean will always suspend the incoming transaction, therefore any operation within the bean, for example, scheduling a timer, will be performed in a different transaction. 
This PR switch back to CMT, removing the risk of a short delay timer colliding with the parent transaction. 
Since CMT was changed to BMT to be able to control delays and recovery when the timer is executed, as an alternative, we use CMT with requires_new to force a new transaction (achieving the same result than by handling UserTransaction). Also, the recoverTimerJobInstance was refactor a bit to make it clear that essentially, whatever the scenario (sessionNotFound, intervalTrigger or failed times) the recovery is executed within another transaction (as the regular call)

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -65,6 +65,8 @@ public class EJBTimerScheduler {
     private static final Integer TIMER_RETRY_LIMIT = Integer.parseInt(System.getProperty("org.kie.jbpm.timer.retry.limit", "3"));
 
 	private static final Integer OVERDUE_WAIT_TIME = Integer.parseInt(System.getProperty("org.jbpm.overdue.timer.wait", "20000"));
+	
+	private static final Integer OVERDUE_CHECK_TIME = Integer.parseInt(System.getProperty("org.jbpm.overdue.timer.check", "200"));
 
 	private boolean useLocalCache = Boolean.parseBoolean(System.getProperty("org.jbpm.ejb.timer.local.cache", "false"));
 
@@ -92,30 +94,24 @@ public class EJBTimerScheduler {
         EjbTimerJob timerJob = (EjbTimerJob) timer.getInfo();
         TimerJobInstance timerJobInstance = timerJob.getTimerJobInstance();
         logger.debug("About to execute timer for job {}", timerJob);
-
-        String timerServiceId = ((EjbGlobalJobHandle) timerJobInstance.getJobHandle()).getDeploymentId();
-
         // handle overdue timers as ejb timer service might start before all deployments are ready
         long time = 0;
-        while (TimerServiceRegistry.getInstance().get(timerServiceId) == null) {
-            logger.debug("waiting for timer service to be available, elapsed time {} ms", time);
-            try {
-                Thread.sleep(500);
-            } catch (InterruptedException e) {
-                e.printStackTrace();
+        try {
+            while (TimerServiceRegistry.getInstance().get(((EjbGlobalJobHandle) timerJobInstance.getJobHandle()).getDeploymentId()) == null) {
+                logger.debug("waiting for timer service to be available, elapsed time {} ms", time);
+                Thread.sleep(OVERDUE_CHECK_TIME);
+                time += OVERDUE_CHECK_TIME;
+                if (time > OVERDUE_WAIT_TIME) {
+                    logger.debug("No timer service found after waiting {} ms", time);
+                    break;
+                }
             }
-            time += 500;
-
-            if (time > OVERDUE_WAIT_TIME) {
-                logger.debug("No timer service found after waiting {} ms", time);
-                break;
-            }
-        }
+        } catch (InterruptedException e) {
+            logger.warn("Thread has been interrupted", e);
+            Thread.currentThread().interrupt();
+        }        
         try {
             invokeTransaction(this::executeTimerJobInstance, timerJobInstance);
-        } catch (SessionNotFoundException e) {
-            logger.warn("Process instance is not found. More likely already completed. Timer {} won't be recovered", timerJobInstance, e);
-            removeUnrecoverableTimer(timerJob, timer);
         } catch (Exception e) {
             recoverTimerJobInstance(timerJob, timer, e);
         }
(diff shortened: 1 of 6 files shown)


jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -97,112 +94,78 @@ public void executeTimerJob(Timer timer) {
...
-        } catch (Exception e1) {
-            logger.warn("There was a problem during timer removal {}", ejbTimerJob.getTimerJobInstance(), e1);
-        }
-    }
-
-    private void recoverTimerJobInstance(EjbTimerJob ejbTimerJob, Timer timer, Exception e) {
-        if (isSessionNotFound(e)) {
-            logger.warn("Trying to recover timer. Not possible due to process instance is not found. More likely already completed. Timer {} won't be recovered", ejbTimerJob.getTimerJobInstance(), e);
+    private void recoverTimerJobInstance(EjbTimerJob ejbTimerJob, Timer timer,  Exception cause) {
+        Transaction<TimerJobInstance> tx;
```
gmunozfe, line 125: It could be defined when use later

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -97,112 +94,78 @@ public void executeTimerJob(Timer timer) {
...
-    private void removeUnrecoverableTimer(EjbTimerJob ejbTimerJob, Timer timer) {
-        try {
-            Transaction<TimerJobInstance> tx = timerJobInstance -> {
-                if (!this.removeJob(timerJobInstance.getJobHandle(), timer)) {
-                    logger.warn("Session not found for timer {}. Timer could not removed.", ejbTimerJob.getTimerJobInstance());
-                }
-            };
-            transaction(tx, ejbTimerJob.getTimerJobInstance());
-        } catch (Exception e1) {
-            logger.warn("There was a problem during timer removal {}", ejbTimerJob.getTimerJobInstance(), e1);
```
gmunozfe, line 146: This warning is not kept in the refactoring

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -97,112 +94,78 @@ public void executeTimerJob(Timer timer) {
...
+    private void recoverTimerJobInstance(EjbTimerJob ejbTimerJob, Timer timer,  Exception cause) {
+        Transaction<TimerJobInstance> tx;
+        if (isSessionNotFound(cause)) {
             // if session is not found means the process has already finished. In this case we just need to remove
             // the timer and avoid any recovery as it should not trigger any more timers.
-            removeUnrecoverableTimer(ejbTimerJob, timer);
-            return;
+            tx = timerJobInstance -> {
+                logger.warn("Trying to recover timer. Not possible due to process instance is not found. More likely already completed. Timer {} won't be recovered", timerJobInstance, cause);
+                if (!removeJob(timerJobInstance.getJobHandle(), timer)) {
```
gmunozfe, line 131: Could it throw an exception?

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
```diff
@@ -99,34 +98,12 @@ public JobHandle scheduleJob(Job job, JobContext ctx, Trigger trigger) {
...
-        try {
-            tm.registerTransactionSynchronization(new TransactionSynchronization() {
-                @Override
-                public void beforeCompletion() {
-                }
-
-                @Override
-                public void afterCompletion(int status) {
-                    if (status == TransactionManager.STATUS_COMMITTED) {
-                        logger.debug("remove job {} after commited", jobHandle);
```
gmunozfe, line 118: This trace is lost with the refactoring

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -97,112 +94,78 @@ public void executeTimerJob(Timer timer) {
...
-                };
-                transaction(tx, ejbTimerJob.getTimerJobInstance());
-            } catch (Exception e1) {
-                logger.warn("Could not schedule the interval trigger {}", ejbTimerJob.getTimerJobInstance(), e1);
-            }
-            return;
+            tx = timerJobInstance -> {
+                logger.warn("Execution of time failed Interval Trigger failed. Skipping {}", timerJobInstance);
+                if (removeJob(timerJobInstance.getJobHandle(), null)) {
+                    internalSchedule(timerJobInstance);
```
gmunozfe, line 142: Could it throw an exception that in the previous implementation was caught?
`logger.warn("Could not schedule the interval trigger {}", ejbTimerJob.getTimerJobInstance(), e1);`

gmunozfe (commented):
Some minor comments and questions (mainly about tracing and catching exceptions made in a different way than in the previous implementation).
Reproducer works like a charm with this patch.
Superb work @fjtirado !

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java
fjtirado: Traces added

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -70,17 +65,19 @@ public class EJBTimerScheduler {
     private static final Integer TIMER_RETRY_LIMIT = Integer.parseInt(System.getProperty("org.kie.jbpm.timer.retry.limit", "3"));
 
 	private static final Integer OVERDUE_WAIT_TIME = Integer.parseInt(System.getProperty("org.jbpm.overdue.timer.wait", "20000"));
+	
+	private static final Integer OVERDUE_CHECK_TIME = Integer.parseInt(System.getProperty("org.jbpm.overdue.timer.check", "200"));
```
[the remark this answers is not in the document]
fjtirado: yes, thats intentional, 500 is too large in my opinion, timerService should not take so long to start

elguardian:
ok. some notes in there. changing to BMT won't allow to control ejb timer retry mechanism and this will cause an escalation of timers and logs in the user. (infinite numbers of retries when you are using interval triggers)

BMT does not suspend any transaction in this case as this is a timer timeout call (there was no previous tx in flight)
There is no difference regarding tx or not if you use BMT or CMT in this case regarding any deadlock.

Require NEW is the same as Required in this case when you call an ejb timeout method because there was not any tx in flight

CMT or BMT are just demarcation of the tx and the timeout should only happen in one place so it is unlikely to be the root cause of a deadlock as this is the first demarcation in the call. (tx.begin cannot cause the deadlock)

To cause a a deadlock you need two different resources and try to hold them in one tx. Let's try to simulate an scenario:

1 tx -> BMT tx demarcation holding resource A then resource B
2 tx -> BMT tx demarcation hodling resource B and the resource A

or any different
1 tx -> CMT tx demarcation holding resource A then resource B
2 tx -> CMT tx demarcation holding resource B then resource A

calling tx being it is not going to fix the situation in a deadlock. You need to see which tables are involved in the deadlock first to see where the problem lies. (deadlock analisys in the database. This depends on the database but usually it comes with a tool)

@martinweiler

fjtirado:
@elguardian 
BMT suspend the transaction, see attached picture
image(5)


This fix was tested with two reproducers and in customer environmment and it prevents the deadlock.

martinweiler:
@fjtirado are the reproducers you mentioned available somewhere?

fjtirado:
> @fjtirado are the reproducers you mentioned available somewhere?

@martinweiler Gonzalo attached the reprrodcued BPMN to the JIRA

gmunozfe:
> > @fjtirado are the reproducers you mentioned available somewhere?
> > 
> > @martinweiler Gonzalo attached the reprrodcuer [BPMN](https://issues.redhat.com/secure/attachment/13076271/NotifProcess.bpmn) to the JIRA

@martinweiler here is the automated test (`timerIsNotTriggerUntilTaskCommitted`) with the reproducer (we're using 2 kie-servers and PostgreSQL containers)
https://github.com/kiegroup/jbpm-playground/compare/master...gmunozfe:jbpm-playground:timer_test

martinweiler:
@gmunozfe @fjtirado I ran the reproducer locally, and confirm that there are no more deadlocks with your fix. I also ran the reproducer from JBPM-8688 (as this is where BMT was introduced) and it still works. I also tested the reproducer from RHPAM-3826 / RHPAM-3574, as these are related to the removeUnrecoverableTimer method which has been removed, but did not find an issue with this test either. Great work @fjtirado 

My main overall concern is that this is quite a significant change for a micro release, and given the history of issues we had to deal with in regards to timers, that we will introduce further issues down the road, and/or introduce a regression (which might not be caused by the tests).

gmunozfe (approved):

martinweiler:
@elguardian @fjtirado regarding the discussion on CMT vs BMT, my knowledge in that area is limited, however I'd like to share the following. On 7.13.4/8.0.4 (unpatched), the timer creation happens *before* the transaction of the caller is ended:
```
2023-11-16 10:28:06,645 DEBUG [org.jboss.as.ejb3.timer] (default task-1) Scheduling a single action timer [id=76d18752-b1a6-47bd-9955-6a01a7855e97 timedObjectId=kie-server.kie-server.EJBTimerScheduler auto-timer?:false persistent?:true timerService=org.jboss.as.ejb3.timerservice.TimerServiceImpl@19c26ae1 previousRun=null initialExpiration=Thu Nov 16 10:28:07 MST 2023 intervalDuration(in milli sec)=0 nextExpiration=Thu Nov 16 10:28:07 MST 2023 timerState=ACTIVE info=EjbTimerJob [timerJobInstance=GlobalJpaTimerJobInstance [timerServiceId=deadlock-test:1.0.0-SNAPSHOT-timerServiceId, getJobHandle()=EjbGlobalJobHandle [uuid=462-430-timer-1]]]] starting at 972 milliseconds from now
2023-11-16 10:28:10,651 INFO  [stdout] (default task-1) === end sleep Exception Before Task3 ===
```

With the changed code, the timer creation happens at the end of the caller transaction (*after* the sleep task):
```
2023-11-16 10:15:58,676 INFO  [stdout] (default task-1) === end sleep Exception Before Task3 ===
2023-11-16 10:15:58,739 DEBUG [org.jboss.as.ejb3.timer] (default task-1) commit timer creation: [id=6d500923-80be-438a-a0c3-0c292fe6bb5e timedObjectId=kie-server.kie-server.EJBTimerScheduler auto-timer?:false persistent?:true timerS
ervice=org.jboss.as.ejb3.timerservice.TimerServiceImpl@7fac3af7 previousRun=null initialExpiration=Thu Nov 16 10:15:55 MST 2023 intervalDuration(in milli sec)=0 nextExpiration=Thu Nov 16 10:15:55 MST 2023 timerState=CREATED info=Ejb
TimerJob [timerJobInstance=GlobalJpaTimerJobInstance [timerServiceId=deadlock-test:1.0.0-SNAPSHOT-timerServiceId, getJobHandle()=EjbGlobalJobHandle [uuid=458-429-timer-1]]]]
2023-11-16 10:15:58,740 DEBUG [org.jboss.as.ejb3.timer] (default task-1) Scheduling a single action timer [id=6d500923-80be-438a-a0c3-0c292fe6bb5e timedObjectId=kie-server.kie-server.EJBTimerScheduler auto-timer?:false persistent?:t
rue timerService=org.jboss.as.ejb3.timerservice.TimerServiceImpl@7fac3af7 previousRun=null initialExpiration=Thu Nov 16 10:15:55 MST 2023 intervalDuration(in milli sec)=0 nextExpiration=Thu Nov 16 10:15:55 MST 2023 timerState=ACTIVE
 info=EjbTimerJob [timerJobInstance=GlobalJpaTimerJobInstance [timerServiceId=deadlock-test:1.0.0-SNAPSHOT-timerServiceId, getJobHandle()=EjbGlobalJobHandle [uuid=458-429-timer-1]]]] starting at -3,093 milliseconds from now
```

This prevents the timer execution to clash with the active process execution, so I'd say the new behavior is correct.

What concerns me is that this change is quite significant and happening in a very sensitive area, and it seems this PR got merged in a rush. cc @krisv @porcelli

porcelli:
I share similar concern of @martinweiler in regard the rush to merge. I felt that we were still discussing it.... and all of sudden it got merged.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -218,30 +181,16 @@ private boolean isSessionNotFound(Exception e) {
...
-        } catch (Exception e) {
-            try {
-                utx.rollback();
-            } catch (Exception re) {
-                logger.error("transaction could not be rolled back", re);
-            }
-
-            throw e;
-        }
+    @TransactionAttribute(value = TransactionAttributeType.REQUIRES_NEW)
```
martinweiler, line 187: REQUIRES_NEW seems redundant here. Doesn't the container already start a tx on timer start? To me it seems REQUIRED should be sufficient here.

fjtirado: You are right about container creating a transaction already, however, here we want to split the "sunny day" path (the timer task execution) and the "rainy day" path (tha handling of any excepton launhed by the timer task) in two different transactions. It is a bit "heavyweight" but safer. However see change [here](https://github.com/kiegroup/jbpm/pull/2365/files#diff-5669039d6beade81bd11322820ff3c35dca83d5ada6986d268e1ef9879f06a65R192) to ensure rollback is happening.

fjtirado:
> I share similar concern of @martinweiler in regard the rush to merge. I felt that we were still discussing it.... and all of sudden it got merged.

@porcelli @martinweiler 
There was no rush, but quick resolution because this PR fixes a critical customer issue that has being around for months. The issues was caused by this change  https://github.com/kiegroup/jbpm/pull/1520/files#diff-5669039d6beade81bd11322820ff3c35dca83d5ada6986d268e1ef9879f06a65R53 

There are still issues with timers, not related with this PR, that we are addressing with https://github.com/kiegroup/jbpm/pull/2365. These issues were introduced by https://github.com/kiegroup/jbpm/pull/2143/files#diff-155546a1876a109b453b736d70a280e1721992e71ab093adfe42949b064aa852R104-R107

porcelli:
@fjtirado i understand that these changes aimed to fix a problem. but concerns were raised and i felt that we didn't have enough time to properly discuss.

Sorry to disagree on terms addressing critical customer issue... this is the open source project, and there are better mechanisms to address downstream needs to provide quick fixes to customers.

fjtirado:
> Sorry to disagree on terms addressing critical customer issue... this is the open source project, and there are better mechanisms to address downstream needs to provide quick fixes to customers.

 @porcelli The discussion was not about the content of the code changes, but about this PR being part of a micro release, which is not related with open source either, as far as I know. Quoting Martin "_My main overall concern is that this is quite a significant change for a micro release, and given the history of issues we had to deal with in regards to timers, that we will introduce further issues down the road, and/or introduce a regression (which might not be caused by the tests)_.". Since to officially release the patch to the customer. it MUST be merged into main, I think the discussion is beyond this PR.

dgutierr:
I understand all the shared concerns. It is clear this is a very sentive area in the code. We are confident about the quality of the changes despite the fact that this area is inherently sensitive and still not fully covered neither by the previous implementation nor the new one. The changes merged provide not only the same coverage as before but it also includes now manual test cases reproducers that were not working with the previous implementation and are working now. So This PR does provide even better coverage than before. 

Regarding the suitability to merge this PR into main, there was no rush to merge in my opinion, since the team (@gmunozfe @fjtirado) has been more than 2 months  working on reproducers, testing, and finally bringing a reasonable fix. Besides, we continue improving the code quality with some other edge cases that we found from the past implementation, so out aim is to still bring some more improvements and  extra coverage going forward.

PR #2369, opened 2023-11-27
[JBPM-10097] Improving performance of getTimerByName

fjtirado:
**JIRA**:

[link](https://issues.redhat.com/browse/JBPM-10197)

A huge hack, because it made certain assumptions that should not be made if TimerService API was functional for the usage we are giving it (which is even a huger hack in my opinion ;)), but maybe worthy  for users having performance issues with` getTimerByName()`. 

If Timer implementation is wildfly and it is using DB as persistence mechanism and if DB is oracle or postgresql, we can actually try to reduce the number of timers to be checked in getTimerByName using a bit of reflection and a couple of native queries. The hacky part is isolated into a single class. 

The Wildfly hacky assumptions are:

1.  TimerService implementation wildffy class name
2.  The field containing the persistence instance is called `persistence`
3.  The class name  of the instance assigned to field persistence
4.  The method used to deserialize info column is called deSerialize
5.  That there is a table called ejb_timer with a string column called info. 

If assumptions 1), 2) and 3) are changed by a future wildfly release, linear search will be done (so we are covered on that regard). 4) and 5) are trickier, since the exception will get propagated. 

**DISCLAIMER**
I did not want to do this, but I consider this approach better than potentially creating duplicates (see closed PRs over the same JIRA) or performing a linear search over  thousands of timers once we know how Wildfly is working.  If Application Server is not widlfly, or the persistence is not on DB or widfly change the class name, we will revert to linear search, so it is just a hack that boost performance of a certain wildfly setup and a couple of DBs, it does not imply losing any existing functionality if the setup is not matched . 

**PS**
If we are ok with the possibility of assuming Wildfly, my idea, before moving into "ready for review" and once this has been tested with an existing problematic setup, is to switch to ServiceLoader mechanism and try to add as much DBs as possible, so the classes related with Wildfly are not in the EJB timer module, but in a separate one that might be added only for wildfly setup (this will allow including the Wildfly dependencues, the usage of instanceof and even calling the deSerialize method through direct call). Before that I want to verify that this approach really boost performance.

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/DefaultEJBTimerRetriever.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/DefaultEJBTimerRetriever.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.services.ejb.timer;
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+import javax.ejb.Timer;
+import javax.ejb.TimerService;
+
+class DefaultEJBTimerRetriever extends EJBTimerRetriever<Void> {
+    
+    protected DefaultEJBTimerRetriever(TimerService timerService) {
+        super(timerService);
+    }
+
+    @Override
+    public Collection<Object> getTimers(String jobName, Void accepted) {
+        return timerService.getTimers().stream().map(Timer::getInfo).collect(Collectors.toList());
+    }
+}
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerDB.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerDB.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.services.ejb.timer;
+
+enum EJBTimerDB {
+    ORACLE("select info from jboss_ejb_timer where utl_raw.cast_to_varchar2(utl_encode.base64_decode(utl_raw.cast_to_raw(dbms_lob.substr(info,2000,1)))) like '%?%'"),
+    POSTGRESQL("select info from jboss_ejb_timerwhere decode(info, 'base64') like '%?%'");
+    
+    private final String query;
+    
+    private EJBTimerDB (String query) {
+        this.query = query;
+    }
+    
+    public String getQuery() {
+        return query;
+    }
+}
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerRetriever.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerRetriever.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2023 Red Hat, Inc. and/or its affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.jbpm.services.ejb.timer;
+
+
+import java.util.Collection;
+import java.util.Optional;
+
+import javax.ejb.TimerService;
+
+import org.jbpm.process.core.timer.impl.GlobalTimerService;
+
+public abstract class EJBTimerRetriever<T>  {
+    
+     protected final TimerService timerService;
+     
+     protected EJBTimerRetriever (TimerService timerService) {
+         this.timerService = timerService;
+     }
+    
+     @SuppressWarnings("squid:S1172")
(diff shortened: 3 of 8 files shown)


elguardian:
This is not really a good approach IMO you are going to cause more problem that anything and it is completely unnecesary. As performance problems was fixed storing the ejb timer info in the table. It is standard and and supported by Java EE environments.Also it supports clustering properly.

elguardian (changes requested):
See my comment above.

fjtirado:
> This is not really a good approach IMO you are going to cause more problem that anything and it is completely unnecesary. As performance problems was fixed storing the ejb timer info in the table. It is standard and and supported by Java EE environments.Also it supports clustering properly.

I agree except for the necesity (if it was not necessary, I wont be coding this ;)). The reason we are proposing this hack is because the EJBTimerMapping solution is far from bulletproof (there are users seeing lot of EJBTimerMapping entryies pointing to a null external timer)  and, in that situation, the system collapses performing the serial search when timers are registered at startup (since there is an exception processing the EJBTimerMapping, the system performs the linear seach for all the id null entries). Any other solution that avoid that search (see other closed PRs for this JIRA) will potentially lead to duplicate timers, so breaking encapsulation (which we can hide anyway by using service loader) seems one possibiilty (the other option being manually rebuilding EJBTimerMappingInfo cache offline)

elguardian:
We are aware of the problem but the solution is not following up with a hack but fix properly the null (just an effect of another problem). If that happens means there is something wrong with it (root  cause not clear in here). Walking this path it is not a solution for other app servers or even a real thing as there configuration in the eap could make this not work (like using a different DB or other persistent solutions) within the ejb timer subsystem. So this solution will work in a very specific cases.
Necessity is not defined by if this piece of code can fix (more likely just mitigate the issue) but fix the solution in place is correct and make it behave correctly.
If you have a solution that is compliant with the spec and should work in any app server, we should strive to make the solution work properly and not circumvent with another piece of code that will only work in specific cases as you are not accessing the ejb db timer solution.
I am seeing lately very vague analysis in very complex portioins of code that could lead to potential bit problems. We should analyse the problem carefully and make a proper diagnose of the problem (root cause) and no the effeects (null).

@martinweiler @porcelli ^^ be aware of this as this solution should not be merged.

martinweiler:
I agree with @elguardian that introducing a solution specific to only a subset of the supported databases is not meeting the supportability requirements of the downstream product, and is thus problematic to be merged even upstream. 
@fjtirado your comment highlights issues with the TimerMappingInfo logic. Are you able to pinpoint those so that these can be fixed?

albfan:
Solution is already provided: https://issues.redhat.com/browse/EAPSUP-1336

Keep an index of timers that can be lost is an smell coding. That shouldn't be a problem if traverse all timers is not that slow, but it is: For each N active timer, due to column info is not cached, JBPM needs to do N select queries to get that info (really slow) just to read into info column the index JBPM knows.

That issue propose to be able to search for that know index (the usual concept of an externalid) as EJB timers is just a service to provide timers to external apps, it should manage the concept of real index the external app uses for timers.

This hack is just the evidence that TImerMappingInfo should be removed, and that externaId should be added to EJB timers spec

The goal here is not to avoid getTimersByName (basically what TimerMappingInfo does) but point out that we cannot have a method that slown down JBPM because we have millions of timers.

I tried to convince engineers of EJB timers about this obvious change on spec (the hack just prove it) but there's no progress on it

fjtirado:
@elguardian @martinweiler @porcelli  We are not planning to merge this. As I already mention (twice) it is a hack that can be used as specific patch for specific users, nothing more than that. 
@albfan Analysing why TimerMappingInfo was corrupted was not feasible since we do not have all the required information to do that (we do not even know if the table is corrupted because of manual interaction, because of a migration or because a bug). We just know that somehow it becomes corrupted. 
@elguardian I really appreciate your constructive criticism, but the word lately related with timer changes is not accurate, since the change (which should not have being done) to move from BMT to CMT (bases of the false assumption that transactions are inherited by BMT, see https://github.com/kiegroup/jbpm/pull/2362#issuecomment-1812797584 and https://github.com/kiegroup/jbpm/pull/1520/files#diff-5669039d6beade81bd11322820ff3c35dca83d5ada6986d268e1ef9879f06a65R53) was done a few years ago  and the change to add TimerMappingInfo (which can also be considered a hack, because it is persistence cache and introduced another set of issues, see one [example](https://issues.redhat.com/browse/JBPM-10088) was done a year ago.

fjtirado:
@martinweiler @elguardian 
A user reported that, for several rows in TimerMappingInfo, both externalTimerId and info columns are null. 
The only way I found this can happen, considering the following code https://github.com/kiegroup/jbpm/blob/main/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java#L211-L263 is if GlobalJpaTimerJobInstance has externalTimeId and getTimerInfo returns null. 
For EJB timers, externalTimerId and info might not be set  if hastNextFireTime is null here https://github.com/kiegroup/jbpm/blob/main/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java#L205
Also, I have doubts about this logic https://github.com/kiegroup/jbpm/blob/main/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EjbSchedulerService.java#L71-L79
If ctx.isNew is false, what happens here https://github.com/kiegroup/jbpm/blob/main/jbpm-flow/src/main/java/org/jbpm/process/instance/timer/TimerManager.java#L278, then, if the timer is not found, a new one will be created and the performance will be linear in any case, because it has to check all existing timers before triggering a new one.

elguardian:
@fjtirado my concern about changing from BMT to CMT was related to transaction demarcation. There is nothing in my observations related to inheriting anything (I am talking about tx demarcation). I did never said that. I would suggest to stay on the topic instead answering to something that noone was putting in question.

The main problem of CMT is that you will lose control of retry mechanism and wildly does have a strategy in place to deal with problems (two policies actually) one coming from the spec (immediate automatic retry) and the other coming from the app server itself (several retries after some time). Actually that functionality is lost or broken at this point.

Please read again my observation again (https://github.com/kiegroup/jbpm/pull/2362#issuecomment-1812724388) which is the first sentence.

Related to TimerMappingInfo  is a solution compliant with the ejb spec and respected the design of ejb. It is a way to introduce a way to exchange information among cluster members so it is not a cache

fjtirado:
> @fjtirado my concern about changing from BMT to CMT was related to transaction demarcation. There is nothing in my observations related to inheriting anything (I am talking about tx demarcation). I did never said that. I would suggest to stay on the topic instead answering to something that noone was putting in question.
> 
> The main problem of CMT is that you will lose control of retry mechanism and wildly does have a strategy in place to deal with problems (two policies actually) one coming from the spec (immediate automatic retry) and the other coming from the app server itself (several retries after some time). Actually that functionality is lost or broken at this point.
> 
> Please read again my observation again ([#2362 (comment)](https://github.com/kiegroup/jbpm/pull/2362#issuecomment-1812724388)) which is the first sentence.
> 
> Related to TimerMappingInfo is a solution compliant with the ejb spec and respected the design of ejb. It is a way to introduce a way to exchange information among cluster members so it is not a cache

@elguardian Switching to BMT suspend the transaction that schedules the timer. That causes the timer to be scheduled before the transaction that schedule it is commited (if the timer delay is short enough). The broken Wildfly retry mechanism can be avoided (using CMT or BMT) by not propagating any exception back to the container (so the container thinks the timer execution has been completed successfully)

elguardian:
@fjtirado The tx related to the timer is the one in flight when the timer/consumed is registered so not very certain what is related to suspend thing you are mentioned. You can look into the code yourself

Related to the @Timeout method invocation.
https://github.com/wildfly/wildfly/blob/main/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimedObjectInvokerImpl.java#L72-L85

The callout will invoke interceptors including tx if there is any. If the bean is not BMT it won't invoke any tx so BMT won't suspend any transaction in a timeout call. if that is what you are referring.

In any case the timer service can span it's own tx
https://github.com/wildfly/wildfly/blob/main/ejb3/src/main/java/org/jboss/as/ejb3/timerservice/TimerServiceImpl.java#L417-L455

If the timeout fails and you don't propagate the error, you will lose the timer for sure.

fjtirado:
@elguardian Im **not** referring to the timeout method. Im referring to this one https://github.com/kiegroup/jbpm/blob/main/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java#L202
That one is invoked by the JBPM engine when a timer needs to be scheduled. If we use BMT, the container transaction in which the engine is running is supended, TimerService creates a new one, so a timer is scheduled before the engine has actually comitted the transaction that creates it. It is all described in the PR that switch back to CMT and there is a reproducer for that issue. To continue this discussion is nonsense in my opinion.

PR #2372, opened 2023-12-05
duplicate timer issue

elguardian:
issue: https://github.com/bamoe/private-bamoe-issues/issues/1326


elguardian:
@martinweiler due to changes in the scheduler this collides badly with the current impl.

martinweiler:
@elguardian I think this is no longer needed as the issue was addressed in https://issues.redhat.com/browse/RHPAM-4864

PR #2373, opened 2023-12-06
[private-bamoe-issues#1326] Avoid duplicate timer execution with long running timers and async nodes

martinweiler:
JIRA: https://issues.redhat.com/browse/RHPAM-4864

Add equals/hashCode methods in order to ensure List.contains works properly.

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
@@ -372,6 +372,25 @@ public class GlobalTimerService implements TimerService, InternalSchedulerServic
             return null;
         }
 
+        @Override
+        public int hashCode() {
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + (int) (this.getId() ^ (this.getId() >>> 32));
+            return result;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if ( this == obj ) return true;
+            if ( obj == null ) return false;
+            if ( getClass() != obj.getClass() ) return false;
+            final DefaultJobHandle other = (DefaultJobHandle) obj;
+            if (this.getId() != other.getId())
+                return false;
+            return true;
+        }
+
     }
     
     public static class DisposableCommandService implements InternalLocalRunner {
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
@@ -374,21 +374,16 @@ public class GlobalTimerService implements TimerService, InternalSchedulerServic
 
         @Override
         public int hashCode() {
-            final int prime = 31;
-            int result = 1;
-            result = prime * result + (int) (this.getId() ^ (this.getId() >>> 32));
-            return result;
+            return (int)getId();
         }
 
         @Override
         public boolean equals(Object obj) {
             if ( this == obj ) return true;
-            if ( obj == null ) return false;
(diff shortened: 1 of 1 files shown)


jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
```diff
@@ -372,6 +372,25 @@ public String getUuid() {
             return null;
         }
 
+        @Override
+        public int hashCode() {
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + (int) (this.getId() ^ (this.getId() >>> 32));
```
fjtirado, line 379: Just curious, How do you come with this line? ;)

fjtirado (approved):

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
```diff
@@ -372,6 +372,25 @@ public String getUuid() {
...
+
+        @Override
+        public boolean equals(Object obj) {
+            if ( this == obj ) return true;
+            if ( obj == null ) return false;
+            if ( getClass() != obj.getClass() ) return false;
+            final DefaultJobHandle other = (DefaultJobHandle) obj;
+            if (this.getId() != other.getId())
+                return false;
+            return true;
```
fjtirado, line 391:
```suggestion
          return getId() != ((DefaultJobHandle)obj).getId();
```

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
```diff
@@ -372,6 +372,25 @@ public String getUuid() {
...
+            int result = 1;
+            result = prime * result + (int) (this.getId() ^ (this.getId() >>> 32));
+            return result;
+        }
+
+        @Override
+        public boolean equals(Object obj) {
+            if ( this == obj ) return true;
+            if ( obj == null ) return false;
+            if ( getClass() != obj.getClass() ) return false;
```
fjtirado, line 387:
```suggestion
          if (obj == null || !obj.getClass().equals(getClass())) {
               return false;
          }
```

fjtirado (approved):
Minor suggestions. 
Great work @martinweiler

elguardian (approved):

gmunozfe (commented):
Looks good to me, though @fjtirado suggestions seem fine to be considered

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
gmunozfe: SonarCloud marks it as a "code smell", because it's better not to use the if-then clause, but the boolean result to be returned as indicated by @fjtirado .

martinweiler: @fjtirado thanks - I simplified the method based on your input

jbpm-flow/src/main/java/org/jbpm/process/core/timer/impl/GlobalTimerService.java
```diff
@@ -372,6 +372,25 @@ public String getUuid() {
             return null;
         }
 
+        @Override
+        public int hashCode() {
+            final int prime = 31;
+            int result = 1;
+            result = prime * result + (int) (this.getId() ^ (this.getId() >>> 32));
+            return result;
```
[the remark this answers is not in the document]
martinweiler: @fjtirado thanks - I simplified the method based on your input.

PR #2374, opened 2024-01-05
[RHPAM-4872] Fix RepeatMode=FIXED when deleted records > RecordsPerTransaction

martinweiler:
This is following up on the work from #2292 

The problem with job rescheduling still occurs when 'pagination' exists (There are more records to delete than the 'recordsPerTransaction' parameter). When this happens, in each page it ends up moving the date and time of next execution by not taking into account the date of the first page of the current cycle but using the current one.

This PR addresses the issue by storing the execution time of the original LogCleanupCommand.

### DIFF
--- a/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
+++ b/jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
@@ -132,7 +132,9 @@ public abstract class AbstractAvailableJobsExecutor {
                         ctx.setData("ClassLoader", cl);
 
                         // add scheduled execution time
-                        ctx.setData("scheduledExecutionTime", request.getTime());
+                        if(ctx.getData("scheduledExecutionTime") == null) {
+                            ctx.setData("scheduledExecutionTime", request.getTime());
+                        }
 
                         cmd = classCacheManager.findCommand(request.getCommandName(), cl);
                         // increment execution counter directly to cover both success and failure paths
--- a/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/LogCleanupCommandTest.java
+++ b/jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/LogCleanupCommandTest.java
@@ -222,4 +222,52 @@ public class LogCleanupCommandTest {
         assertTrue(diff >= 2000 && diff < 3000);
         System.clearProperty("byteman.jpaaudit.sleep");
     }
+
+    @Test(timeout=10000)
+    @BMScript(value = "byteman-scripts/simulateSlowLogCleanupCommand.btm")
+    public void logCleanupNextRunFixedPaginationTest() throws InterruptedException {
+        CountDownAsyncJobListener countDownListener = configureListener(1);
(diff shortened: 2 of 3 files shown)


jbpm-services/jbpm-executor/src/main/java/org/jbpm/executor/impl/AbstractAvailableJobsExecutor.java
```diff
@@ -132,7 +132,9 @@ public void executeGivenJob(RequestInfo request) {
                         ctx.setData("ClassLoader", cl);
 
                         // add scheduled execution time
-                        ctx.setData("scheduledExecutionTime", request.getTime());
+                        if(ctx.getData("scheduledExecutionTime") == null) {
```
fjtirado, line 135:
```suggestion
                        if(ctx.getData("scheduledExecutionTime") != null) {
```

fjtirado: I guess the intention of this change is to add the value only if it is NOT null

mareknovotny:
jenkins do fdb

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/LogCleanupCommandTest.java
```diff
@@ -222,4 +222,52 @@ public void logCleanupNextRunFixedIntervalTooSmallTest() throws InterruptedExcep
...
+        assertEquals(2, executedRequests.size());
+
+        executorService.cancelRequest(queuedRequests.get(0).getId());
+
+        long firstExecution = executedRequests.get(0).getTime().getTime();
+        long lastExecution = queuedRequests.get(queuedRequests.size()-1).getTime().getTime();
+
+        // time difference between first and last should be around 10 seconds, even if the original command got split into two executions due to RecordsPerTransaction
+        long diff = lastExecution - firstExecution;
+        assertTrue(diff < 11000);
```
gmunozfe, line 270: Test timeout is set to 10 seconds, so probably it could be greater in case to let this comparison as it is.

gmunozfe (approved):
Looks good to me, good work @martinweiler 
Just a minor comment related to the test timeout in case the test execution is close to 10 seconds, it could be greater.

jbpm-services/jbpm-executor/src/test/java/org/jbpm/executor/LogCleanupCommandTest.java
martinweiler: @gmunozfe Thanks for the input. However, the test is not waiting for the NextRun=10s to actually getting executed, test execution should be done in ~4s. The reason I chose 10s for the NextRun parameter is to make the assert clearer.

mareknovotny:
Failed tests not related to changes:
- org.kie.server.integrationtests.jbpm.jms/JmsResponseHandlerIntegrationTest/testStartAndCompleteTaskUseOfAsyncResponseHandler_JSON_
  `Synchronization failed for defined timeout: 240000 milliseconds.`
- org.kie.bc.DeploymentIT.org.kie.bc.DeploymentIT - this is random flaky test

gmunozfe:
@mareknovotny this is ready to merge

PR #2397, opened 2024-03-13
[JBPM-10225] Considering time for Dates

fjtirado:

[link](https://issues.redhat.com/browse/JBPM-10225)

### DIFF
--- a/jbpm-flow/src/main/java/org/jbpm/process/core/datatype/impl/type/ObjectDataType.java
+++ b/jbpm-flow/src/main/java/org/jbpm/process/core/datatype/impl/type/ObjectDataType.java
@@ -137,11 +137,12 @@ public class ObjectDataType implements DataType {
     }
 
     private static Collection<DateFormat> dateFormats = Arrays.asList(new SimpleDateFormat("yyyy-MM-ddHH:mm:ss"),
+            new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss"),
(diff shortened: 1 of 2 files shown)


fjtirado:
jenkins run fdb

jbpm-flow/src/main/java/org/jbpm/process/core/datatype/impl/type/ObjectDataType.java
```diff
@@ -137,11 +137,12 @@ private Optional<Object> getObjectFromClass(final Object value) {
     }
 
     private static Collection<DateFormat> dateFormats = Arrays.asList(new SimpleDateFormat("yyyy-MM-ddHH:mm:ss"),
+            new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss"),
```
gmunozfe, line 140: What about also adding a blank for covering more scenarios?
`new SimpleDateFormat("yyyy-MM-dd HH:mm:ss"),`

gmunozfe (commented):
Looks good to me, just a proposal to add also a format with blank in the middle

jbpm-flow/src/main/java/org/jbpm/process/core/datatype/impl/type/ObjectDataType.java
fjtirado: Thats is already covered, you can see by manually adding the test case

gmunozfe (approved):

PR #2401, opened 2024-03-19
Allow to configure default abort parent behaviour

albfan:
Behaviour after RHPAM 7.5 was modified. Any aborted subprocess will abort its parent.

This property allows to get the old behaviour so migrations from versions previous to RHPAM 7.5 behaves the same in recent RHPAM versions


jbpm-flow/src/main/java/org/jbpm/workflow/core/node/SubProcessNode.java
```diff
@@ -238,7 +239,7 @@ public boolean isAbortParent() {
         
         String abortParent = (String) getMetaData("customAbortParent");
         if (abortParent == null) {
-            return true;
+            return defaultAbortParent;
```
fjtirado, line 242:
```suggestion
            return Boolean.getBoolean("org.kie.jbpm.defaultAbortParent",true);
```

mareknovotny:
ok to test

mareknovotny:
jenkins do fdb

mareknovotny:
jenkins do fdb

albfan:
@elguardian can you add a review here?

Just to explain the scenario. A subprocess calling a Rest WIH (in sync thread of start node) where the external rest service is down, will abort the subprocess and abort the parent if "abort parent is set".

As I show different behaviour on this for RHPAM 7.3 with RHPAM 7.13.2, so comparing

jbpm-flow/src/main/java/org/jbpm/workflow/instance/node/SubProcessNodeInstance.java

this could be the relevant change:

https://github.com/kiegroup/jbpm/commit/50a61c9018#diff-c08f01bff8c2ced14acdfbcbe934979871b5ee6edb08e7f59de7f4f4ed073566L228-R237

we are now handling the error on start of process too:

https://issues.redhat.com/browse/RHPAM-2782

Since RHPAM 7.7.1 I think, so before that change any error starting a subprocess is not handled and just goes up

"abort parent" unset  just has same effect, ignore and send the exception up

For migrations before 7.7.1, this change might be allow easier migrations, while users are aware they need to modify their subprocess calls

albfan:
Will test a bit and let you know

PR #2412, opened 2024-04-18
[JBPM-10226] Integrate DB2 ddl-scripts with creation of required sequences

danielemams:
Integrate DB2 ddl-scripts with creation of required sequences

JIRA: [JBPM-10226](https://issues.redhat.com/browse/JBPM-10226)


fjtirado (approved):

elguardian:
@martinweiler this is fine. In the past the increments were a bit different depending on the allocation on the jpa entities not the case anymore.

https://github.com/kiegroup/jbpm/blob/main/jbpm-db-scripts/src/main/resources/db/ddl-scripts/postgresql/postgresql-bytea-jbpm-schema.sql#L1-L34

PR #2413, opened 2024-05-03
[JBPM-10228] Fixing wrong flush call

fjtirado:
**JIRA**:

[link](https://issues.redhat.com/browse/JBPM-10228)


gmunozfe (approved):
Good catch @fjtirado !!

PR #2439, opened 2024-08-20
[7.67.x-blue] [incubator-kie-issues#1049] Add index to ProcessInstanceLog table to improve LogCleanupCommand performance

baldimir:
Removes the index from the scripts for database systems that don't support defining indexes with "where" clause. It is only supported in SQLServer and PostgreSQL.


baldimir:
This needs to be backported to other branches, when merged.

elguardian (approved):
LGTM

martinweiler (approved):

PR #2440, opened 2024-08-22
Fix SQL scripts for MySQL.

baldimir:
- Fixes drop indexes, it needs table specification on MySQL, otherwise it fails. 
- Fixes TimerMappingInfo creation on MySQL, the generated keyword doesn't work on MySQL.


elguardian (approved):
LGTM

PR #2441, opened 2024-08-27
[JBPM-10242] Allow the possibility of disabling linear search for removeJob and getTimerByName operations

fjtirado:
Fix https://issues.redhat.com/browse/JBPM-10242

### DIFF
--- a/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
+++ b/jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
@@ -282,9 +282,16 @@ public class EJBTimerScheduler {
                     return false;
                 }
                 return true;
+            } else {
+                logger.warn("No TimerHandle found for {}: {}", ejbHandle, ejbTimerHandle);
             }
         } else {
             logger.warn("No timerJobInstance available for {}", ejbHandle);
+        }
+
+        if (!searchIfFailed) {
+            logger.warn("Timer not found for {} and {}, skipping list search", ejbHandle, ejbTimer);
+            return false;
         }
 
 		for (Timer timer : timerService.getTimers()) {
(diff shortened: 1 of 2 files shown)


gmunozfe:
jenkins do fdb

gmunozfe (approved):
Looks good to me! Thanks for your work @krisv , it's a weird scenario and this should make the code more robust by skipping the searching also in case there's no exception caught.

jbpm-services/jbpm-services-ejb/jbpm-services-ejb-timer/src/main/java/org/jbpm/services/ejb/timer/EJBTimerScheduler.java
```diff
@@ -271,15 +278,22 @@ public boolean removeJob(JobHandle jobHandle, Timer ejbTimer) {
...
                 }
                 return true;
+            } else {
+                logger.warn("No TimerHandle found for {}: {}", ejbHandle, ejbTimerHandle);
             }
         } else {
             logger.warn("No timerJobInstance available for {}", ejbHandle);
         }
 
+        if (!searchIfFailed) {
```
fjtirado, line 292: Is this change really needed? ( I have not gone through latest traces yet), but I feel this will prevent searching in case the search migh be needed ( to be honest, I do not completely understand the need for the search, so I was being conservative in the first commit and only disable it when it is clear that the times is there and the cancel has failed)

fjtirado: After reviewing the logs, this change makes sense. Probably we need more...

fjtirado:
jenkins run fdb

fjtirado:
jenkins run fdb

fjtirado:
jenkins run fdb

fjtirado:
jenkins run fdb

gmunozfe:
jenkins do fdb

gmunozfe (approved):
Looks good to me.
Works like a charm: tested regression scenarios with jbpm-playground tests for timers (complex scenarios) with and without cache and enabling/disabling flags.
Amazing work @fjtirado !

fjtirado:
thanks @gmunozfe

PR #2453, opened 2025-02-11
[CVE] CVE-2021-20250

LightGuard:
Issue in the remoting code that we're putting in from JBoss. EAP 7.4.20 is using wildfly, so aligning with that version.

See https://github.com/kiegroup/droolsjbpm-build-bootstrap/pull/2555


LightGuard:
jenkins run cdb

PR #2455, opened 2025-04-11
[JBPM-10249] Forcing dispose for command if transaction is not ready

fjtirado:

**JIRA**:

[link](https://issues.redhat.com/browse/JBPM-10249)

**referenced Pull Requests**: 
https://github.com/kiegroup/droolsjbpm-knowledge/pull/651

### DIFF
--- a/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
+++ b/jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
@@ -365,7 +365,11 @@ public abstract class AbstractRuntimeManager implements InternalRuntimeManager {
         }
     }
     
-    protected boolean canDispose(RuntimeEngine runtime) {
+    protected boolean canDispose (RuntimeEngine runtime) {
+    	return canDispose(runtime, false);
+    }
+    
+    protected boolean canDispose(RuntimeEngine runtime, boolean force) {
         
         if (runtime instanceof RuntimeEngineImpl) {
             RuntimeEngineImpl impl = (RuntimeEngineImpl)runtime;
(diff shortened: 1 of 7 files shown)


jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
```diff
@@ -385,7 +389,7 @@ protected boolean canDispose(RuntimeEngine runtime) {
         try {
             // check tx status to disallow dispose when within active transaction       
             TransactionManager tm = getTransactionManagerInternal(getEnvironment(runtime));
-            if (tm.getStatus() != TransactionManager.STATUS_NO_TRANSACTION
+            if (!force && tm.getStatus() != TransactionManager.STATUS_NO_TRANSACTION
```
fjtirado, line 392: The whole purpose of adding the value is to prevent that check to avoid disposing on RuntimeManager created for a Disposable command if the status of the transaction is unknown/active.

fjtirado:
jenkins retest this

gmunozfe (approved):
Looks good to me, awesome, thanks @fjtirado !

jbpm-runtime-manager/src/main/java/org/jbpm/runtime/manager/impl/AbstractRuntimeManager.java
martinweiler: @fjtirado it is not clear what this PR is trying to resolve. Can you provide more information about the root cause? What happened that caused the timer service not to release the lock?

fjtirado: @martinweiler  This change is intended to ignore the transaction status when a dispose request is performed from a timer. Normally, the dispose request is ignored if the transaction status is not the desired one, but in the case of a timer I feel the transaction status is not relevant (because the timer is terminated for all practical purposes). So when a timer is done, I think the dispose should go through regardless of the transaction status (or we have the risk of the thread holding the lock forever, as it was apparently happening in the setup of the user reporting the issue)

ghjboss: I've added some logging details to the jira https://issues.redhat.com/browse/RHPAM-5024 for your reference.

The root cause is not clear but is related to a transaction timeout.  It looks to me when a transaction timeout occurs sometimes the transaction manager may not retrieve the right status.

PR #2458, opened 2025-05-06
[JBPM-10251] Update TimerMappingInfo entry if EjbSchedulerService.internalSchedule is called

martinweiler:
See https://issues.redhat.com/browse/JBPM-10251 for details


elguardian (approved):
Very good work.
The listener should be triggered always but it was bypassed because ti was supposed the uuid was only scheduled once.
also good optimization regarding the return if the external timer id are the same.

LightGuard (approved):
Looks good. Maybe a comment pointing to the JIRA issue would be good in the code to know why we made the change, but the code looks fine.

PR #2505, opened 2026-01-26
[DBACLD-202894] LogCleanupCommand fails on MySQL with RecordsPerTrans…

abhijithumbe:
[KI DT456054] LogCleanupCommand fails on MySQL with RecordsPerTransaction=0

### DIFF
--- a/jbpm-audit/src/main/java/org/jbpm/process/audit/query/AbstractAuditDeleteBuilderImpl.java
+++ b/jbpm-audit/src/main/java/org/jbpm/process/audit/query/AbstractAuditDeleteBuilderImpl.java
@@ -32,12 +32,16 @@ import org.kie.api.runtime.Context;
 import org.kie.internal.query.ParametrizedUpdate;
 import org.kie.internal.query.QueryParameterIdentifiers;
 import org.kie.internal.runtime.manager.audit.query.AuditDeleteBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


jbpm-audit/src/main/java/org/jbpm/process/audit/query/AbstractAuditDeleteBuilderImpl.java
```diff
@@ -245,7 +249,29 @@ protected Subquery applyParameters(Subquery subquery) {
...
+    private boolean isRequestInfoDelete() {
+        String table = getQueryTable();
+        return table != null && table.endsWith("RequestInfo");
+    }
+
     public ParametrizedUpdate build() {
+        final int effectiveRecordsPerTransaction;
+        if(isMySqlDatabase() && isRequestInfoDelete() && recordsPerTransaction == 0) {
+            effectiveRecordsPerTransaction=1000;
+            logger.debug("MySQL dialect detected, Enabling batched delete (recordsPerTransaction={}) to avoid MySQL DELETE subquery limitation",effectiveRecordsPerTransaction);
```
martinweiler: Since this is quite important, I think logging it at `INFO` level is justified.

martinweiler (approved):
Nice work @abhijithumbe

jbpm-audit/src/main/java/org/jbpm/process/audit/query/AbstractAuditDeleteBuilderImpl.java
abhijithumbe: Ack. Changed logging level to Info

PR #2506, opened 2026-01-27
[DBACLD-196085] Timer issues with HumanTask suspend operation if nega…

abhijithumbe:
Timer issues with HumanTask suspend operation if negative value used for SuspendUntil

### DIFF
--- a/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
+++ b/jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
@@ -16,6 +16,8 @@
 
 package org.jbpm.kie.services.impl;
 
+import java.time.Duration;
+import java.time.format.DateTimeParseException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -25,6 +27,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.jbpm.kie.services.impl.admin.commands.UpdateTaskCommand;
+import org.jbpm.process.core.timer.DateTimeUtils;
 import org.jbpm.services.api.DeploymentService;
 import org.jbpm.services.api.RuntimeDataService;
 import org.jbpm.services.api.TaskNotFoundException;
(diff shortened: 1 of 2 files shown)


jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
```diff
@@ -140,6 +142,22 @@ protected void validateTask(String deploymentId, Long taskId, UserTaskInstanceDe
 	        throw new TaskNotFoundException("Task with id " + taskId + " is not associated with " + deploymentId);
 	    }
 	}
+
+	protected void validateTimerExpression(Map<String, Object>  parameters) {
+		logger.info("Validating timer expression {} ", parameters.get("suspendUntil"));
```
martinweiler: Please use `debug` here

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
```diff
@@ -140,6 +142,22 @@ protected void validateTask(String deploymentId, Long taskId, UserTaskInstanceDe
...
+		String isoDuration = (String) parameters.get("suspendUntil");
+		if(isoDuration == null) {
+			return;
+		}
+		try {
+			Duration duration = Duration.parse(isoDuration);
```
martinweiler: We also need to support simple time format (eg. `10s`)

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
abhijithumbe: changed log level to debug.

jbpm-services/jbpm-kie-services/src/main/java/org/jbpm/kie/services/impl/UserTaskServiceImpl.java
abhijithumbe: Updated code to support simple time format.

martinweiler (approved):
LGTM, thank you @abhijithumbe

